PR #32, opened 2019-06-17
Install script

maxitg:
## Changes

* Implements script that installs the built paclet.
* Only macOS is supported at the moment.

## Tests

* In Terminal, `cd` to the root repository directory.
* Run `./build.wls` to build the paclet.
* Run `./install.wls` to install the paclet.
* Restart Wolfram Kernel in Mathematica.
* Evaluate ``<<SetReplace` `` to import the package.
* Evaluate `SetReplace[{{0}}, {{x_}} :> {{x}}, Method -> "C++"]`.
* If the paclet is installed correctly, `{{0}}` will be returned and no messages will be produced.

### DIFF
--- a/install.wls
+++ b/install.wls
@@ -1,7 +1,7 @@
 #!/usr/bin/env wolframscript
 (* ::Package:: *)
 
-pacletInfo = (Association @@ Import["PacletInfo.m"]);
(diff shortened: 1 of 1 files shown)


maxitg:
Yes, prerequisites in README will come in a different PR.

Issue #14, opened 2019-04-05
Documentation for C++ code


maxitg:
Documentation should be added to the C++ source code of `SetReplace`.

PR #39, opened 2019-06-20
Abort C++ evaluation

### LINKED ISSUE #23: Abort C++ evaluation of SetReplace
Right now, C++ implementation of `SetReplace` called through LibraryLink cannot be aborted during evaluation, except by killing the kernel.

Apparently, there is a way to add some code to the library so that standard way of aborting in Mathematica would abort the evaluation of C++ code as well.

This should be implemented.

maxitg:
## Changes

* Closes #23.
* Allows aborting evaluation of C++ code.
* Abort check is inserted inside the low-level matching code, so it should happen fast in every case.
* Add a unit test.

## Tests

* The following should finish with `$Aborted` in about 1 second (otherwise it would take 3):
```
In[.] := TimeConstrained[
 SetReplace[{{0}}, 
  FromAnonymousRules[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> \
{{0}}}], 30], 1]
```
```
Out[.] = $Aborted
```
* Try aborting the following interactively:
```
In[.] := SetReplace[{{0}}, 
 FromAnonymousRules[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> \
{{0}}}], 30]
```
```
Out[.] = $Aborted
```
* Run unit tests
```
In[.] := TestReport["path_to_SetReplace.wlt"]
```
image

### DIFF
--- a/SetReplace.wlt
+++ b/SetReplace.wlt
@@ -247,6 +247,15 @@ VerificationTest[
 	MemoryConstraint -> 5*^6
 ]
 
+(* SetReplace: C++ aborting *)
+(* assumes example below runs slow, may need to be replaced in the future *)
+VerificationTest[
+	1.0 < Timing[TimeConstrained[SetReplace[
+			{{0}},
+  		FromAnonymousRules[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> {{0}}}],
+  		30], 1]][[1]] < 1.5
+]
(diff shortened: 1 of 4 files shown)


maxitg:
Agree, there is actually an issue about it https://github.com/maxitg/SetReplace/issues/14.

PR #42, opened 2019-06-20
Fix min time in Abort test

maxitg:
## Changes

It turns out it is possible for a `TimeConstrained[...]` evaluation to finish earlier than the specified time, at least this happens on Windows.

So, to avoid a failing test because of this, the min time constraint is changed to be below 1 second.

## Tests

* Evaluate unit tests (especially on Windows):
```
In[.] := TestReport["path_to_SetReplace.wlt"]
```
* Check all tests pass.
* Check if `CPUTimeUsed` in test 49 is smaller than 1 second.

### DIFF
--- a/SetReplace.wlt
+++ b/SetReplace.wlt
@@ -250,10 +250,10 @@ VerificationTest[
 (* SetReplace: C++ aborting *)
 (* assumes example below runs slow, may need to be replaced in the future *)
 VerificationTest[
-	1.0 < Timing[TimeConstrained[SetReplace[
+	0.8 < Timing[TimeConstrained[SetReplace[
 			{{0}},
   		FromAnonymousRules[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> {{0}}}],
-  		30], 1]][[1]] < 1.5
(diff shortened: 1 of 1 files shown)


SetReplace.wlt
```diff
@@ -250,10 +250,10 @@ VerificationTest[
 (* SetReplace: C++ aborting *)
 (* assumes example below runs slow, may need to be replaced in the future *)
 VerificationTest[
-	1.0 < Timing[TimeConstrained[SetReplace[
+	0.8 < Timing[TimeConstrained[SetReplace[
```
zhuwenzhen: Why not 0.9?

SetReplace.wlt
```diff
@@ -250,10 +250,10 @@ VerificationTest[
...
 			{{0}},
   		FromAnonymousRules[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> {{0}}}],
-  		30], 1]][[1]] < 1.5
+  		30], 1]][[1]] < 1.2
```
zhuwenzhen: I would really write comments here (in code) to explain the number.

PR #91, opened 2019-10-24
HypergraphPlot: "HyperedgeType" option

maxitg:
## Changes

* Implements "HyperedgeType" option of HypergraphPlot.
* Possible values are "Ordered" (sequence), "Cyclic" (last vertex connect to first), and "Unordered" (all vertices in a hyperedge form a complete graph).
* Specification is per-pattern, i.e., {"Ordered", {_, _, _} -> "Cyclic"} will make hyperedges "Ordered", except for 3-edges, which will be "Cyclic".

## Context

The idea is to have three options: "HyperedgeType", GraphLayout (both vertex- and edge-layout parts), and "HyperedgeShapeFunction" (responsible for drawing).

"HyperedgeType" will have an effect on all other options, and possible can be specified with `Hyperedge` head in the future.

`"VertexLayout"` will take for the most part the same layouts as `GraphPlot`, with possible an addition of `"BraneElectricalEmbedding"`.

`"EdgeLayout"` will return a list of curves for hyperedge components (possibly a polygon, but might not necessarily).

`"HyperedgeShapeFunction"` will take these curves, and just do the drawing, no layout.

## Tests

* Run unit tests: `./build.wls && ./install.wls && ./test.wls`.
* Try different layouts:
```
In[] := HypergraphPlot[{{1, 2, 3, 4, 5}, {5, 6, 7, 8, 9}}, 
   VertexLabels -> Automatic, "HyperedgeType" -> #] & /@ {"Ordered",
   "Cyclic", "Unordered"}
```
image

```
In[] := HypergraphPlot[{{1, 2, 3, 4, 5, 6, 7, 8, 9}, {1, 4, 7}}, 
   VertexLabels -> Automatic, "HyperedgeType" -> #] & /@ {"Ordered",
   "Cyclic", "Unordered"}
```
image

* Use different layouts for different hyperedges:
```
In[] := HypergraphPlot[{{1, 2, 3, 4, 5, 6}, {1, 2, 3, 4}}, 
 VertexLabels -> Automatic, 
 "HyperedgeType" -> {"Ordered", {_, _, _, _} -> "Cyclic"}]
```
image

### DIFF
--- a/SetReplace/HypergraphPlot.m
+++ b/SetReplace/HypergraphPlot.m
@@ -69,7 +69,8 @@ HypergraphPlot[edges_, o : OptionsPattern[]] := 0 /;
 
 Options[HypergraphPlot] = Join[{
 	"HyperedgeLayout" -> "Ordered",
-	PlotStyle -> (ColorData[97][# + 1] &)},
+	PlotStyle -> (ColorData[97][# + 1] &),
+	DirectedEdges -> True},
 	Options[GraphPlot]];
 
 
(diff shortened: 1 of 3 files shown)


maxitg:
This approach is not actually going to work, because "HyperedgeShapeFunction" will need to have information from all edges in order to make a sensible layout. Further, what is implemented here is actually a "VertexLayout" part of "GraphLayout", so it should be called such.

PR #93, opened 2019-10-28
NodeNamingFunction

### LINKED ISSUE #74: NodeNamingFunction
Currently, the newly created vertices are named automatically with unique symbols. We would like some customization for that, which could be `"NodeNamingFunction"` option of `WolframModel`.

By default, In the output from `WolframMode`l, all nodes could be labeled sequentially starting from 1, across all generations in the evolution. 

`"NodeNamingFunction" -> All` renames all nodes to be labeled sequentially starting from 1 , across all generations in the evolution. 

`"NodeNamingFunction" -> Automatic` uses any existing labels, and labels new nodes with names such as `vi`.

`"NodeNamingFunction" -> f` names nodes with `f[name, list]`, where name is the existing name of a node, or `None` if the node is new, and list is a list of the names of all nodes created so far, in the order they were created.  *Alternative: `f[name, count]`, where count is the number of nodes that exist so far.*

maxitg:
## Changes

* Closes #74 .
* Implements `"NodeNamingFunction"` option of `WolframModel`, that renames atoms in the output.
  * `"NodeNamingFunction" -> None` leaves the atoms as-is, which in the case of pattern rules is an obvious default. In the case of anonymous rules, it uses nodes named `v*` as previously.
  * `"NodeNamingFunction" -> All` renames all atoms sequentially with integers (including ones coming from initial condition).
  * `"NodeNamingFunction" -> Automatic` keeps existing names as-is (including any output of `Module` of pattern rules, but not including internally produced `Module` for anonymous rules).
* Adds unit tests for `"NodeNamingFunction"`.

## Tests

* Run unit tests: `./build.wls && ./install.wls && ./test.wls`
* By default, if anonymous rules are used, integers are used for new atoms:
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {3, 2}}, {{1, 2}}, "FinalState"]
```
```
Out[] = {{1, 3}, {3, 2}}
```
* The atoms used in the initial condition are avoided, and unused numbers are used for new atoms:
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {3, 2}}, {{1, 2}, {2, 
   4}}, "FinalState"]
```
```
Out[] = {{1, 3}, {3, 2}, {2, 5}, {5, 4}}
```
* This is called `"NodeNamingFunction" -> Automatic`. For pattern rules, it does not have any effect, because all new atoms are introduced explicitly:
```
In[] := WolframModel[<|
  "PatternRules" -> {{a_, b_}} :> 
    Module[{c}, {{a, c}, {c, b}}]|>, {{1, 2}}, "FinalState"]
```
```
Out[] = {{1, v1038}, {v1038, 2}}
```
* Other available options are `None` and `All`. `All` renames all atoms, even initial condition, and can be used for pattern rules as well:
```
In[] := WolframModel[<|
  "PatternRules" -> {{a_, b_}} :> 
    Module[{c}, {{a, c}, {c, b}}]|>, {{1, 2}}, "FinalState", 
 "NodeNamingFunction" -> All]
```
```
Out[] = {{1, 3}, {3, 2}}
```
* For anonymous rules it treats all expression on level 2 as atomic, consistent with `ToPatternRules`:
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {3, 2}}, {{s[1], 
   s[2]}}, "FinalState"]
```
```
Out[] = {{s[1], 1}, {1, s[2]}}
```
* `"NodeNamingFunction" -> None` leaves the names generated by a module, which is the behavior before this PR:
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {3, 2}}, {{1, 2}}, "FinalState", 
 "NodeNamingFunction" -> None]
```
```
Out[] = {{1, v1041}, {v1041, 2}}
```


zhuwenzhen:
No one reviews for you? Poor Max...

PR #180, opened 2020-01-17
GeneralizedGridGraph

maxitg:
## Changes
* Implements `GeneralizedGridGraph`, which is similar to `GridGraph`, but can make the graph `"Circular"` or / and `"Directed"` in each dimension.

## Tests
* Make an ordinary grid graph:
```
In[] := GeneralizedGridGraph[{10, 10}]
```
image
* Make it circular in one of the dimensions, and directed in another:
```
In[] := GeneralizedGridGraph[{10 -> "Circular", 10 -> "Directed"}]
```
image
* Make the same dimension both circular and directed:
```
In[] := GeneralizedGridGraph[{10 -> {"Circular", "Directed"}, 5}]
```
image
* Make a directed hypercube graph:
```
In[] := Graph[GeneralizedGridGraph[Thread[{2, 2, 2, 2} -> "Directed"]], 
 GraphLayout -> "SpringElectricalEmbedding"]
```
image

### DIFF
--- a/SetReplace/GeneralizedGridGraph.m
+++ b/SetReplace/GeneralizedGridGraph.m
@@ -1,0 +1,71 @@
+Package["SetReplace`"]
+
+PackageExport["GeneralizedGridGraph"]
+
+(* Documentation *)
+
+GeneralizedGridGraph::usage = usageString[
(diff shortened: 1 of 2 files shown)


maxitg:
Good catch with coordinates! Turns out there is a `{"GridEmbedding", "Dimension" -> {_, _}}` layout I can use. I'll add that.

maxitg:
@GetJonWithIt Embedding for 2D linear case is fixed. Please take a look.

PR #181, opened 2020-01-23
Hash vertex names

### LINKED ISSUE #146: NodeNamingFunction -> Hash
There are two properties we might want to have for node names:

1. They are deterministic, i.e., rerunning the `WolframModel` will yield exactly the same evolution object.
2. They are local, i.e., changes to parts of the set will not affect node names elsewhere.

One possible solution is to have node names computed as a hash of the instantiated rule inputs. This way they are deterministic, and changes elsewhere would not affect the relevant hashes.

This also brings the following subtlety: what if we have a rule that is purely (or partially) constructive, and the same set of inputs is used multiple times. Then, the atoms we will create will have the same names, even though they are (presumably) supposed to be different.

There are two possible solutions:

1. Name expressions (edges) as well. This can just be a hash of the atoms. This way, if the rule copies its inputs to its outputs, the has of the expressions will change. The atom names can then include the input expression hashes in their hash.
2. Embrace the auto-deduplication this approach provides, and consider the evolution which will auto-deduplicate atoms if they are created from the same inputs. It will probably make sense to deduplicate the expressions (multi-edges) as well in this case. The evolution will be different, but it still might be interesting. I'm also curious to know what implications that would have for quantum physics...

maxitg:
***WIP: Do not merge***

## Changes
* Closes #146.
* Changes C++ code to use name vertices as hashes of the edges in the corresponding input (indexed so that if multiple vertices are created they have different names).
* Edges are now assigned IDs computed as hashes of ids of inputs of an event that created that expression.
* This produces the same evolution as before, however, the vertex names are now consistent if different event ordering functions are used.
* To see these vertex names, use `"EventOrderingFunction" -> None` for now.

## Issues
* Does not compile on Linux due to using `std::hash` of `__int128` which is not supported everywhere (tests pass on a Mac except ones for node naming function).
* Hash function might have collisions. Might be better to replace with SHA-1 unless there are serious performance implications.
* We don't need the old approach in C++ code anymore, however, we need to clean up how the vertex names are returned in all cases of `WolframModel`, including when pattern rules with a module are used (in which case hashed vertices should not be returned).
* While vertex names in C++ code use `__int128`, `int` is passed back to WL code, which will introduce collisions at around 65k vertices.
* Hash function may not be the same across platforms / compiler / std versions.

## Tests
* There are way more tests needed, however, one can already see the vertex names being consistent for different evolution orders (which was not the case before):
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, {{1, 
   1}}, 2, "FinalState", "NodeNamingFunction" -> None]
```
```
Out[] = {{-7635323901126876262, 6227095357045480841}, {-5337242763971511340, 
  1}, {1, -7635323901126876262}, {1, -7635323901126876262}, {1, 
  854235261605426321}, {1, 854235261605426321}, {854235261605426321, 
  6227095357045480841}, {6227095357045480841, -5337242763971511340}, \
{6227095357045480841, -5337242763971511340}}
```
```
In[] := Sort[WolframModel[{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, {{1, 1}}, 2, 
  "FinalState", "NodeNamingFunction" -> None, 
  "EventOrderingFunction" -> "Random"]]
```
```
Out[] = {{-7635323901126876262, 6227095357045480841}, {-5337242763971511340, 
  1}, {1, -7635323901126876262}, {1, -7635323901126876262}, {1, 
  854235261605426321}, {1, 854235261605426321}, {854235261605426321, 
  6227095357045480841}, {6227095357045480841, -5337242763971511340}, \
{6227095357045480841, -5337242763971511340}}
```
```
In[] := % === %%
```
```
Out[] = True
```


maxitg:
Closing for now. We will need this in the future, but not soon.

PR #182, opened 2020-01-25
New property names

### LINKED ISSUE #85: Consistent evolution object property names
Currently, `WolframModel` use names such as *state* and *update* instead of *set* and *event*. However, evolution object properties that take arguments use other names.

We should make them consistent by renaming evolution object properties.

maxitg:
## Changes
* Closes #85.
* Implemented new properties for `WolframModelEvolutionObject`:
  * `"FinalStatePlot"`: `WolframModelPlot`s the last state,
  * `"StatesPlotsList"`: yields a list of `WolframModelPlot`s,
  * `"PartialGenerationsCount"`: total - complete generations,
  * `"GenerationComplete"`: yields true/false depending whether the last (or specified) generation is complete,
  * `"GenerationEventsCountList"`: event counts in between generations,
  * `"GenerationEventsList"`: events grouped by generation,
  * `"VertexCountList"`: counts of vertices at each generation,
  * `"EdgeCountList"`: counts of edges at each generation,
  * `"EventsStatesList"`: riffled states and events.
* Changed behavior of `"GenerationsCount"`, which now returns a list `{"CompleteGenerationsCount", "PartialGenerationsCount"}`.
* Renamed properties (old names are still allowed):
  * `"UpdatedStatesList"` -> `"AllEventsStatesList"`,
  * `"AllExpressions"` -> `"AllEventsEdgesList"`,
  * `"CreatorEvents"` -> `"EdgeCreatorEventIndices"`,
  * `"DestroyerEvents"` -> `"EdgeDestroyerEventIndices"`,
  * `"MaxCompleteGeneration"` -> `"CompleteGenerationsCount"`,
  * `"EventGenerations"` -> `"AllEventsGenerationsList"`,
  * `"ExpressionGenerations"` -> `"EdgeGenerationsList"`,
  * `"EventsCount"` -> `"AllEventsCount"`,
  * `"EventsList"` -> `"AllEventsList"`,
  * `"AtomsCountFinal"` -> `"FinalDistinctElementsCount"`,
  * `"AtomsCountTotal"` -> `"AllEventsDistinctElementsCount"`,
  * `"ExpressionsCountFinal"` -> `"FinalEdgeCount"`,
  * `"ExpressionsCountTotal"` -> `"AllEventsEdgesCount"`,
  * `"SetAfterEvent"` -> `"StateAfterEvent"`

## Review questions
* The events in the output of `"EventsStatesList"` refer to indices of expressions in “AllExpressions” rather than to the states that surround them, which make it rather confusing to read and unusable without additional information. Do you think it would be better to reindex them so that they refer to the indices in surrounding states? Or is it better to keep events spec consistent for all properties? Another alternative would be to return states as well as lists of indices in “EventsStatesList”.
* Please look at the new property names, and comment if it's clear from the names what they do. Some of these names are quite confusing to me.
* @sskini2: This is the largest piece needed to make `WolframModel` consistent with documentation.

## Tests
* Plot final state directly without calling `WolframModelPlot`:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
   1}}, 6, "FinalStatePlot"]
```
image
* Label vertices:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 1}}, 
  3]["FinalStatePlot", VertexLabels -> Automatic]
```
image
* Plot all generations:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
   1}}, 5, "StatesPlotsList"]
```
image
* See how the number of incomplete generations increases with the events count:
```
In[] := ListPlot[WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
      1}}, <|"MaxEvents" -> #|>, "PartialGenerationsCount", 
    "EventOrderingFunction" -> "Random"] & /@ Range[200]]
```
image
* Check if a given generation is complete:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 1}}, <|
  "MaxEvents" -> 1000|>, "EventOrderingFunction" -> "Random"]
```
image
```
In[] := %["GenerationComplete", 4]
```
```
Out[] = False
```
* Check how many events occur in between generations:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
   1}}, 10, "GenerationEventsCountList"]
```
```
Out[] = {1, 2, 4, 8, 16, 32, 64, 128, 256, 512}
```
* Show events grouped by generation:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 1}}, 3, 
  "GenerationEventsList"] // Column
```
image
* Check how the number of vertices grows with generations:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
   1}}, 10, "VertexCountList"]
```
```
Out[] = {1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024}
```
* Same for the number of edges:
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 
   1}}, 10, "EdgeCountList"]
```
```
Out[] = {1, 3, 7, 15, 31, 63, 127, 255, 511, 1023, 2047}
```
* Show events and states in the same list (edge indices are given in terms of `"AllEventsEdgesList"`):
```
In[] := WolframModel[{{1, 1}} -> {{1, 2}, {2, 2}, {2, 2}}, {{1, 1}}, 3, 
  "EventsStatesList"] // Grid
```
image

### DIFF
--- a/SetReplace/WolframModel.wlt
+++ b/SetReplace/WolframModel.wlt
@@ -271,7 +271,7 @@
       (*** Generations and Events ***)
 
       VerificationTest[
-        WolframModel[1 -> 2, {1}, 2]["GenerationsCount"],
+        WolframModel[1 -> 2, {1}, 2]["TotalGenerationsCount"],
         2
       ],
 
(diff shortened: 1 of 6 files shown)


maxitg:
> I quite like the idea of modifying AllEventsEdgesList so that it specifies states as lists of indices too, so that at least we have consistency between the two schemes.

I think I completely confused what I wanted to say. `"AllEventsEdgesList"` does not return states, so it would not make sense to return edges as indices there (it would just be a range of its length). What we could do, however, is to return states in `"EventsStatesList"` as list of indices from `"AllEventsEdgesList"`.

This way, one could see how events and states in that list relate to each other.

Stephen and I talked about it a few days ago, and we decided on this:
> "EventsStatesList" gives pairs of events and the states they produce.  The events as specified as list of edges from "AllEventsEdgesList".
However, we didn't consider an idea to use indices for specifying states in `"EventsStatesList"`.

PR #193, opened 2020-02-05
"ArrowheadLength" -> 0.15

maxitg:
## Changes
* Increases default `"ArrowheadLength"` to `0.15`.
* While we are trying to figure out the strategy to scale arrowheads for all possible plot ranges and images sizes (if it's even possible), I think the constant value of `0.15` works better than the current `0.10`.

## Tests
* The new default:
```
In[] := WolframModelPlot[Partition[Range[16], 3, 2, 1]]
```
image
* Comparison (old on the left, new on the right):
```
In[] := Table[WolframModelPlot[Partition[Range[size], 3, 2, 1], 
     "ArrowheadLength" -> #] & /@ {0.1, 0.15}, {size, {6, 10, 
    16}}] // Grid
```
image


maxitg:
There is a new, more elaborate spec from Jeremy that I will implement instead.

PR #197, opened 2020-02-11
Automatic init

### LINKED ISSUE #133: Automatic initial conditions
It would be nice to be able to use `Automatic` as a second argument for `WolframModel`, which would insert an appropriate number of lists of zeros, and, i.e.,
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, Automatic]
```
will be equivalent to
```
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, {{0, 0}}]
```

maxitg:
## Changes
* Closes #133.
* Allows use of `Automatic` instead of the initial state.
* All anonymous rules are supported, but pattern rules are not.

## Tests
* Use "compactified" left-hand side of the rule as an initial condition:
```
In[] := WolframModel[{{1, 2}, {2, 3}} -> {{1, 2}, {2, 4}, {4, 1}, {4, 3}}, 
  Automatic][0]
```
```
Out[] = {{1, 1}, {1, 1}}
```
```
In[] := WolframModel[{{1, 2}, {2, 3}} -> {{1, 2}, {2, 4}, {4, 1}, {4, 
    3}}, Automatic, 10, "FinalStatePlot"]
```
image
* In the case of multiple rules, the minimal combination of both is used such that each rule can still be applied:
```
In[] := WolframModel[{{{1}, {1, 2}} -> {{1, 2}, {2}}, {{1, 2}, {2, 3}} -> {{1,
      3}, {3, 2}, {2, 4}}}, Automatic, 5, "StatesPlotsList"]
```
image

### DIFF
--- a/SetReplace/WolframModel.m
+++ b/SetReplace/WolframModel.m
@@ -80,7 +80,24 @@ fromRulesSpec[<|"PatternRules" -> rules_|>] := rules
 (*Init*)
 
 
-fromInitSpec[initSpec_] := initSpec
+fromInitSpec[rulesSpec_, initSpec_List] := initSpec
+
+
+fromInitSpec[rulesSpec_Rule, Automatic] := fromInitSpec[{rulesSpec}, Automatic]
+
+
+fromInitSpec[rulesSpec_List, Automatic] := Catenate[
+	If[#2 === 0, ConstantArray[1, #1], ConstantArray[1, {##}]] & @@@
+		Reverse /@ Sort[Normal[Merge[Counts /@ Map[Length, If[ListQ[#], #, {#}] & /@ rulesSpec[[All, 1]], {2}], Max]]]]
+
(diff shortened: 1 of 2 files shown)


maxitg:
> Is there a particular reason why we cannot support pattern rules (i.e. by finding the minimal hypergraph that matches a particular pattern)?

We probably can in special cases, but I'm worried about cases like this:
```
{a_?EvenQ, b_?OddQ} :> {a + b}
```
where in general it would be completely non-trivial to find an example. And if the pattern rules are just normal hypergraph rules, we can just as well use the anonymous version of them, so I did not see much point implementing `Automatic` in that case (since carefully parsing the `Module` while correctly holding all expressions is not trivial).

PR #261, opened 2020-03-09
Branchial graph EdgeStyle

maxitg:
## Changes
* Adds `"EdgeStyle"` for `"BranchialGraph"` in `style.m`.

## Tests
* Look up branchial color:
```
In[] := WolframPhysicsProjectStyleData["BranchialGraph", "EdgeStyle"]
```
image


maxitg:
> Should I be concerned about this failed test?

That was one poorly written test. The fix is here: #262.

PR #266, opened 2020-03-13
Multiway styles

GetJonWithIt:
Updated style.m to include MultiwaySystem-specific graph styles (StatesGraph, EvolutionCausalGraph and BranchialGraph).

### DIFF
--- a/Kernel/style.m
+++ b/Kernel/style.m
@@ -1,3 +1,5 @@
+(* ::Package:: *)
+
 Package["SetReplace`"]
 
 PackageExport["$WolframPhysicsProjectPlotThemes"]
@@ -48,6 +50,11 @@ PackageScope["$ruleArrowPadding"]
 PackageScope["$ruleArrowStyle"]
 PackageScope["$ruleGridColor"]
 PackageScope["$ruleImageSizePerPlotRange"]
+PackageScope["$statesGraphVertexStyle"]
+PackageScope["$statesGraphEdgeStyle"]
+PackageScope["$evolutionCausalGraphEvolutionEdgeStyle"]
+PackageScope["$evolutionCausalGraphCausalEdgeStyle"]
+PackageScope["$branchialGraphEdgeStyle"]
 
 $styleNames = KeySort /@ KeySort @ <|
   "EvolutionObject" -> <|"Icon" -> $evolutionObjectIcon|>,
@@ -98,9 +105,18 @@ $styleNames = KeySort /@ KeySort @ <|
     "ImageSizePerPlotRange" -> $ruleImageSizePerPlotRange
   |>,
 
-  (* For future use *)
+  (* MultiwaySystem styles *)
+  
+  "StatesGraph" -> <|
+    "VertexStyle" -> $statesGraphVertexStyle,
+    "EdgeStyle" -> $statesGraphEdgeStyle
+  |>,
+  "EvolutionCausalGraph" -> <|
+    "EvolutionEdgeStyle" -> $evolutionCausalGraphEvolutionEdgeStyle,
+    "CausalEdgeStyle" -> $evolutionCausalGraphCausalEdgeStyle
+  |>,
   "BranchialGraph" -> <|
-    "EdgeStyle" -> $branchialEdgeStyle
+    "EdgeStyle" -> $branchialGraphEdgeStyle
   |>
 |>;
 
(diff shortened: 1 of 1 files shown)


Kernel/style.m
```diff
@@ -170,7 +186,17 @@
...
+  
+  (* MultiwaySystem styles *)
+  
+  (* States graph *)
+  $statesGraphVertexStyle -> Directive[Opacity[0.7], Hue[0.62, 0.45, 0.87]],
+  $statesGraphEdgeStyle -> Directive[{Hue[0.75, 0, 0.35], Dashing[None]}],
+
+  (* Evolution causal graph *)
+  $evolutionCausalGraphEvolutionEdgeStyle -> Directive[{Hue[0.75, 0, 0.24], Dashing[None]}],
+  $evolutionCausalGraphCausalEdgeStyle -> Directive[{Hue[0.07, 0.78, 1], Dashing[None]}],
```
maxitg: @sskini2 Is it intentional that this is different from `$causalGraphEdgeStyle` (`EdgeStyle` for a non-multiway causal graph)?

Kernel/style.m
```diff
@@ -98,9 +105,18 @@
...
   |>,
 
-  (* For future use *)
+  (* MultiwaySystem styles *)
+  
+  "StatesGraph" -> <|
+    "VertexStyle" -> $statesGraphVertexStyle,
+    "EdgeStyle" -> $statesGraphEdgeStyle
+  |>,
+  "EvolutionCausalGraph" -> <|
```
maxitg: Don't we need a vertex style(s) here? And for the branchial graph?

Kernel/style.m
```diff
@@ -1,3 +1,5 @@
+(* ::Package:: *)
```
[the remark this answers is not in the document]
GetJonWithIt: No, I think that just got added automatically. I'm happy to remove if it's inconsistent.

Kernel/style.m
```diff
@@ -170,7 +186,17 @@
   $ruleArrowStyle -> GrayLevel[0.65],
   $ruleGridColor -> GrayLevel[0.85],
   $ruleImageSizePerPlotRange -> 128,
+  
+  (* MultiwaySystem styles *)
+  
+  (* States graph *)
+  $statesGraphVertexStyle -> Directive[Opacity[0.7], Hue[0.62, 0.45, 0.87]],
+  $statesGraphEdgeStyle -> Directive[{Hue[0.75, 0, 0.35], Dashing[None]}],
```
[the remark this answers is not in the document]
GetJonWithIt: That's just a verbatim copy from the style guidelines that I was sent by Jeremy. I agree, it shouldn't affect anything (so I'm happy to delete if necessary).

Kernel/style.m
GetJonWithIt: Yes, the design spec is different, as a means of making the causal edges more distinct from the evolution edges in the evolution causal graph (recall, causal edges used to be dashed here; SW then decided that they should be solid instead, but with a different color scheme).

Kernel/style.m
GetJonWithIt: I don't think so, since the two relevant vertex styles (for state and event vertices) can be inherited from the state and causal graph cases. Likewise for the branchial graph. Do you think it's worth repeating them here?

Kernel/style.m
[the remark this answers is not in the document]
maxitg: In this case, I think it's better to remove it from here since this file was not originally created by the front end, and does not have other related stuff (like sections, etc.)

Kernel/style.m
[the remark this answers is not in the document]
maxitg: Yes, I think we should remove it if it does not do anything (Jeremy probably just left it there for historical reasons).

Kernel/style.m
maxitg: Yes, thanks for the clarification.

Kernel/style.m
maxitg: I think it is, for the sake of clarity. We can point them to the same variables (i.e., `"StateVertexStyle" -> $statesGraphVertexStyle`, etc.). But keep in mind, `WolframPhysicsProjectStyleData` is exposed to users, and I think it will be hard for people to figure out where did the vertex style come from if it's not there.

maxitg: Isn't it better to call them `"EventVertexStyle"` and `"StateVertexStyle"`? One could argue events are also part of the evolution...

Kernel/style.m
```diff
@@ -170,7 +187,17 @@
   $ruleArrowStyle -> GrayLevel[0.65],
   $ruleGridColor -> GrayLevel[0.85],
   $ruleImageSizePerPlotRange -> 128,
+  
+  (* MultiwaySystem styles *)
+  
+  (* States graph *)
+  $statesGraphVertexStyle -> Directive[Opacity[0.7], Hue[0.62, 0.45, 0.87]],
+  $statesGraphEdgeStyle -> Directive[{Hue[0.75, 0, 0.35]}],
```
maxitg: Sorry for nit, but we don't need a `Directive` if there is only one argument. Same for the two cases below.

maxitg:
@GetJonWithIt Do you want me to merge once CI passes? (So that you don't have to wait for it.)

PR #267, opened 2020-03-15
README documentation

### LINKED ISSUE #253: Update README for 1.0
README file currently uses outdated syntax for many functions. Moreover, it is severely incomplete as it does not use any functionality introduced since 0.2.0. It should be updated before launch because otherwise, it is very hard to figure out how to use the package.

maxitg:
## Changes
* Closes #253.
* Adds complete documentation to README.
* This is a user guide README, it does not include contributing guidelines.

## Comments
* I put everything in a single file. It is quite large (57 pages if printed), however, it loads quite quickly with my Internet connection (~4.5 seconds) and is not too large (7 MB). Navigation is not too difficult either, as I added navigation bars at the beginning of each large ((sub)sub)section.
* Let me know if it's too slow for you, or hard to navigate.
* Of course, any other suggestions are welcome.
* I will wait for a review from both @GetJonWithIt and @sskini2 before merging.

## Tests
* Read the README, and comment on parts that are confusing.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/267)


README.md
```diff
@@ -1,67 +1,76 @@
...
 ```
-In[.] := SetReplace[{{0, 1}},
- Echo @ FromAnonymousRules[{{0, 1}} -> {{0, 2}, {2, 1}}], 2]
+{{1, 2, 3}, {2, 4, 5}, {4, 6, 7}}
 ```
+which we can render as a collection of ordered hyperedges:
 ```
->> {{v1_, v2_}} :> Module[{v3}\, {{v1, v3}, {v3, v2}}]
-Out[.] = {{v13, 1}, {0, v14}, {v14, v13}}
+In[] := HypergraphPlot[{{1, 2, 3}, {2, 4, 5}, {4, 6, 7}},
```
sskini2: Using HypergraphPlot on purpose?

README.md
```diff
@@ -72,6 +72,32 @@ In[] := HypergraphPlot[
 
 Exploring the models of this more complicated variety is what this package is mostly designed for.
 
+# Getting Started
+
+## Prerequisites
+
+To start using `SetReplace` you only need two things.
+
+* [Wolfram Language 12.1+](https://www.wolfram.com/language/) including [WolframScript](https://www.wolfram.com/wolframscript/). A free (although not open-source) version is available as [Wolfram Engine](https://www.wolfram.com/engine/).
```
sskini2: I think we should say "A free version is available as Wolfram Engine". Not no-cost

README.md
```diff
@@ -72,6 +72,32 @@ In[] := HypergraphPlot[
...
+COMPILER=CCompilerDriver\`ClangCompiler\`ClangCompiler COMPILER_INSTALLATION=/usr/bin ./build.wls
+```
+Here `ClangCompiler` can be replaced with one of `"Compiler" /. CCompilers[Full]` (run ``<< CCompilerDriver` `` to use `CCompilers`), and `COMPILER_INSTALLATION` is a directory in which the compiler binary can be found.
+
+3. Run `./install.wls` to install the paclet into your Wolfram system.
+4. Evaluate `PacletDataRebuild[]` in all running Wolfram kernels.
+5. Evaluate ``<< SetReplace` `` every time prior to using package functions.
+
+A less frequently updated version is available through the Wolfram's public paclet server and can be installed by running `PacletInstall["SetReplace"]`.
+
```
sskini2: Do you also want to mention WFR here?

README.md
```diff
@@ -98,6 +98,65 @@ Here `ClangCompiler` can be replaced with one of `"Compiler" /. CCompilers[Full]
 
 A less frequently updated version is available through the Wolfram's public paclet server and can be installed by running `PacletInstall["SetReplace"]`.
 
+# Symbols and Functions
+
+## SetReplace, SetReplaceList, SetReplaceAll, SetReplaceFixedPoint, SetReplaceFixedPointList
+
+`SetReplace` (and related `SetReplaceList`, `SetReplaceAll`, `SetReplaceFixedPoint` and `SetReplaceFixedPointList`) are the functions the package is named after. They are quite simple, don't have a lot of options, and simply perform replacement operations either one-at-a-time (as in the case of `SetReplace`), to all non-overlapping subsets (`SetReplaceAll`), or until no more matches can be made (`SetReplaceFixedPoint`). A suffix `*List` implies the function will return a set after each replacement instead of just the final result.
+
+These functions are good for their simplicity, but we don't use them much anymore as a more advanced `WolframModel` incorporates all of these features plus other utilities helpful for the exploration of our models.
```
sskini2: I for some reason don't like this sentence. I think it is a little negative..  I would probably say something like this. 

These functions are good for their simplicity and can be primarily used for its low level usage. WolframModel is advanced version of SetReplace and incorporates ... ... of our models.

README.md
```diff
@@ -98,6 +98,65 @@ Here `ClangCompiler` can be replaced with one of `"Compiler" /. CCompilers[Full]
 
 A less frequently updated version is available through the Wolfram's public paclet server and can be installed by running `PacletInstall["SetReplace"]`.
 
+# Symbols and Functions
+
+## SetReplace, SetReplaceList, SetReplaceAll, SetReplaceFixedPoint, SetReplaceFixedPointList
+
+`SetReplace` (and related `SetReplaceList`, `SetReplaceAll`, `SetReplaceFixedPoint` and `SetReplaceFixedPointList`) are the functions the package is named after. They are quite simple, don't have a lot of options, and simply perform replacement operations either one-at-a-time (as in the case of `SetReplace`), to all non-overlapping subsets (`SetReplaceAll`), or until no more matches can be made (`SetReplaceFixedPoint`). A suffix `*List` implies the function will return a set after each replacement instead of just the final result.
```
sskini2: Are you going to not say in the beginning what is the argument structure for any SetReplace functions?

README.md
```diff
@@ -98,6 +98,65 @@ Here `ClangCompiler` can be replaced with one of `"Compiler" /. CCompilers[Full]
...
+Out[] = {6, 3, 8}
+```
+
+`SetReplaceFixedPoint` and `SetReplaceFixedPointList` perform replacements for as long as possible as previously mentioned:
+```
+In[] := SetReplaceFixedPoint[{1, 2, 5, 3, 6}, {a_, b_} :> a + b]
+Out[] = {17}
+```
+```
+In[] := SetReplaceFixedPointList[{1, 2, 5, 3, 6}, {a_, b_} :> a + b]
```
sskini2: perhaps mention SetReplaceFixedPointList is equivalent to SetReplaceList[XXX, Infinity]

README.md
maxitg: Yes, because I think `WolframModelPlot` would be confusing given that I have not introduced not only `WolframModel`, but even any kind of evolution at this point. Do you think it's a bad idea?

README.md
maxitg: Ok, I agree. I was trying to avoid confusion with free as open-source, but I guess it's clear from the context.

README.md
maxitg: Not particularly, because the WFR version is somewhat unstable (not tested, requires a constant Internet connection, and there is the whole asynchronous updating problem), does not have all the features (by necessity), and I would much prefer if people use the paclet directly, as it would be easier for both them and me (since I would not have to provide technical support). Also, I don't want people to create issues about WFR functions here on GitHub, and mentioning WFR might cause that to happen.

README.md
maxitg: Good point. No need to criticize the first introduced function. I've updated the sentence similar to your suggestion.

README.md
maxitg: Well, I first describe what they do, and then how to use them. The argument structure is basically what the whole section is about after the `WolframModel` mention in the second paragraph. Do you think I should riffle the first two paragraphs with the rest?

README.md
maxitg: Good idea, I added a sentence where `SetReplaceList` is introduced.

README.md
sskini2: Can you put the image there without showing the input. And you can say you will talk about plotting them in detail later

README.md
sskini2: Fair enough

README.md
sskini2: I am not sure what is the best way to do this. I think giving the examples first is fine but it is probably overall argument structure at the end. Like say I am somebody who doesn't know WL very well, actually writing SetReplace[list, rule, options] might help me visualize how to do it.

README.md
```diff
@@ -157,6 +157,45 @@ Out[] = {{1, 2, 5, 3, 6}, {5, 3, 6, 3}, {6, 3, 8}, {8, 9}, {17}}
 
 All of these functions have `Method`, `TimeConstraint` and `"EventOrderingFunction"` options. `TimeConstraint` is self-evident, the other two work the same way as they do in `WolframModel` and will be described further in the `WolframModel` part of this README.
 
+## ToPatternRules
+
+`ToPatternRules` is a convenience function used to quickly enter rules such as the one mentioned previously
```
sskini2: I am finding it hard to understand what this is doing. Can you explain to me what this example is doing and perhaps I will have an idea what should be explained here?
In[] := ToPatternRules[{{v1, v2, v3}, {v2, v4, v5}} ->
  {{v5, v6, v1}, {v6, v4, v2}, {v4, v5, v3}}]
Out[] = {{v1_, v2_, v3_}, {v2_, v4_, v5_}} :>
 Module[{v6}, {{v5, v6, v1}, {v6, v4, v2}, {v4, v5, v3}}]

README.md
maxitg: I would prefer not to. For consistency, I want to keep the inputs for all images. I think it would be better to just start using `WolframModelPlot` right away. Ideally, I would mention that `WolframModelPlot` supports plotting other features of `WolframModel`s, which is why it's called this way, but of course it does not do that yet. Or, as a crazy idea, I can not use `WolframModelPlot` at all, and use `HypergraphPlot` everywhere, and adjust its usage message to not say it's deprecated.

README.md
```diff
@@ -196,6 +196,93 @@ Out[] = {{{v1_, v2_}} :> Module[{v3}, {{v1, v2}, {v2, v3}}], {{v1_, v2_}} :>
...
+* `"Generations"` shows how many layers of predecessors a given edge has.
+* `"Rules"` is an exact copy of the `WolframModel` input.
+* `"MaxCompleteGenerations"` shows the smallest generation such that the final state does not contain any matches composed solely of expressions from this generation. In this particular case, it is the same as the largest generation of any edge, but it might be different if a more elaborate [step specification](#step-limiters) is used.
+* `"TerminationReason"` shows the reason evaluation was stopped. See the [`"TerminationReason"`](#terminationreason) property for more details.
+* Finally, `"EventRuleIDs"` shows which rule was used for each event. It's rather boring in this particular case as only one rule is used in this example.
+
+A specific property can be requested from an evolution object in a similar way as a property for an `Entity`. The list of available properties can be found [below](#properties).
```
sskini2: Are we assuming people have WL experience? I think saying "similar way as a property for an Entity" is not going to help. People who know WL will intuitively get it but people who don't know WL might not

README.md
```diff
@@ -196,6 +196,93 @@ Out[] = {{{v1_, v2_}} :> Module[{v3}, {{v1, v2}, {v2, v3}}], {{v1_, v2_}} :>
...
+  6}, {7, 2, 9}, {7, 14, 6}, {14, 11, 4}, {11, 7, 8}}
+```
+
+If a property does not take any arguments, it can be specified directly in `WolframModel` as a shorthand:
+```
+In[] := WolframModel[{{1, 2, 3}, {2, 4, 5}} -> {{5, 6, 1}, {6, 4, 2}, {4, 5,
+    3}}, {{1, 2, 3}, {2, 4, 5}, {4, 6, 7}}, 10, "EdgeCountList"]
+Out[] = {3, 4, 6, 8, 12, 18, 24, 36, 54, 76, 112}
+```
+
```
sskini2: Are you explaining these properties thoroughly later? I think it doesn't clearly document what each property are and which ones has second arg and which one don't.

README.md
maxitg: Oh, I see what you are saying. I added `SetReplace[set, rule]` pseudo-input at the beginning. The same for `WolframModel`. I did not do that for other functions as they only have one (or two similar like `HypergraphUnifications`) argument.

README.md
maxitg: Well, `{{v1, v2, v3}, {v2, v4, v5}} -> {{v5, v6, v1}, {v6, v4, v2}, {v4, v5, v3}}` is the rule written in the ordinary `WolframModel` format (just with `v1`, `v2`, ... used instead of `1`, `2`, `3`, ...). It is converted to the pattern rule `{{v1_, v2_, v3_}, {v2_, v4_, v5_}} :> Module[{v6}, {{v5, v6, v1}, {v6, v4, v2}, {v4, v5, v3}}`, which is (almost) the kind of rule `SubsetReplace` is using, and is similar to normal `Replace` rules.

The second, pattern-rule format with the `Module`, is in a sense a "low-level", "power-user" format that allows for a lot of freedom (like choosing your own name for the new vertices, custom nesting, custom conditions on the vertices, etc.).

The first anonymous-rule format (which is equivalent to `{{1, 2, 3}, {2, 4, 5}} -> {{5, 6, 1}, {6, 4, 2}, {4, 5, 3}}`) is a simplified format, which is easier to read and enter, but does not allow for as much customization.

README.md
```diff
@@ -196,6 +196,93 @@ Out[] = {{{v1_, v2_}} :> Module[{v3}, {{v1, v2}, {v2, v3}}], {{v1_, v2_}} :>
...
+    3}}, {{1, 2, 3}, {2, 4, 5}, {4, 6, 7}}, 10]
+```
+![WolframModelBasicEvolution10](READMEImages/WolframModelBasicEvolution10.png)
+
+Note this call is different from using the `SetReplace` function in a variety of ways:
+* The order of arguments is switched, the rule goes first.
+* The rule is specified in the "anonymous" form (i.e., `ToPatternRules` is done implicitly).
+* The number of steps here is defined the same way as in `SetReplaceAll`, which is also known as the number of generations. Here each edge can have at most 10 generations of predecessors.
+* The output is not a final state, but instead an object which contains the entire evolution (similar to `SetReplaceList`) but with additional information about which rules are being used at each replacement. From the information field on that object one can see that the evolution was done for 10 generations (i.e., a fixed point has not been reached early), and 109 replacements (aka events) were made in total. More properties can be computed from an evolution object, more on that later.
+
```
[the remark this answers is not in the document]
maxitg: Well, not that soon, properties are still a few subsections away. I removed "more on that later", and added a link from "properties" to that subsection, that should make it clearer.

README.md
```diff
@@ -196,6 +196,93 @@ Out[] = {{{v1_, v2_}} :> Module[{v3}, {{v1, v2}, {v2, v3}}], {{v1_, v2_}} :>
...
+  "Generations" -> {0, 0, 0, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3},
+  "AtomLists" -> {{1, 2, 3}, {2, 4, 5}, {4, 6, 7}, {5, 8, 1}, {8, 4, 2}, {4, 5, 3},
+    {7, 9, 8}, {9, 6, 4}, {6, 7, 2}, {1, 10, 4}, {10, 8, 5}, {8, 1, 3}, {4, 11, 7},
+    {11, 6, 9}, {6, 4, 8}, {5, 12, 1}, {12, 8, 10}, {8, 5, 4}},
+  "Rules" -> {{1, 2, 3}, {2, 4, 5}} -> {{5, 6, 1}, {6, 4, 2}, {4, 5, 3}},
+  "MaxCompleteGeneration" -> 3, "TerminationReason" -> "MaxGenerationsLocal",
+  "EventRuleIDs" -> {1, 1, 1, 1, 1}|>]
+```
+
+The most important part of that association is `"AtomLists"` which includes all set elements (aka expressions or edges) ever created throughout history. Note, this does not correspond to any particular step, rather all steps are combined. They are not just catenated states as well, as if a particular expression was never used as an input for any replacement in a particular step, it would not be duplicated in that list. To see how that works, compare it to `"StatesList"` and observe that a catenated `"StatesList"` would contain more expressions than `"AtomLists"` does.
```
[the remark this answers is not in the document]
maxitg: Yes, I agree. Fixed (also everywhere else it appears).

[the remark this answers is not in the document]
maxitg: Done. I rewrote this explanation starting with where the edges came from and followed by explaining they don't appear twice in `"AtomLists"` if they appear in multiple steps (with an example).

README.md
maxitg: Removed reference to `Entity`.

README.md
maxitg: Absolutely! It's a significant chunk of this document (the Properties section). I added a more explicit reference to it right after the first example of a property.

README.md
```diff
@@ -1,67 +1,76 @@
...
-
-Here, `{{0, 1}, {1, 2}, {2, 3}}` is the initial condition, `{{a_, b_}, {b_, c_}} :> {{a, c}}` is the rule that matches pairs of edges sharing a vertex and replaces them with a single edge, and `2` the number of steps, i.e., the number of times the rule is applied.
-
-One can also imagine an inverse of the rule above
+In **SetReplace**, this can be expressed as (the new element is put at the end)
 ```
-In[.] := SetReplace[{{0, 1}}, {{a_, b_}} :>
-  Module[{new}, {{a, new}, {new, b}}], 2]
+In[] := SetReplace[{1, 2, 5, 3, 6}, {a_, b_} :> {a + b}]
+Out[] = {5, 3, 6, 3}
```
sskini2, line 23: At some point, you had a note that 1+2 = 3 is the last element of the list.. I think that note was very helpful

README.md
sskini2: I think this is good as is.. I am going to resolve this

README.md
maxitg: The note is still there, just before this input. I also now explicitly say that the new element is `1 + 2 -> 3`.

README.md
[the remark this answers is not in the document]
maxitg: I think you are looking at the outdated version... The new version does not say that anymore. It now says "... They are quite simple, and perform replacement operations either one-at-a-time ...".

PR #274, opened 2020-03-23
WolframModelPlot drawing optimizations

### LINKED ISSUE #240: WolframModelPlot optimization
`WolframModelPlot` is currently about 5 times slower than `GraphPlot`.

And while it does produce better results in many cases (i.e., arrow/vertex sizes are consistent, transparency is preserved for large graphs, etc.), its performance should be made more similar to that of `GraphPlot`.

maxitg:
## Changes

* Closes #240.
* Does three optimizations:
  * Speeds up drawing of arrowheads by (1) skipping them altogether is `"ArrowheadLength" -> 0`, and optimizing the computation of shape embeddings otherwise.
  * Fixes a weed by which `drawEmbedding` would still make a call to `GraphPlot` to obtain `VertexLabels` even if `None` were requested.
  * The largest optimization: streamlines `drawEmbedding` function by only including a style once in the result if no elementwise styles were specified.

## Commens

* Note that the largest optimization is not on the kernel side, but on the frontend side (the filesize of the images should decrease as well, although I did not check that).
* This is because the graphics objects produced are now much smaller, only contain element styles once, and don't have nested `Rotate`, `Translate`, and `Scale` to draw arrowheads. This makes drawing by the frontend significantly faster, however, `AbsoluteTiming` cannot detect that, so the timing needs to be measured manually.
* To get a timing in the status bar like I have in the examples below, set `EvaluationCompletionAction` to `"ShowTiming"` in front-end settings:

Screen Shot 2020-03-23 at 21 50 54

## Tests

* This example used to take 18 seconds, but now only takes 8.5 seconds:

```
In[] := state = WolframModel[{{1, 2, 3}, {4, 5, 6}, {1, 4}} -> {{2, 7, 8}, {3,
       9, 10}, {5, 11, 12}, {6, 13, 14}, {8, 12}, {11, 10}, {13, 
      7}, {14, 9}}, {{1, 1, 1}, {1, 1, 1}, {1, 1}, {1, 1}, {1, 1}}, 
   10, "FinalState"];
In[] := AbsoluteTiming[WolframModelPlot[state]]
```

Before:

Screen Shot 2020-03-23 at 19 15 34

After:

Screen Shot 2020-03-23 at 19 12 31

Even better performance can be achieved by setting `"ArrowheadLength" -> 0` (reasonable to do for large graphs, but not a default since we don't know what resolution or plot range the output image will be used for):

Screen Shot 2020-03-23 at 19 19 01

which is quite close to `GraphPlot`:

Screen Shot 2020-03-23 at 19 21 05

We can also use `"HyperedgeRendering" -> "Subgraphs"`,  but note this will produce a different embedding, as it would not close edges into polygons:

Screen Shot 2020-03-23 at 21 45 57

### DIFF
--- a/Kernel/WolframModelPlot.m
+++ b/Kernel/WolframModelPlot.m
@@ -454,9 +454,11 @@ addConvexPolygons[edgeType_][edge_, subgraphsShapes_] := Module[{points, region,
 
 (** Drawing **)
 
-applyStyle[style : Except[_List], shapes_] := {style, shapes}
+applyStyle[style : Except[_List], shapes_] := With[{trimmedShapes = DeleteCases[shapes, {}]},
+	If[trimmedShapes === {}, Nothing, {style, trimmedShapes}]
+]
 
-applyStyle[style_List, shapes_] := Transpose[{style, shapes}]
+applyStyle[style_List, shapes_] := Replace[DeleteCases[Transpose[{style, shapes}], {_, {}}], {} -> Nothing]
 
 drawEmbedding[
 			styles_,
--- a/Kernel/WolframModelPlot.m
+++ b/Kernel/WolframModelPlot.m
@@ -463,8 +463,8 @@ drawEmbedding[
 		embeddingShapes[[2]]], {$edgeLine, $edgePolygon, $edgePoint}][[2, All]];
 
 	(* would only work if coordinates consist of a single point *)
-	labels = If[VertexLabels === None,
-		Nothing,
+	labels = If[vertexLabels === None,
+		Graphics[{}],
 		GraphPlot[
 			Graph[embedding[[1, All, 1]], {}],
 			VertexCoordinates -> embedding[[1, All, 2, 1, 1]],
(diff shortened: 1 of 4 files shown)


maxitg:
Fixed the merge conflict with adaptive arrowhead sizes. `WolframModelPlot` is currently 2.5 times slower than non-looped `GraphPlot` in the case of binary edges, the target is 2 times. I will need to do further, less safe (in terms of layout consistency) optimizations. These will come in a separate PR in case we will need to revert them.

GetJonWithIt:
Merge conflict appears to have been resolved correctly...

maxitg:
> all tests successfully pass - approved to merge

@GetJonWithIt, I just want to point out that it's not necessary to check during the review if the tests passed, it is *impossible* to merge a PR without all tests passing, so it's completely reasonable, for example, to review in parallel with CI running. It is useful however to look at the code and unit tests, and check if there may be any issues the CI might miss.

PR #308, opened 2020-04-15
Add header cstdint to IDTypes providing int64_t

phcerdan:
To be self-consistent and include all the headers IDTypes.hpp needs.

I am guessing IDTypes.hpp is included
in a compilation unit where cstdint is included.

cpp-reference: https://en.cppreference.com/w/cpp/types/integer

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/308)

### DIFF
--- a/libSetReplace/IDTypes.hpp
+++ b/libSetReplace/IDTypes.hpp
@@ -1,6 +1,8 @@
 #ifndef IDTypes_h
 #define IDTypes_h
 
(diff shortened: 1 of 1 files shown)


maxitg:
Looks good to me, also checked compilation on Mac and Windows is not broken. Thanks for contributing! :tada:

PR #311, opened 2020-04-17
Reduced number of hash table lookups and memory allocations

aokellermann:
## Changes
* Optimized some cases of hash table lookups such that they occur less
* Reduced number of `std::vector::push_back` calls in favor of either `std::vector::reserve`+`std::vector::emplace_back` or `std::vector::resize`+`std::vector::operator[]` to reduce number of memory allocations
* A few other small changes, such as spelling/code style consistency

## Comments
* No functionality is changed

## Examples
The following is a basic example of reducing hash table lookups (worst case O(n) with size of container)

Before:
```c++
if (index_.count(atom)) {
    return index_.at(atom);
} else {
    return {};
}
```
After:
```c++
const auto resultIterator = index_.find(atom);
return resultIterator != index_.end() ? resultIterator->second : std::unordered_set<ExpressionID>();
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/311)

### DIFF
--- a/libSetReplace/Expression.cpp
+++ b/libSetReplace/Expression.cpp
@@ -49,11 +49,8 @@ namespace SetReplace {
         }
         
         const std::unordered_set<ExpressionID> expressionsContainingAtom(const Atom atom) const {
-            if (index_.count(atom)) {
-                return index_.at(atom);
-            } else {
-                return {};
-            }
+            const auto resultIterator = index_.find(atom);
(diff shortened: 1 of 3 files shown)


maxitg:
Notebook with performance testing: [AntonyOptimization.nb.zip](https://github.com/maxitg/SetReplace/files/4494516/AntonyOptimization.nb.zip)

maxitg:
Thanks a lot for contributing, by the way. 👍

PR #328, opened 2020-04-22
Cpp optimizations

aokellermann:
## Changes
* Increased usage of STL functions for efficiency.
* Reduced instances of pass-by-value to reduce amount of copying, particularly vectors. 
* Reduced amount of hash lookups (sets/maps). Although complexity is constant, lookups are quite expensive.
* Increased usage of foreach loops
* Some other misc. optimizations

Overall performance is increased by up to 37% in one test I ran of WolframModel.

## Comments
* No functionality changes.
* As a bonus, these commits include auto code formatting. This partially addresses #319. You may see a lot of erasures on blank lines. Many IDEs including mine erase white space to reduce file size.

## Examples
Testing setup functions:
```wl
Attributes[meanAroundTiming] = {HoldAll};
meanAroundTiming[expr_, measurementsCount_ : 5] := 
 MeanAround@Table[First@AbsoluteTiming[expr], measurementsCount]
```
```wl
speedupDelta[old_, new_] := UnitConvert[(old - new)/old, "Percent"]
```
```wl
run := meanAroundTiming[WolframModel[#1, Automatic, #2]] & @@@ {
   {{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, 10},
   {{{1, 2, 3}, {4, 3, 5}, {3, 6}} -> {{6, 7, 8}, {6, 9, 10}, {11, 8, 
       10}, {5, 2, 9}, {9, 9}, {1, 9}, {7, 5}, {8, 5}}, 13},
   {{{1, 2, 2}, {3, 2, 4}} -> {{5, 4, 4}, {4, 3, 5}, {3, 5, 1}}, 
    10000},
   {{{1, 2}, {2, 1}, {1, 3}, {2, 3}, {3, 1}, {3, 2}} -> {{1, 2}, {2, 
       1}, {1, 3}, {2, 3}, {3, 1}, {3, 2}, {1, 4}, {2, 4}, {3, 4}, {4,
        1}, {4, 2}, {4, 3}}, 4},
   {{{1}, {1}, {1}} -> {{1}, {1}, {1}, {1}}, 10}
   }
```
Running from master, 
```wl
master = run
```
yields
2020-04-22-203022_1920x1080_scrot
Running from this branch,
```wl
this = run
```
yields
2020-04-22-203908_1920x1080_scrot
To compare the two, running
```wl
speedupDelta[master, this]
```
yields
2020-04-22-180159_1920x1080_scrot

---
[...]

PR #330, opened 2020-04-29
Small cleanups

KorKozan:
## Changes
* changed the non-exported functions to have static linkage by putting them into an anonymous namespace
* cleaned up a few std::move pessimizations that were preventing copy elision.
* fixed the #include order (local files should be #included before the system ones)
* commented out unused parameter names

## Comments
* making the non-exported functions have static linkage reduces the binary size and allows further compiler optimizations (ie. inlining)

## Examples

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/330)

### DIFF
--- a/libSetReplace/SetReplace.cpp
+++ b/libSetReplace/SetReplace.cpp
@@ -1,21 +1,21 @@
+#include "SetReplace.hpp"
+#include "Set.hpp"
+
 #include <chrono>
 #include <random>
 #include <string>
 #include <unordered_map>
 
-#include "SetReplace.hpp"
+namespace SetReplace { namespace {
 
-#include "Set.hpp"
-
-mint getData(const mint* data, mint length, mint index) {
-    if (index >= length || index < 0) {
-        throw LIBRARY_FUNCTION_ERROR;
-    } else {
-        return data[index];
+    mint getData(const mint* data, mint length, mint index) {
+        if (index >= length || index < 0) {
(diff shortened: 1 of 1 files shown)


aokellermann:
> 
> 
> > *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/330)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @aokellermann and @KorKozan)
> 
> ___
> *[libSetReplace/SetReplace.cpp, line 188 at r1](https://reviewable.io/reviews/maxitg/setreplace/330#-M6N09H1-ki8kTIitZs7:-M6N09H24uChj0sIQEKr:bfdohwr) ([raw file](https://github.com/maxitg/setreplace/blob/deebe30e01b6a9511a20ae257ad51e0d5f72ceca/libSetReplace/SetReplace.cpp#L188)):*
> > ```cpp
> >     }
> > 
> >     int setDelete(WolframLibraryData /*libData*/, mint argc, MArgument *argv, MArgument /*result*/) {
> > ```
> 
> Why comment out `libData` and `result`? Same below.
> 
> 
> 
> 

The main reason this would be done is to suppress unused parameter compiler warnings I suspect.

maxitg:
Between #328 and #337, these changes are now in master, so I'm closing this PR. With that said, thanks for contributing, @KorKozan! Some of the suggestions from here were very useful, especially the idea to enable all warnings and warnings-as-errors, which has caught some questionable code already.

PR #339, opened 2020-05-13
CMakeLists.txt for libSetReplace

aokellermann:
## Changes
* Addition of CMakeLists.txt for libSetReplace.
* Grants ability to build libSetReplace without WL portion, in order to weed whack C++ code easier.
* Includes FindMathematica functionality for easy linking. 
* Includes all compiler warnings added in #337. 
* On CMake-based IDEs (eg. CLion), provides error highlighting for compiler warnings without having to compile.

## Comments
* This should be merged after #337, due to addition of compiler warnings.
* Included [FindMathematica](https://github.com/sakra/FindMathematica) is under MIT license.

## Examples
IDE error highlighting for compiler warning:
2020-05-13-122226_1920x1080_scrot

Manually compiling libSetReplace with CMake:
```bash
mkdir cmake-build-debug
cd cmake-build-debug
cmake ..
make
```
Example of error produced:
```
/home/antony/Drive/Programming/SetReplace/libSetReplace/Match.cpp:375:30: error: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Werror=sign-compare]
  375 |             if (nextInputIdx == -1) {
      |                 ~~~~~~~~~~~~~^~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [CMakeFiles/SetReplace.dir/build.make:96: CMakeFiles/SetReplace.dir/libSetReplace/Match.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:96: CMakeFiles/SetReplace.dir/all] Error 2
make: *** [Makefile:104: all] Error 2

```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/339)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -168,7 +168,7 @@ If you'd like to implement a small utility useful throughout the package, put it
 
 ### libSetReplace
 
-libSetReplace is the C++ library that implements the `"LowLevel"` method of [`WolframModel`](/Kernel/WolframModel.m). It lives in [`libSetReplace`](/libSetReplace) directory, and there is also the [Xcode project](/SetReplace.xcodeproj) for it. [`SetReplace.cpp`](/libSetReplace/SetReplace.cpp) and [`SetReplace.hpp`](/libSetReplace/SetReplace.hpp) implement the interface with Wolfram Language code.
+libSetReplace is the C++ library that implements the `"LowLevel"` method of [`WolframModel`](/Kernel/WolframModel.m). It lives in the [`libSetReplace`](/libSetReplace) directory, and there is also an [Xcode project](/SetReplace.xcodeproj) as well as [CMakeLists.txt](/CMakeLists.txt) for it. [`SetReplace.cpp`](/libSetReplace/SetReplace.cpp) and [`SetReplace.hpp`](/libSetReplace/SetReplace.hpp) implement the interface with Wolfram Language code.
 
 The C++ implementation keeps an index of all possible rule matches and updates it after every replacement. The reindexing algorithm looks only at the local region of the graph close to the rewrite site. Thus time complexity does not depend on the graph size as long as vertex degrees are small. The downside is that it has exponential complexity (both in time and memory) in the vertex degrees. Currently, it also does not work for non-local rules (i.e., rule inputs that do not form a connected hypergraph) and rules that are not hypergraph rules (i.e., pattern rules that have non-trivial nesting or conditions).
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ Build/
 # Built paclets
 
 *.paclet
+
+# CMake build directory
+
+cmake-build-debug/
--- a/CMake/Mathematica/FindMathematica.cmake
+++ b/CMake/Mathematica/FindMathematica.cmake
@@ -1,0 +1,4529 @@
+# - Try to find Mathematica installation and provide CMake functions for its C/C++ interface
+#
+# See the FindMathematica manual for usage hints.
+#
+#=============================================================================
+# Copyright 2010-2020 Sascha Kratky
+#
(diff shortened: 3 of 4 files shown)


phcerdan:
If you want a complete CMakeLists, have a look at my fork (providing python wrappings).
https://github.com/phcerdan/wolfram_model/blob/master/CMakeLists.txt

- Setting CMAKE_CXX_FLAGS in a regular variable is not a great idea (overrides the flags the user wants to provide). Use a CACHE variable instead.
- The Mathematica libraries are not used in the c++ libraries. Not sure about the utility of FindMathematica.
- GLOBing for getting source files is also not recommended.
- There are no install rules, nor config files (if any third-party project wants to use `SetReplace`).

I would be happy to merge my CMakeLists.txt in the fork, if you are interested.

maxitg:
> * The Mathematica libraries are not used in the c++ libraries. Not sure about the utility of FindMathematica.

I agree. While we might add callbacks to the Wolfram Kernel at some point, we need to make sure the main C++ code still compiles without it, so that we don't break things like @phcerdan's fork. And I would prefer to avoid introducing dependencies if not necessary. One problem though `SetReplace.cpp/hpp` would not show compiler warnings in CLion. @aokellermann, do you have any thoughts on that?

> * GLOBing for getting source files is also not recommended.
>
> * There are no install rules, nor config files (if any third-party project wants to use `SetReplace`).
>
>
> I would be happy to merge my CMakeLists.txt in the fork, if you are interested.

Yes, I think that's a good idea, no reason to unnecessarily duplicate. @aokellermann, could you check if compiler warnings show up correctly in @phcerdan's [wolfram_model fork](https://github.com/phcerdan/wolfram_model)? But @phcerdan, please hold on working on PR yet, until we make a decision on what to do here.

aokellermann:
This CMakeLists was only intended to provide syntax highlighting and libSetReplace-specific compilation (for error checking), not library headers or object. 

> > * The Mathematica libraries are not used in the c++ libraries. Not sure about the utility of FindMathematica.
> 
> I agree. While we might add callbacks to the Wolfram Kernel at some point, we need to make sure the main C++ code still compiles without it, so that we don't break things like @phcerdan's fork. And I would prefer to avoid introducing dependencies if not necessary. One problem though `SetReplace.cpp/hpp` would not show compiler warnings in CLion. @aokellermann, do you have any thoughts on that?

If including SetReplace.cpp/hpp in the compilation, FindMathematica is needed. The linker can't find WolframLibrary.h otherwise. @phcerdan doesn't use SetReplace.cpp/hpp, so he doesn't need FindMathematica.

> 
> > * GLOBing for getting source files is also not recommended.
> > * There are no install rules, nor config files (if any third-party project wants to use `SetReplace`).
> > 
> > I would be happy to merge my CMakeLists.txt in the fork, if you are interested.
> 
> Yes, I think that's a good idea, no reason to unnecessarily duplicate. @aokellermann, could you check if compiler warnings show up correctly in @phcerdan's [wolfram_model fork](https://github.com/phcerdan/wolfram_model)? But @phcerdan, please hold on working on PR yet, until we make a decision on what to do here.

I agree. @phcerdan's CMakeLists is a lot more complete than mine and even provides a library. I think it would make sense for @phcerdan to merge a modified version of his CMakeLists into this branch, without the python wrappings. In addition to the library target, a second target that includes SetReplace.cpp/hpp and links against WolframLibrary.h (using FindMathematica on this branch) should be created in order to provide syntax highlighting nad compilation capabilities.

@maxitg and @phcerdan, does this sound like a good plan to you?

maxitg:
> @maxitg and @phcerdan, does this sound like a good plan to you?

Sounds good to me. @phcerdan, could you split your `CMakeLists.txt` into the C++ and Python parts (I assume it's possible) and make a PR with the C++ part as @aokellermann suggests?

phcerdan:
Follow on #340

PR #340, opened 2020-05-14
Add CMake support

phcerdan:
Providing:
- target (library) `SetReplace`
- config files for third-parties CMake projects
to use SetReplace with no need of FindSetReplace.cmake
- Usable from build and install trees.
- Option to use with or without Mathematica.
- Infrastructure to perform cpp tests with gtest.

CMake options:

- `SET_REPLACE_BUILD_TESTING`:
Download gtest at build time (with FetchContent) and enable
basic cpp testing.
OFF by default.

- `SET_REPLACE_WITH_MATHEMATICA`:
Compiles `SetReplace.cpp` which provides the interface for Wolfram Language.
For this `FindMathematica.cmake` is needed.
`FindMathematica.cmake` is taken from:
https://github.com/sakra/FindMathematica/blob/master/CMake/Mathematica/FindMathematica.cmake
that seems used and up-to-date.
Please note that FindX.cmake is deprecated in CMake, and CONFIG files should be
provided by the project (Mathematica). But this is an upstream issue,
out of our control.
OFF by default.

- `SET_REPLACE_ENABLE_ALLWARNINGS`:
Enable a lot of extra warnings for GNU/Clang compilers. And treat them
as errors. Useful for CI and development.
OFF by default.

Closes #339

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/340)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -209,6 +209,51 @@ The C++ implementation keeps an index of all possible rule matches and updates i
 
 Every time the `"LowLevel"` implementation of [`WolframModel`](/Kernel/WolframModel.m) is called, an instance of class [`Set`](/libSetReplace/Set.hpp) is created. [`Set`](/libSetReplace/Set.hpp) in turn uses the [`Matcher`](/libSetReplace/Match.hpp) class to perform the matching of set elements to rule inputs. [This class](/libSetReplace/Match.cpp) is the core of *SetReplace*.
 
+#### Compile C++ library with CMake
+
+The `SetReplace` library can be used outside of Mathematica. It provides a CMake project for easy interaction with the C++ ecosystem.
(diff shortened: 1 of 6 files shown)


phcerdan:
Please note that I haven't test it with `SET_REPLACE_WITH_MATHEMATICA=ON`. But it should work if `FindMathematica.cmake` behaves as expected.

phcerdan:
I think I am going to add another target with the Mathematica interface (SetReplaceMathematica?), instead of adding to the existing core library (SetReplace). Incoming.

phcerdan:
Tested with `SET_REPLACE_WITH_MATHEMATICA=ON` after downloading WolframEngine for developers. Working.

aokellermann:
Thanks for contributing a full CMakeLists! This is very useful. I have just a few requests before you merge.

PR #356, opened 2020-05-14
performanceTest.wls checkout original SHA

aokellermann:
## Changes
* After completing tests on both input SHAs, performs `git checkout` on original git SHA.

## Examples
Originally, if running from `SHA_1`:
```bash
./performanceTest.wls SHA_2 SHA_3
```
would leave you on `SHA_3`.

With this change, running the same command would leave you on SHA_1.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/356)


aokellermann:
Example of weird behavior:
2020-05-14-205217_3000x2160_scrot

Now it's checking out an unrelated branch...

maxitg:
If we could extract the information about the current HEAD position from git (including a branch of any), and then check it back out, that would be the best approach. But I’m not sure how to do that.

PR #364, opened 2020-05-23
C++ multithread rules

aokellermann:
## Changes
* Adds multithreading for rule matching, currently using one thread per rule.
* Uses thread-safe techniques, such as locking.
* Uses [C++ threading library](https://en.cppreference.com/w/cpp/thread), so compilation on other systems shouldn't be an issue.
* This PR should speed up matching linearly with the number of input rules.

## Comments
This PR also fixes a bunch of linting errors that were introduced in #361 (not caused by this PR) and causing CI to fail the build. Some changes I made to address these errors:
* Fix all files' `#include`s.
* Add `-build/c++11` and `-build/c++14`, since `SetReplace` is a C++17 library.
* Add `CPPLINT.cfg` file in root to manage `cpplint` options easier.
* Add `extensions` option to `cpplint` to include `.hpp` files as well.

If you don't want to use `CPPLINT.cfg` and would prefer to put all the options in `lint.sh` instead, I can change that.

## Examples
Performance results on a 6-core machine:
```
Single-input rule                       -2.7 ± 0.5 %
Medium rule                             -0.9 ± 0.4 %
Sequential rule                         -0.5 ± 0.6 %
Large rule                              -0.06 ± 0.24 %
Exponential-match-count rule            -1.1 ± 0.6 %
CA emulator                             65.14 ± 0.17 %
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/364)


maxitg:
Performance results on a 6-CPU machine:
```
Single-input rule                       -2.7 ± 0.5 %
Medium rule                             -0.9 ± 0.4 %
Sequential rule                         -0.5 ± 0.6 %
Large rule                              -0.06 ± 0.24 %
Exponential-match-count rule            -1.1 ± 0.6 %
CA emulator                             65.14 ± 0.17 %
```

PR #366, opened 2020-05-26
CMake: Fix name of test file

phcerdan:
The file is named `Set_test.cpp`, changed from `SetTest.cpp`

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/366)

### DIFF
--- a/libSetReplace/test/CMakeLists.txt
+++ b/libSetReplace/test/CMakeLists.txt
@@ -1,6 +1,6 @@
 set(_link_libraries SetReplace ${GTEST_LIBRARIES})
 
-add_executable(SetTest SetTest.cpp)
-target_link_libraries(SetTest ${_link_libraries})
+add_executable(Set_test Set_test.cpp)
+target_link_libraries(Set_test ${_link_libraries})
 
-gtest_discover_tests(SetTest)
+gtest_discover_tests(Set_test)


aokellermann:
Fixed in #364.

phcerdan:
So, a simple fix is now dependent on a bigger PR like #364 ?

Why don't merge this, and rebase #364 on top of the new master?

In fact #364 has a lot of things that seems unrelated to the purpose of its title:
```
    Rename Set_test.cpp to SetTest.cpp to follow same pattern as the rest of the files (it also wouldn't compile due to discrepency with CMake file.
    Fix all files' #includes.
    Add -build/c++11 and -build/c++14, since SetReplace is a C++17 library.
    Add CPPLINT.cfg file in root to manage cpplint options easier.
    Add extensions option to cpplint to include .hpp files as well.
```

I understand that is hard to work in interdependent parallel branches, but shoving things into one big PR makes things harder to review, and slows down development.

maxitg:
I agree we should merge this separately. Our general policy is to [keep PRs as small as possible](https://github.com/maxitg/SetReplace/blob/master/.github/CONTRIBUTING.md#opening-a-pull-request) (although I have to admit, I sometimes struggle with that rule myself).

phcerdan:
> Looks good (I assume you tried to compile it to make sure it works).

It does work. Thanks

#364 renames the file `Set_test.cpp` to `SetTest.cpp`, instead of modifying the `CMakeLists.txt`. I think we changed the name because of constraints in `cpplint`.

PR #367, opened 2020-05-26
Local Multiway System Proposal

### LINKED ISSUE #334: Local multiway system description
Before implementing the local multiway system, I need to describe what it's going to do, which is just going to be a .md document.

This should mention problems with the current `MultiwaySystem`, the explanation of how the local multiway system will solve them. It should also explain the different options that will be available to control evolution and the implementation milestones.

maxitg:
## Changes
* Closes #334.
* Adds a bulletin explaining how the local multiway system works, and showing some examples for both match-all and purely spacelike multiway systems.

## Comments
* This is a "living" document, and it will be improved (in future PRs), in particular, once we have a proper implementation of the local multiway system in `master`.
* The purpose of this pull request is essentially a peer-review of this bulletin. Feel free to leave comments for both the physics content, the system description, as well as small/technical comments about the explanation.
* If you are not a *SetReplace* collaborator, you can still comment below.

## Examples
* All examples are in the bulletin itself. Here is the [readable version](https://github.com/maxitg/SetReplace/blob/localMultiwayProposal/Research/LocalMultiwaySystem/LocalMultiwaySystem.md).

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/367)


maxitg:
This is merged into #390.

Issue #335, opened 2020-05-11
Match-all local multiway system


maxitg:
The simplest version of the local multiway system to implement is the one that matches any subsets of expressions, even if they are not spacelike separated.

This is simple to do because it does not require any testing of relationships between edges. Of course, this breaks current evolution, so it should be controlled with an option.

While this might not seem useful, it's an interesting model in itself, which allows interference between branches by applying branch-merging events.

The implementation requires two steps:
- [x] Add an option in the C++ code to disable removing edges after each event, and instead, remove only a single match that was evaluated. Note, this will not cause duplicate events, because new matches indexed must involve a just created edge.
- [ ] Change Wolfram Language code to deal with cases where the same expression (edge) could have multiple destroyer events. This requires modifying the structure of the evolution object (think about compatibility with old objects), and doing something with properties like `"StatesList"`.

PR #368, opened 2020-05-27
EventSelectionFunction in libSetReplace

### LINKED ISSUE #372: "EventsList" does not preserve the order of inputs
### The problem
One would expect the list of input edge indices in `"EventsList"` to correspond to the same order as specified in the rule itself. However, it is not the case:
```wl
In[] := WolframModel[{{1, 2}, {2, 3, 4}} -> {{1, 2, 3, 4}}, {{2, 3, 4}, {1, 
   2}}, "EventsList"]
Out[] = {{1, {1, 2} -> {3}}}
```
The output here should actually be `{{1, {2, 1} -> {3}}}`.

If we substitute the edge indices in the output above with edge expressions, one can see the order is not preserved:
```wl
In[] := With[{evolution = 
   WolframModel[{{1, 2}, {2, 3, 4}} -> {{1, 2, 3, 4}}, {{2, 3, 4}, {1,
       2}}]}, evolution["EventsList"][[1, 2]] /. 
  n_Integer :> evolution["AllExpressions"][[n]]]
Out[] = {{2, 3, 4}, {1, 2}} -> {{1, 2, 3, 4}}
```

### Expected behavior
The output should be:
```wl
In[] := WolframModel[{{1, 2}, {2, 3, 4}} -> {{1, 2, 3, 4}}, {{2, 3, 4}, {1, 
   2}}, "EventsList"]
Out[] = {{1, {2, 1} -> {3}}}
```

### Version
```wl
```
image

```wl
In[] := $SetReplaceGitSHA
```
9620568bf18c009a472399a27f0bc546dfeb8fe0

### Additional context
This is tricky to fix, as the information about the order of inputs is not returned from `libSetReplace`. We only know the creator/destroyer events for each edge, which is not sufficient. A way to fix that would be to switch to event-based data structure for the evolution object, where the inputs/outputs for each event will be returned explicitly.

This will be fixed in version 2 of the evolution object. We just need to make both LowLevel and Symbolic code to use it:
- [ ] Create evolution object version 2 in `Method -> "LowLevel"`;
- [ ] Create evolution object version 2 in `Method -> "Symbolic"`.

maxitg:
## Changes
* Partially implements #335.
* Adds `EventSelectionFunction` to `libSetReplace`. Two are supported at the moment:
  * `GlobalSpacelike`: default singleway evolution;
  * `None`: match-all multiway system.
* `libSetReplace` fully supports both now, however, WL code always calls `GlobalSpacelike` (for now).
* Refactors the causal relationships tracking in `libSetReplace`. Specifically, the evolution history is now events-based instead of expressions-based, and is tracked by the `CausalGraph` class outside of `Set`:
  * In my opinion, this is a better approach. Besides addressing the performance problem which would otherwise arise from tracking multiple destroyer events, it lets us fix #372, and makes it easier to deal with things like partially empty events.
  * Currently, there is nothing algorithmically complex in it, but eventually, it will track the spacelike/branchlike/timelike relationships between edges/events, which is quite complicated, and will require a nontrivial data structure.
  * WL code still uses expressions-based structure for the evolution object, and [converts](https://github.com/maxitg/SetReplace/blob/ba7803bc9ff9a339631860678e63c8e288cebbb6/Kernel/setSubstitutionSystem%24cpp.m#L163) to it immediately after receiving the object from `libSetReplace`. That will be changed in a future PR.
* Note, this ***breaks the API of `libSetReplace`***. Constructor for `Set` and `SetExpression` members are changed. Not marked as breaking in the PR because it does not break the Wolfram Language API.

## Comments
* The only part that is not pure refactoring is [here](https://github.com/maxitg/SetReplace/blob/ba7803bc9ff9a339631860678e63c8e288cebbb6/libSetReplace/Set.cpp#L109), but it's not currently accessible from WL.
* `None` event selection function only deletes the match that was used but leaves all the input expressions in the set, allowing them to be matched again.
* Tests will be implemented in a future PR (in Wolfram Language).
* There is a slight dip in performance for simple-to-match rules, but it is caused by [`decodeExpressions`](https://github.com/maxitg/SetReplace/blob/ba7803bc9ff9a339631860678e63c8e288cebbb6/Kernel/setSubstitutionSystem%24cpp.m#L163), the function that converts the evolution object from events- to expressions-based representation. Since this function will be removed in a future PR, this performance dip is temporary.

## Examples
* No examples yet because no functionality is changed in the Wolfram Language. There are examples using slightly different code but implementing the same system [here](https://github.com/maxitg/SetReplace/blob/localMultiwayProposal/Research/LocalMultiwaySystem/LocalMultiwaySystem.md).
* Performance results:
```
Single-input rule                       -19.8 ± 2.6 %
Medium rule                             -5.0 ± 0.8 %
Sequential rule                         -4.9 ± 1.2 %
Large rule                              -0.4 ± 0.4 %
Exponential-match-count rule            -0.8 ± 1.7 %
CA emulator                             -1.0 ± 1.4 %
```
* Disabling `decodeExpressions` 98f7a15b03cca65e3e0135ed8255453cd6d4f5f3 we get:
```
Single-input rule                       -9. ± 4. %
Medium rule                             0.2 ± 1.0 %
Sequential rule                         -1.3 ± 1.1 %
Large rule                              -0.04 ± 0.30 %
Exponential-match-count rule            0.9 ± 0.8 %
CA emulator                             0.3 ± 1.7 %
```
* Run `gtest`:
```
➜  mkdir build && cd build && cmake .. -DSET_REPLACE_BUILD_TESTING=YES && make && make test
Running tests...
Test project /Users/maxitg/Development/SetReplace/build
    Start 1: Set.globalSpacelike
1/2 Test #1: Set.globalSpacelike ..............   Passed    0.00 sec
    Start 2: Set.matchAllMultiway
2/2 Test #2: Set.matchAllMultiway .............   Passed    0.00 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =   0.01 sec
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/368)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ message(STATUS "SET_REPLACE_WITH_MATHEMATICA: ${SET_REPLACE_WITH_MATHEMATICA}")
 message(STATUS "SET_REPLACE_EXTRA_COMPILE_OPTIONS: ${SET_REPLACE_EXTRA_COMPILE_OPTIONS}")
 
 set(libSetReplace_headers
+    Event.hpp
     Rule.hpp
     IDTypes.hpp
     Expression.hpp
@@ -52,6 +53,7 @@ set(libSetReplace_headers
     Set.hpp
     )
 set(libSetReplace_sources
+    Event.cpp
     Expression.cpp
     Match.cpp
     Set.cpp
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -28,6 +28,7 @@ $cpp$setCreate = If[$libraryFile =!= $Failed,
 		"setCreate",
 		{{Integer, 1}, (* rules *)
 			{Integer, 1}, (* initial set *)
+			Integer, (* event selection function *)
 			{Integer, 1}, (* ordering function index, forward / reverse, function, forward / reverse, ... *)
 			Integer}, (* random seed *)
 		Integer], (* set ptr *)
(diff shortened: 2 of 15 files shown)


maxitg:
> Looks good to me. I like having an explicit CasualGraph there tracking the events.
> As a side note, have you thought about testing the API with a simple cpp test?

I added missing files to `CMakeLists`, fixed the old test for the API, and added a simple test for `EventSelectionFunction::None`.

maxitg:
I just realized working on the WL side of this that it's much more convenient to return the initial event from the `CausalGraph`. I.e., the list of all expressions up to a given generation can then simply be obtained by catenating all outputs for rules until that generation. (And otherwise, one has to manually identify initial expressions). So, I'm going to change that here.

maxitg:
> So, I'm going to change that here.

Done.

PR #380, opened 2020-06-16
Event-based evolution object

### LINKED ISSUE #372: "EventsList" does not preserve the order of inputs

maxitg:
## Changes
* Resolves #372.
* Adds the `"Version"` key to the evolution object, so that the versions can be tracked and migrated.
* Changes the evolution object from expression-based to event-based data format.
* Implements migration from version 1 to version 2. A message is printed to alert the user about #372.

## Comments
* The new keys are `"Version"`, `"Rules"`, `"MaxCompleteGeneration"`, `"TerminationReason"`, `"AtomLists"`, `"EventRuleIDs"`, `"EventInputs"`, `"EventOutputs"`, and `"EventGenerations"`.
* Because we now explicitly track inputs for each event, `"EventList"` property can now return them in the correct order.
* Symbolic evolution is also changed to use event-based data structures internally.

## Examples
* The weed #372 mentioned previously is now fixed. The event inputs are returned in the correct order:
```wl
In[] := WolframModel[{{1, 2}, {2, 3, 4}} -> {{1, 2, 3, 4}}, {{2, 3, 4}, {1, 
   2}}, "EventsList"]
Out[] = {{1, {2, 1} -> {3}}}
```
* Check the structure of the new evolution object:
```wl
In[] := WolframModel[{{1, 2, 3}, {2, 4, 5}} -> {{5, 6, 1}, {6, 4, 2}, {4, 5, 
     3}}, {{1, 2, 3}, {2, 4, 5}, {4, 6, 7}}, 3][[1]]
Out[] = <|"Version" -> 2, 
 "Rules" -> {{1, 2, 3}, {2, 4, 5}} -> {{5, 6, 1}, {6, 4, 2}, {4, 5, 
     3}}, "MaxCompleteGeneration" -> 3, 
 "TerminationReason" -> "MaxGenerationsLocal", 
 "AtomLists" -> {{1, 2, 3}, {2, 4, 5}, {4, 6, 7}, {5, 8, 1}, {8, 4, 
    2}, {4, 5, 3}, {7, 9, 8}, {9, 6, 4}, {6, 7, 2}, {1, 10, 4}, {10, 
    8, 5}, {8, 1, 3}, {4, 11, 7}, {11, 6, 9}, {6, 4, 8}, {5, 12, 
    1}, {12, 8, 10}, {8, 5, 4}}, "EventRuleIDs" -> {0, 1, 1, 1, 1, 1},
  "EventInputs" -> {{}, {1, 2}, {5, 3}, {6, 4}, {7, 8}, {10, 11}}, 
 "EventOutputs" -> {{1, 2, 3}, {4, 5, 6}, {7, 8, 9}, {10, 11, 
    12}, {13, 14, 15}, {16, 17, 18}}, 
 "EventGenerations" -> {0, 1, 2, 2, 3, 3}|>
```
* Convert the old object to the new object:
```wl
In[] := WolframModelEvolutionObject[<|
   "CreatorEvents" -> {0, 0, 0, 0, 1, 2, 3}, 
   "DestroyerEvents" -> {1, 1, 2, 2, 3, 3, Infinity}, 
   "Generations" -> {0, 0, 0, 0, 1, 1, 2}, 
   "AtomLists" -> {{1, 2}, {2, 3}, {3, 4}, {4, 5}, {1, 3}, {3, 5}, {1,
       5}}, "Rules" -> {{1, 2}, {2, 3}} -> {{1, 3}}, 
   "MaxCompleteGeneration" -> 2, "TerminationReason" -> "FixedPoint", 
   "EventRuleIDs" -> {1, 1, 1}|>][[1]]

During evaluation of In[] := WolframModelEvolutionObject::migrationInputOrdering:
  Migrating the evolution object from version 1.
  The ordering of event inputs may be incorrect.

Out[] = <|"Version" -> 2, "Rules" -> {{1, 2}, {2, 3}} -> {{1, 3}}, 
 "MaxCompleteGeneration" -> 2, "TerminationReason" -> "FixedPoint", 
 "AtomLists" -> {{1, 2}, {2, 3}, {3, 4}, {4, 5}, {1, 3}, {3, 5}, {1, 
    5}}, "EventRuleIDs" -> {0, 1, 1, 1}, 
 "EventInputs" -> {{}, {1, 2}, {3, 4}, {5, 6}}, 
 "EventOutputs" -> {{1, 2, 3, 4}, {5}, {6}, {7}}, 
 "EventGenerations" -> {0, 1, 1, 2}|>
```
* Also, the performance regression introduced in #368 is now fixed:
```
➜  SetReplace git:(eventBasedEvolutionObject) ./performanceTest.wls 
...
Single-input rule                       20.9 ± 1.4 %
Medium rule                             16.2 ± 0.9 %
Sequential rule                         13.7 ± 1.6 %
Large rule                              0.5 ± 0.4 %
Exponential-match-count rule            5.0 ± 2.4 %
CA emulator                             2.3 ± 1.4 %
```
```
➜  SetReplace git:(eventBasedEvolutionObject) ./performanceTest.wls b478ebb461d040e073bb9f13de4656106e269b35
...
Single-input rule                       6.8 ± 1.3 %
Medium rule                             10.0 ± 0.8 %
Sequential rule                         2.8 ± 0.7 %
Large rule                              1.4 ± 0.8 %
Exponential-match-count rule            0.0 ± 0.9 %
CA emulator                             0.5 ± 2.1 %
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/380)

### DIFF
--- a/Kernel/setSubstitutionSystem$wl.m
+++ b/Kernel/setSubstitutionSystem$wl.m
@@ -120,16 +120,16 @@ setReplace$wl[set_, rules_, stepSpec_, vertexIndex_, returnOnAbortQ_, timeConstr
 		TimeConstrained[
 			CheckAbort[
 				{FixedPoint[
-					AbortProtect[Module[{newResult, deletedExpressions, ruleIndices},
+					AbortProtect[Module[{newResult, deletedExpressions, events},
 						If[eventsCount++ == Lookup[stepSpec, $maxEvents, Infinity],
 							Throw[{previousResult, $maxEvents}, $$setReplaceResult];
 						];
-						{newResult, {deletedExpressions, ruleIndices}} = Reap[
+						{newResult, {deletedExpressions, events}} = Reap[
 							Catch[
 								Replace[#, normalRules],
 								$$reachedAtomDegreeLimit,
 								Throw[{previousResult, $maxFinalVertexDegree}, $$setReplaceResult] &],
-							{$$deletedExpressions, $$ruleIndex}];
+							{$$deletedExpressions, $$events}];
 						If[vertexCount[vertexIndex] > Lookup[stepSpec, $maxFinalVertices, Infinity],
 							Throw[{previousResult, $maxFinalVertices}, $$setReplaceResult];
 						];
@@ -137,7 +137,7 @@ setReplace$wl[set_, rules_, stepSpec_, vertexIndex_, returnOnAbortQ_, timeConstr
 							Throw[{previousResult, $maxFinalExpressions}, $$setReplaceResult];
 						];
 						Map[Sow[#, $$deletedExpressions] &, deletedExpressions, {2}];
-						Map[Sow[#, $$ruleIndex] &, ruleIndices, {2}];
+						Map[Sow[#, $$events] &, events, {2}];
 						previousResult = newResult]] &,
 					List @@ set], $fixedPoint},
 				If[returnOnAbortQ,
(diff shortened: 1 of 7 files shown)


Kernel/setSubstitutionSystem$wl.m
```diff
@@ -348,42 +346,41 @@ All operations here should evaluate in O(1).*)
...
-				If[MissingQ[possibleInfinityResult],
-					possibleInfinityResult,
-					Min[possibleInfinityResult, intermediateEvolution["TotalGenerationsCount"]]]],
-			$terminationReason -> Replace[
-				intermediateEvolution[[1, Key[$terminationReason]]],
-				$fixedPoint ->
-					If[intermediateEvolution["TotalGenerationsCount"] == Lookup[stepSpec, $maxGenerationsLocal, Infinity],
-						$maxGenerationsLocal,
-						$fixedPoint]]|>]]
+		$maxCompleteGeneration -> If[MissingQ[maxCompleteGenerationAssumingExceedingStepSpecGenerationsDontExist],
```
delaix, line 372: Now _that_ is a variable name...

maxitg: It's too obscure. I tried to come up with a better name. Suggestions are welcome.

Also, please [use Reviewable](https://github.com/maxitg/SetReplace/blob/master/.github/CONTRIBUTING.md#code-review) in the future (link on the bottom of the PR message). It makes it easier to keep track of comments.

delaix: That was mostly a statement of amusement, though as you notice a name like that does tend to suggest you stuffed to much meaning into a single variable. It happens.

maxitg:
> I will say that more smaller PR's are better than fewer larger ones. Were all of the changes included in this PR necessary or could they have been staged in several increments?

I [agree](https://github.com/maxitg/SetReplace/blob/master/.github/CONTRIBUTING.md#opening-a-pull-request) with you on that. Believe it or not, I did that already (the [original PR](https://github.com/maxitg/SetReplace/pull/381) was ~3000 lines and is now split in three parts).

It introduces the new format for the evolution object, which is used in WL evolution code, and all properties of the evolution object, which is why this PR is so large.

I can split off the WL evolution (changes to `setSubstitutionSystem$wl.m`, 137 lines) into a separate PR if you insist, but this PR will still be quite large unless I implement the converter from version 2 to version 1 of the evolution object and reuse the old code for computing properties.

Do you think we should have a smaller limit on the PR size than the current limit of [1000 lines](https://github.com/maxitg/SetReplace/blob/master/.github/CONTRIBUTING.md#opening-a-pull-request)?

delaix:
There’s no point forcing it. As long as you are aware of the issue and trying to keep PR’s compact then you are doing what you can. Sometimes big PR’s are just a necessity.

PR #381, opened 2020-06-16
EventSelectionFunction

### LINKED ISSUE #335: Match-all local multiway system

maxitg:
## Changes
* Closes #335.
* Adds `"EventSelectionFunction"` option, which allows one to evaluate multiway systems.
  * `"EventSelectionFunction" -> "GlobalSpacelike"` is the default, which evaluates the singleway spacelike system, same as before.
  * `"EventSelectionFunction" -> None` evaluates the match-all system, i.e., all matches are turned into events indiscriminately, even if the input edges are branchlike or timelike.

## Comments
* Almost half of this PR has to do with testing how all evolution object properties behave when multiple destroyer events are present for a single expression.
* The other big chunk, `eventSelectionFunction.wlt`, is the tests for various matching situations verifying both `"GlobalSpacelike"` and match-all systems work as expected.
* Note, `"EventSelectionFunction"` is disabled for `SetReplace*` functions because they always return the `"FinalState"`, which would not work for multiway systems.
* Only `"LowLevel"` implementation supports the option, `"Symbolic"` one does not.
* The rest of the PR consists of passing the option value through to the C++ code, ensuring things like `Automatic` steps still work, and the error messages are produced if, i.e., `"MaxVertices"` is requested as a step spec.

## Examples
* Run the evolution that matches branchlike edges in a single event:
```
In[] := WolframModel[{{{1, 2}, {2, 3}} -> {{1, 2, 3}}, {{1, 2, 3}, {1, 2, 
     4}} -> {{1, 2, 3, 4}}}, {{1, 2}, {2, 3}, {2, 
   4}}, Infinity, "AllEventsEdgesList", 
 "EventSelectionFunction" -> None]
Out[] = {{1, 2}, {2, 3}, {2, 4}, {1, 2, 3}, {1, 2, 4}, {1, 2, 3, 4}, {1, 2, 4,
   3}}
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/381)


maxitg:
> Not like I'd know if this is correct or not...

Ok, I'll try to find someone else to look at it.

ToddRowland30:
Maybe this is correct.  sameGraphQ will remain unevaluated if the vertices are symbolic, or some other case where Equal stays unevaluated.  I think that it has the right behavior, and it seems the function is only used on tests with integer vertices, but just wanted to mention it.

maxitg:
> Maybe this is correct. sameGraphQ will remain unevaluated if the vertices are symbolic, or some other case where Equal stays unevaluated. I think that it has the right behavior, and it seems the function is only used on tests with integer vertices, but just wanted to mention it.

Good catch! It will indeed work correctly because the test will fail if the result is not `True`, but `SameQ` there would be clearner, so I changed it.

Issue #342, opened 2020-05-14
Test if a pair of expressions is timelike, spacelike or branchlike


maxitg:
Requires #335.

### The problem
The match-all local multiway system matches all groups of expressions including timelike and branchlike ones, which is different from the behavior of both the normal `WolframModel` and [`MultiwaySystem`](https://resources.wolframcloud.com/FunctionRepository/resources/MultiwaySystem).

While it's interesting on its own, we should be able to reproduce the normal evolution in the local multiway system. As a first step, we need a way to identify if a pair of expressions (edges) is timelike, spacelike or branchlike.

### Possible solution
Expressions are timelike if there is a path between their creator events in the causal graph. They are spacelike if the common ancestor is an event, and they are branchlike is their common ancestor is an edge.

Extra credit for implementing a data structure to do this efficiently in one shot, although there is a special issue for that: #344.

### Additional context
It is sufficient to have a C++ implementation for now, not exposed to WL, as communicating this without detached evolution objects [TODO: add issue] would be difficult.

PR #394, opened 2020-08-14
ExpressionsSeparation property

maxitg:
## Changes
* Partially implements #342 (it's not currently accessible from `libSetReplace`, which will be necessary for, i.e., spacelike multiway system).
* Introduces a new property, `"ExpressionsSeparation"`, that computes separation (`"Spacelike"`, `"Branchlike"`, `"Timelike"`, or `"Identical"`) between two expressions.

## Comments
* This works fine for a pair of expressions, but it becomes highly inefficient for computing matrices for large groups. A better implementation will be needed for that (in a separate PR).

## Examples
* Take an expressions-events graph:
```wl
In[] := WolframModel[<|
   "PatternRules" -> {{{v, i}} -> {{v, 1}, {v, 2}}, {{v, 1}} -> {{v, 
        1, 1}, {v, 1, 2}}, {{v, 1, 1}, {v, 2}} -> {{v, f, 1}}, {{v, 1,
         2}, {v, 2}} -> {{v, f, 2}}}|>, {{v, i}}, Infinity, 
  "EventSelectionFunction" -> None]["ExpressionsEventsGraph", 
 VertexLabels -> Placed["Name", After]]
```
image

* And compute separations for all pairs of edges:
```wl
In[] := ArrayPlot[
 Table[WolframModel[<|
     "PatternRules" -> {{{v, i}} -> {{v, 1}, {v, 2}}, {{v, 1}} -> {{v,
           1, 1}, {v, 1, 2}}, {{v, 1, 1}, {v, 2}} -> {{v, f, 1}}, {{v,
           1, 2}, {v, 2}} -> {{v, f, 2}}}|>, {{v, i}}, Infinity, 
    "EventSelectionFunction" -> None]["ExpressionsSeparation", m, 
   n], {m, 7}, {n, 7}], 
 ColorRules -> {"Branchlike" -> 
    WolframPhysicsProjectStyleData["BranchialGraph", "EdgeStyle"], 
   "Spacelike" -> 
    WolframPhysicsProjectStyleData["SpatialGraph", "EdgeLineStyle"], 
   "Timelike" -> 
    WolframPhysicsProjectStyleData["CausalGraph", "EdgeStyle"], 
   "Identical" -> White}]
```
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)


taliesinb (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)* status: all files reviewed, 6 unresolved discussions (waiting on @maxitg)

___
*[README.md, line 928 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MFa7-MAC1ebH_-ejCkb:-MFa7-MAC1ebH_-ejCkc:bosdd1b) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L928)):*
> ```Markdown
> `"CausalGraph"`, `"LayeredCausalGraph"` and `"ExpressionsEventsGraph"` properties all accept [`Graph`](https://reference.wolfram.com/language/ref/Graph.html) options, as was demonstrated above with [`VertexLabels`](https://reference.wolfram.com/language/ref/VertexLabels.html). Some options have special behavior for the [`Automatic`](https://reference.wolfram.com/language/ref/Automatic.html) value, i.e., `VertexLabels -> Automatic` in `"ExpressionsEventsGraph"` displays the contents of expressions, which are not the vertex names in that graph (as there can be multiple expressions with the same contents).
> 
> #### Edge Separations
> ```

You interchangeably use the word 'edge' and 'expression'. I *really* think it should be consistent, and that 'edge' is a very bad choice because of how many different graphs are involved, for which the word 'edge' is both natural and commonplace. When I need to talk about the edges of the causal graph, how am I supposed to refer to it? Or an edge in the expression-event graph? 
___
*[README.md, line 934 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MF_pvGe0OgpkqaN_Kag:-MF_pvGe0OgpkqaN_Kah:b-1erzks) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L934)):*
I would use "i.e." when the condition "they are produced as multiple outputs of a single event" is one of many possible conditions that produce spacelike separation. But isn't it the only, *defining* condition? In that case, you should avoid the word "i.e.", and probably drop the brackets since the definitions are important, not an afterthought. 

Here's what I suggest:

> 1. spacelike -- the edges were produced (directly or indirectly) by a single event 
> 2. branchlike -- the edges were produced (directly or indirectly) by multiple events that matched the same edge
> 3. timelike -- one edge produced or was produced by another edge
___
*[README.md, line 947 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MF_r9FK0an9HEsyytzT:-MF_r9FK0an9HEsyytzU:bb37spg) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L947)):*
> ```Markdown
>   {{{1, 2}} -> {{2, 3}}, {{1, 2}} -> {{3, 4}}},
>   {{{1, 2}} -> {{2, 3}}, {{2, 3}} -> {{3, 4}}}}
> ```
> ```

It would be nice to highlight the two edges that are being compared in these three cases, unless that involves too much extra code.
___
*[README.md, line 952 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MF_rHpx46B6zMocy0T8:-MF_rHpyDPDEZ7Ht4wfo:bwgud0s) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L952)):*
This grammar is a bit confusing. Suggest: "It is not always the case that spacelike separation refers to ..." 

But more importantly, the way you are expression the point is also confusing (to me). I have understood it now, but it took a while to get what you were trying to say. I would suggest this, which makes the point more directly by declaring upfront what the main take-home message is:

> One might be tempted to assume that spacelike separated expressions can always be 'assembled' to produce a possible history for a singleway system. For match-all evolution, however, this is not the case. Match-all rules can match two branchlike-separated expressions, something that is never possible for singleway systems. If such events produce spacelike-separated results, then we will obtain spacelike separated expressions that can be assembled into global states which *do not* correspond to any singleway evolution state. See the following picture:
___
*[README.md, line 989 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MF_rfxyAVSYAmFbZKuL:-MF_rfxyAVSYAmFbZKuM:btk8tze) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L989)):*
> ```Markdown
> 4. If the boundary contains either `A` and `B`, they are timelike separated.
> 5. If any vertices on the boundary are expression-vertices, they are branchlike separated.
> 6. Otherwise, if all vertices on the boundary are event-vertices, they are spacelike separated.
> ```

For me, it is more natural to think of expression-event graph as a "family tree" (there is of course a fair amount of inbreeding, so European royalty is a good model). 

To make this precise, think about the EE graph as a partial order, with `a > b` if `a` is an ancestor of `b` (this choice makes these graph plots in which time goes downwards into Hasse diagrams). 

Let's then define the "meet" and "join" operations as *set*-valued operations:

1. `⋀(a, b) = min {z; (z ≥ a) & (z ≥ b)}`
2. `⋁(a, b) = max {z;  (z ≤ a) & (z ≤ b)}`

In other words, ⋀ finds the set of 'parents', and ⋁ the set of 'children' of two nodes. These are the *extremal* such elements, because the `min` (`max`) remove elements that are larger (smaller) than any other elements in the set (recall that partial orders turn `min` and `max` into set-valued operations). 

In mathematical terms, this isn't lattice theory, because the meet and join are *sets*, they are not single elements. Maybe there is some math structure that models this, it would be nice to know!

Anyway, it is now very easy to define separation if we have these operations:

1. `a` and `b` are timelike separated ⇔ `a|b ∈ ⋀(a, b)`
2. `a` and `b` are branchlike separated ⇔ `e ∈ ⋀(a, b)`, for some expression `e`
3. `a` and `b` are spacelike separated otherwise

Anyway, my understanding is that this is identical to what your algorithm. Is that correct?

A side note about notation: it would be nice to establish `⋀` and `⋁` as symbols that we can use in long discussions to avoid words. Notation is very important, since it makes it easy to talk about the important things without having to use complex phrases. Anyway, we should also have functions that return the set of all ancestors, and all descendants, I suggest `⭡(a)` and `⭣(b)` for this. 

Similarly, for separation I suggest the following compact notation:

1. `a ⋮ b` timelike
2. `a ⋯ b` spacelike
3. `a ⋱ b` branchlike

Thank god for Unicode! On mac it is easy to use System Preferences ⟩ Keyboard ⟩ Text to set up convenient replacements so that typing is easy. I can also recommend the utility app Glyphfinder, which makes it very easy to search for and obtain unicode characters. 
 
___
*[Kernel/WolframModelEvolutionObject.m, line 495 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MFa7wL675UKeR8zDo9O:-MFa7wL675UKeR8zDo9P:b8yl9xy) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/Kernel/WolframModelEvolutionObject.m#L495)):*
> ```objc
> 
> 
> toPositiveStep[min_ : 0, total_, requested_Integer, caller_, name_] /; !(- total - 1 + min <= requested <= total) := (
> ```

Minor nitpick: the conditions are set up to be exhaust the set of possibilities here, but it would just be clearer to use a single pattern and a Which statement so that it was self-documented that the possibilities are exhausted. You would also avoid repeating the same logic in subtle ways.

taliesinb (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)* status: all files reviewed, 6 unresolved discussions (waiting on @maxitg)

___
*[README.md, line 934 at r1](https://reviewable.io/reviews/maxitg/setreplace/394#-MF_pvGe0OgpkqaN_Kag:-MFa9a6UE3J9r12HsOUQ:bogevfq) ([raw file](https://github.com/maxitg/setreplace/blob/c6b51fae21cc35e138ea1a0a6697303777fe90de/README.md#L934)):*

I would use "i.e." when the condition "they are produced as multiple outputs of a single event" is one of many possible conditions that produce spacelike separation. But isn't it the only, *defining* condition? In that case, you should avoid the word "i.e.", and probably drop the brackets since the definitions are important, not an afterthought. 

Here's what I suggest:

> 1. spacelike -- the edges were produced (directly or indirectly) by a single event 
> 2. branchlike -- the edges were produced (directly or indirectly) by multiple events that matched the same edge
> 3. timelike -- one edge produced or was produced by another edge



</blockquote></details>

By "edge" I meant to write "expression". Sorry. I was copying your usage but I strongly believe the word "edge" should not be used to refer to expressions.

taliesinb (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)* status: :shipit: complete! all files reviewed, all discussions resolved

elshatlawy (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)* status: :shipit: complete! all files reviewed, all discussions resolved

elshatlawy (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/394)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #345, opened 2020-05-14
Options for controlling the matching based on relationships between edges


maxitg:
Requires #343, ideally #344.

### The problem
We need to finally make use of the C++ code's ability to detect if a group of expressions is timelike, spacelike, or branchlike. To do that, we need to modify the matching code to filter matches based on their type.

### Possible solution
A simple approach would be to add an option to `WolframModel`, which would take a list of allowed relationships (i.e., `{"PureSpaceLike", "MixedSpaceBranchLike", "PureBranchLike"}`).

### Additional context
Requires implementing a new option on the WL side, but most of the implementation is in C++:
- [ ] Implement the option in `libSetReplace`;
- [ ] Implement the option in `WolframModel`.

PR #397, opened 2020-08-26
Spacelike multiway system

### LINKED ISSUE #342: Test if a pair of expressions is timelike, spacelike or branchlike

maxitg:
## Changes
* Closes #342.
* Partially implements #345.
* Adds `"MultiwaySpacelike"` value to `"EventSelectionFunction"`, which allows one to evaluate multiway systems which only match spacelike separated expressions.

## Examples
* Note that here a branchlike separated pair of expressions is produced, but is not matched:
```wl
In[] := WolframModel[{{{1}} -> {{1, 2}}, {{1}} -> {{1, 2, 3}}, {{1, 2}, {1, 3,
       4}} -> {{1, 2, 3, 4}}}, {{1}}, \[Infinity], 
  "EventSelectionFunction" -> "MultiwaySpacelike"]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, After]]
```
image

* Compare with the singleway system which does not produce the branchlike pair at all:
```wl
In[] := WolframModel[{{{1}} -> {{1, 2}}, {{1}} -> {{1, 2, 3}}, {{1, 2}, {1, 3,
       4}} -> {{1, 2, 3, 
      4}}}, {{1}}, \[Infinity]]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, After]]
```
image

* And a match-all system that matches the pair:
```wl
In[] := WolframModel[{{{1}} -> {{1, 2}}, {{1}} -> {{1, 2, 3}}, {{1, 2}, {1, 3,
       4}} -> {{1, 2, 3, 4}}}, {{1}}, \[Infinity], 
  "EventSelectionFunction" -> None]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, After]]
```
image

* An evolution for a random rule from the registry:
```wl
In[] := WolframModel[{{1, 1, 2}, {3, 4, 2}} -> {{4, 4, 1}, {5, 1, 4}, {5, 3, 
     2}}, {{1, 1, 1}, {2, 3, 1}}, 10, 
  "EventSelectionFunction" -> "MultiwaySpacelike"]["ExpressionsEventsGraph", 
 GraphLayout -> "SpringElectricalEmbedding"]
```
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/397)


taliesinb (changes requested):
> Reviewed 12 of 14 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/397)* status: all files reviewed, 5 unresolved discussions (waiting on @aokellermann and @maxitg)

___
*[README.md, line 1524 at r2](https://reviewable.io/reviews/maxitg/setreplace/397#-MFtgRvtA3oTkB8QzB8p:-MFtgRvtA3oTkB8QzB8q:b-xfnb2g) ([raw file](https://github.com/maxitg/setreplace/blob/812603ca6c1886d9ffcb92d93f4ff14fc1ad7138/README.md#L1524)):*
> ```Markdown
> Note that one can obtain different evolutions in this case depending on [the ordering function](#eventorderingfunction).
> 
> `"Spacelike"` evolution is the multiway version of `"GlobalSpacelike"`.
> ```

Personally, I found the whole terminology of 'local multiway systems' confusing and more complex than necessary, since it isn't a conceptual idea about how matching should work (spacelike is the only logical choice). Instead, it is an idea about how to *implement* spacelike matching on wolfram models efficiently enough that you can COMPUTE the ordinary multiway behavior, which previously wasn't feasible. 

In other, local multiway systems are just ordinary multiway systems, it is the computation strategy that is 'local', where the word 'local' here really means 'dynamic programming based on branchlike/spacelike separation'.

So maybe it would be better to use names that immediately communicate the things that are relevant for the user, rather than reflecting implementation algorithms (and the history of their evolution in this codebase).

For example, `"EventSelectionFunction" ⭢ "SinglewaySpacelike" | "MultiwaySpacelike" | "Acausal"` would be much clearer for someone who wasn't familiar with the particular implementation history. 
___
*[libSetReplace/Event.cpp, line 74 at r2](https://reviewable.io/reviews/maxitg/setreplace/397#-MFzvWb243OIt4cEqIDS:-MFzvWb42TPne0o2eij2:b-x2o7jb) ([raw file](https://github.com/maxitg/setreplace/blob/812603ca6c1886d9ffcb92d93f4ff14fc1ad7138/libSetReplace/Event.cpp#L74)):*
> ```cpp
>       const auto& expression = firstExpressionAndChosenEvent.first;
>       const auto& chosenEvent = firstExpressionAndChosenEvent.second;
>       if (secondDestroyerChoices.count(expression) && secondDestroyerChoices.at(expression) != chosenEvent) {
> ```

A comment would be nice. I think I remember the logic from our discussion (this detects whether you can overlay the destroyer choices consistently), but a one or two line description of what is going on would be nice!
___
*[libSetReplace/Event.cpp, line 109 at r2](https://reviewable.io/reviews/maxitg/setreplace/397#-MFzvniO1jyj8gjbfF2o:-MFzvniO1jyj8gjbfF2p:b-qnzm3o) ([raw file](https://github.com/maxitg/setreplace/blob/812603ca6c1886d9ffcb92d93f4ff14fc1ad7138/libSetReplace/Event.cpp#L109)):*
> ```cpp
>   void addLastEventDestroyerChoices() {
>     if (!isSpacelikeEvolution_) return;
>     const auto& lastEvent = events_.back();
> ```

Comments here would also be nice... what is this function doing, and what is the loop for?
___
*[libSetReplace/Match.cpp, line 386 at r2](https://reviewable.io/reviews/maxitg/setreplace/397#-MFzw5Cn7VELo2RWeGZo:-MFzw5Cn7VELo2RWeGZp:b-wl8wh1) ([raw file](https://github.com/maxitg/setreplace/blob/812603ca6c1886d9ffcb92d93f4ff14fc1ad7138/libSetReplace/Match.cpp#L386)):*
> ```cpp
>   }
> 
>   bool isSpacelikeSeparated(const ExpressionID newExpression, const std::vector<ExpressionID>& previousExpresisons) {
> ```

"previousExpresisons" -- typo.
___
*[libSetReplace/SetReplace.cpp, line 214 at r2](https://reviewable.io/reviews/maxitg/setreplace/397#-MFzwaw_An3C98c0Y2Un:-MFzwaw_An3C98c0Y2Uo:bn6l4su) ([raw file](https://github.com/maxitg/setreplace/blob/812603ca6c1886d9ffcb92d93f4ff14fc1ad7138/libSetReplace/SetReplace.cpp#L214)):*
> ```cpp
>   std::vector<Rule> rules;
>   std::vector<AtomsVector> initialExpressions;
>   Set::SystemType eventSelectionFunction;
> ```

Is eventSelectionFunction a good name for this variable, now?

taliesinb (approved):
> Reviewed 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/397)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aokellermann)

PR #400, opened 2020-08-30
Condition for an entire set

### LINKED ISSUE #66: Condition for an entire set
### The problem

The Wolfram Language (`Method -> "Symbolic"`) implementation of *SetReplace* already supports conditions on individual expressions. For example, the following works just fine:

```wl
In[] := WolframModel[<|
   "PatternRules" -> {x_, y_ /; y == -2} :> {x + y}|>, {1, -2, 
   3}, \[Infinity]]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, After]]
```
image

However, the conditions on the entire set are not supported. For example, the following does not get matched at all because the left-hand side of the rule is treated as a single expressions:

```wl
In[] := WolframModel[<|
   "PatternRules" -> {x_, y_} /; x + y == 0 :> {x + y}|>, {2, -2, 
   3}, \[Infinity]]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, After]]
```
image

### Possible solution

Wolfram Language implementation currently deconstructs the rules and then puts them back together. However, it assumes the left-hand side of the rule is a list, it does not understand conditions.

One will need to manually pick the condition up, keep it through rule transformations (`toNormalRules`), and use it for matching.

maxitg:
## Changes
* Closes #66.
* Allows one to use conditions in pattern rules spanning multiple expressions at once.

## Comments
* This only affects WL implementation, as C++ implementation does not support non-trivial pattern rules.
* Added lots of comments to [setSubstitutionSystem$wl.m](/Kernel/setSubstitutionSystem$wl.m), because it was quite incomprehensible.

## Examples
* Evaluate a rule that adds numbers only if their sum is even:
```wl
In[] := WolframModel[<|"PatternRules" -> {a_, b_} /; EvenQ[a + b] :> {a + b}|>,
   Range[8], \[Infinity]]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, Before]]
```
image

* Evaluate a more complex rule:
```wl
WolframModel[<|
   "PatternRules" -> {a_, b_, c_} /; a >= b + c :> {a - b - c}|>, 
  Range[11], \[Infinity]]["ExpressionsEventsGraph", 
 VertexLabels -> Placed[Automatic, Before]]
```
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/400)


taliesinb (changes requested):
> Reviewed 8 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/400)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/setSubstitutionSystem.m, line 275 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-NBGIEafaK4kMvDD_:-MG-NBGJDFIWta25QqoS:b7vudrf) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/setSubstitutionSystem.m#L275)):*
> ```objc
> 
> inertConditionSimpleRuleQ[
>     inertCondition[left : {{__ ? (AtomQ[#]
> ```

Can you maybe explain what this code does? The pattern is complex and really hard to understand. I tried to simplify it but I couldn't make any progress without understanding what it does.

One way to make it more understandable would be to make the `? (AtomQ[#] || MatchQ[#, ...]` into a named function called `xxxxQ` with a descriptive name. 
___
*[Kernel/setSubstitutionSystem$wl.m, line 8 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-Thri3v1aMRMZxBM1:-MG-Thri3v1aMRMZxBM2:b-1xdu71) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/setSubstitutionSystem$wl.m#L8)):*
> ```objc
>    but is otherwise much slower. Supports arbitrary pattern rules with conditions. Does not support multiway systems. *)
> 
> (* We are going to transform set substitution rules into a list of n! normal rules, where elements of the input subset
> ```

Can I ask if there is a better way to do this that takes advantage of the symbol attributes `Orderless` and `Flat` so that the kernel's internal pattern matcher is doing all the work for you? I don't really know enough to immediately answer this question myself, but here's a simple illustration that it is quite powerful:

```
ClearAll[OrderlessList];
SetAttributes[OrderlessList, {Orderless, Flat}];
init = OrderlessList[{a, b}, {b, c}, {c, d}];
init /. OrderlessList[x_List, y_List] /; EchoHold[DisjointQ[x, y]] :> OrderlessList[Union[x, y]]
```

Notice that the `EchoHold` shows that all ways of matching the slots `x` and `y` are tried until one of them satisfies the `DisjointQ`. If the patterns `x_List` and `y_List` were more elaborate, like `x:{_, a_, _, b_}` and `y:{a_, b_, _, _}` you could presumably enforce arbitrary shared-atom constraints that we need to express hypergraph patterns.

This might be as slow (or slower) than what your matching engine is already doing, depending on what strategy the Mathematica kernel uses, but even if it is equivalent in computational complexity it is probably faster to use it since it has been written in C and optimized over decades etc. 
___
*[Kernel/setSubstitutionSystem$wl.m, line 282 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-PeuhDauxtfLS5ChO:-MG-PeuhDauxtfLS5ChP:b-j4m2rc) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/setSubstitutionSystem$wl.m#L282)):*
> ```objc
> 
> initVertexIndex[$vertexIndex[index_], set_] := (
>   index = Merge[Association[Thread[expressionVertices[#] -> 1]] & /@ set, Total];
> ```

Isn't this clearer as `index = Counts @ Catenate @ Lookup[expressionVertices, set]`? 
___
*[Kernel/setSubstitutionSystem$wl.m, line 289 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-QuXx5C4T7g5Fygl9:-MG-QuXyBHVWOw5IUB57:b-outgxi) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/setSubstitutionSystem$wl.m#L289)):*
> ```objc
> 
> deleteFromVertexIndex[$vertexIndex[index_], expr_] := ((
>       index[#] = Lookup[index, Key[#], 0] - 1;
> ```

You can do this in one line with:

```
If[KeyAddTo[index, #, -1] === 0, index[#] =. ] 
```

This uses the GeneralUtilities function `KeyAddTo` (which has useful sister functions `KeyUnionTo`, `KeyPrependTo`, `KeyAppendTo`, `KeyJoinTo`). That function will assume a default value of 0 for non-existing keys, and will also return the new value. The syntax =. is better than KeyDropFrom since it doesn't require the additional Key[...] wrapper to handle lists.

Lastly, a minor nit, but you should use `Scan` rather than `Map`, for the reason that the resulting list of Nulls is not used, so you avoid taking O(n) memory.

A similar comment applies to `addtoVertexIndex`.
___
*[Kernel/toCanonicalRules.m, line 5 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-SJu3-zSbtM12Woad:-MG-SJu3-zSbtM12Woae:bipb1v3) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/toCanonicalRules.m#L5)):*
> ```objc
> PackageScope["toCanonicalRules"]
> 
> (* Rules can be specified in various ways, in particular, a single rule instead of a list, with or without a Module,
> ```

There is a useful but not widely known syntax for RuleDelayed + Module that looks like this:

```
lhs_ :> Module[{z}, 
  z = someFunc[lhs]; 
  z /; z =!= $Failed
]
``` 

This allows you to put a final check on the replacement value (it can be anything, not just `z =!= $Failed`). It is quite useful when the computation of the replacement value can fail (applying the condition in the pattern itself would require you to compute the replacement value twice). 

Is this syntax supported? If it's not hard to support it might be worth supporting.

taliesinb (commented):
> Reviewed 1 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/400)* status: all files reviewed, 5 unresolved discussions (waiting on @maxitg)

taliesinb (approved):
> Reviewed 3 of 4 files at r3, 1 of 2 files at r4, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/400)* status: all files reviewed, 1 unresolved discussion (waiting on @maxitg)

___
*[Kernel/setSubstitutionSystem$wl.m, line 289 at r1](https://reviewable.io/reviews/maxitg/setreplace/400#-MG-QuXx5C4T7g5Fygl9:-MG4Aui1CGn5urSQvi13:byiwh2q) ([raw file](https://github.com/maxitg/setreplace/blob/fed1a22b0878a4a2bc542947acda14be416658c7/Kernel/setSubstitutionSystem$wl.m#L289)):*

How certain are you that `GeneralUtilities` is not going to get broken in a new Wolfram Language version? If we can be reasonably certain, I agree we should use them, but surely we should introduce them in a separate PR (#402).
</blockquote></details>

Very certain. Vast amounts of System` functionality assume GU is there. But fair point about it being suited for a different PR.

Issue #402, opened 2020-08-31
GeneralUtilities


maxitg:
### The problem
The build-in `GeneralUtilities` package seems to have a lot of useful functionality, and we don't currently use it in *SetReplace*. If we can be reasonably sure that it's not going to break in the future Wolfram Language version, we should start using it.

- [ ] Add `GeneralUtilities` dependency.

### Examples
Some examples where it can be useful (feel free to add more):

#### `Scope` (and `ModuleScope`)
`Scope` allows one to have local variables (like `Block` or `Module`). However, it detects them automatically (at definition time), so that they don't have to be manually tracked. This will reduce potential mistakes due to the mismatch of lists of `Module` variables, and will make the code simpler. This will eliminate the need for a linter to do that.

- [ ] Replace all `Module`s in the WL code with `Scope` or `ModuleScope`.

#### `KeyAddTo` (and `KeyUnionTo`, `KeyPrependTo`, `KeyAppendTo`, `KeyJoinTo`)
`KeyAddTo` assumes a default value of 0 for non-existing keys, and will also return the new value.
It is useful, for example, in `deleteFromVertexIndex` in `setSubstitutionSystem$wl.m`, where the `Scan` function can simply be replaced with
```wl
If[KeyAddTo[index, #, -1] === 0, index[#] =. ] 
```

- [ ] Add `KeyAddTo` where relevant.

taliesinb:
- [ ] Update the guidelines in `.github/CONTRIBUTING.md` to mention `Scope` and `ModuleScope` instead of `With` and `Module`

PR #404, opened 2020-09-14
EventDeduplication -> SameInputSetIsomorphicOutputs

### LINKED ISSUE #341: Auto-deduplicate identical matches
### The problem
Consider a system:
```wl
WolframModel[{{1, 2}, {1, 3}, {1, 4}} -> {{1, 1, 1}}, {{1, 2}, {1, 
   3}, {1, 4}}, "FinalState"]
```
Here, only one event is possible, `{{1, 2}, {1, 3}, {1, 4}} -> {{1, 1, 1}}`, however, there are six different matches (all of which enumerated by the C++ code), because `{{1, 2}, {1, 3}, {1, 4}}` can be matched in different orders.

This yields inefficiencies and produces a larger multiway system, even though, these events can be considered identical.

### Possible solution
The C++ code should automatically detect such cases, and deduplicate them. Note, this requires right-hand sides to have the same automorphism as the left-hand sides, otherwise, the events would actually yield different results if applied. Finally, note, this is a breaking change as it changes the multiway system and the weights in `"EventOrderingFunction" -> "Random"`. So, this behavior should be controlled with an option.

### Alternative solution
A better approach might be just checking the outcome. I.e., if the outputs of two matches are the same (up to the naming of new atoms), treat the matches as identical. The input subsets will have to be the same as well, of course.

maxitg:
## Changes
* Closes #341.
* Adds `"EventDeduplication"` option, which, if set to `"SameInputSetIsomorphicOutputs"`, allows one to treat multiple matches to the same set of inputs producing isomorphic outputs as one.

## Review questions
* Can you think of a better name than `"SameInputSetIsomorphicOutputs"`?

Keep in mind that another possibility for event identification would be, for example, a rule `{{0}, {0}} -> {}` starting with an input `{{0}, {0}, {0}}`. In this case, there are three events possible, but they are still the same in the sense that the states after instantiating any of these events would be identical. So, we might want to add an option for this case as well. Maybe, as an element in a list `"EventDeduplication" -> {"SameInputSetIsomorphicOutputs", ...}`. And if all possible identification are needed, `"EventDeduplication" -> All`.

## Examples
* By default, matches to every possible order of the input subsets will be turned into separate events:
```wl
In[] := WolframModel[{{a, b}, {a, c}} -> {{b, d}, {c, d}}, {{1, 2}, {1, 3}}, 
  "EventSelectionFunction" -> 
   "MultiwaySpacelike"]["ExpressionsEventsGraph", 
 VertexLabels -> Automatic]
```
image

* However, with `"EventDeduplication" -> "SameInputSetIsomorphicOutputs"` they will be identified if the outputs are the same up to the renaming of new atoms:
```wl
In[] := WolframModel[{{a, b}, {a, c}} -> {{b, d}, {c, d}}, {{1, 2}, {1, 3}}, 
  "EventSelectionFunction" -> "MultiwaySpacelike", 
  "EventDeduplication" -> "SameInputSetIsomorphicOutputs"]["ExpressionsEventsGraph", 
 VertexLabels -> Automatic]
```
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404)


taliesinb (changes requested):
I'm not mad about the name "EventIdentification". The word "identification" (to non-mathematicians) mostly connotes "how you recognize something from its properties". It does not typically mean "deduplication". I would suggest "EventDeduplication" ⭢ True/False. Or "EventMerging" ⭢ "ByOutcome"/None. I wouldn't call it isomorphism since that has an existing meaning for (hyper-)graphs that I think is more general than what is meant here (I think, please challenge me if you disagree).

> Reviewed 16 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404#-:-MHmHQnP6aZ9TT92xh0g:b-wbeklf)* status: all files reviewed, 2 unresolved discussions (waiting on @aokellermann and @maxitg)

___
*[Kernel/setSubstitutionSystem.m, line 352 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmDwxyEiJlOGnV52FW:-MHmDwxyEiJlOGnV52FX:bxqn05h) ([raw file](https://github.com/maxitg/setreplace/blob/bbd34811ce6592600a991218ead04267a0980995/Kernel/setSubstitutionSystem.m#L352)):*
> ```objc
>     eventOrderingFunction = parseEventOrderingFunction[caller, OptionValue["EventOrderingFunction"]],
>     eventSelectionFunction = parseParameterValue[
>       caller, "EventSelectionFunction", OptionValue["EventSelectionFunction"], $eventSelectionFunctions],
> ```

FYI: This kind of could might benefit from using `Scope` and `UnpackOptions.. Doesn't have to be this PR of course, just a friendly reminder!
___
*[libSetReplace/Match.cpp, line 597 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmFhVv-dPe2uUjkrvC:-MHmFhVv-dPe2uUjkrvD:b-hhrfsz) ([raw file](https://github.com/maxitg/setreplace/blob/bbd34811ce6592600a991218ead04267a0980995/libSetReplace/Match.cpp#L597)):*
> ```cpp
> 
>     // Matcher does not support disconnected rules, so append the same atom to each expression to ensure connectivity
>     const Atom connectingAtom =
> ```

I'm sure the current code is correct, but why couldn't this just be the equivalent of numeric_limits<Atom>::max()? I don't really follow the logic of what is being done, but I assume there is only one connecting atom and its a dummy value. 
___
*[libSetReplace/Match.cpp, line 669 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmGn-Q2KznB0G9E5g_:-MHmGn-Q2KznB0G9E5ga:bk3ffvu) ([raw file](https://github.com/maxitg/setreplace/blob/bbd34811ce6592600a991218ead04267a0980995/libSetReplace/Match.cpp#L669)):*
> ```cpp
>   }
> 
>   static bool substituteMissingAtomsIfPossible(const std::vector<AtomsVector>& inputPatterns,
> ```

Just a minor complaint: Not for this PR to fix, but I'd love to have *all* of these common functions have a simple comment about what they do and why they do it. It would make it much easier to review the code for correctness, which right now I really cannot do without taking a whole day to understand exactly what is happening.

taliesinb (commented):
> Reviewed 6 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404)* status: 10 of 16 files reviewed, 2 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

taliesinb (commented):
> Reviewed 6 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404)* status: all files reviewed, 2 unresolved discussions (waiting on @aokellermann and @maxitg)

maxitg (commented):
I like `"EventDeduplication"`. But I'm not sure what's wrong with naming it "isomorphism". It's an isomorphism between the names of new vertices. I.e., it's a question of whether one can name new vertices in such a way that the two outputs are exactly identical. The only difference with conventional hypergraph isomorphism is that in this case only some but not all vertices can be renamed.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404#-:-MHmWRjl-cwJhdwhzb6V:b-yvwnvh)* status: 15 of 16 files reviewed, 2 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem.m, line 352 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmDwxyEiJlOGnV52FW:-MHmXsYh1kDeDBOXGUvr:b-1u8646) ([raw file](https://github.com/maxitg/setreplace/blob/a6057b0454d2f6d4618781b973fd97fae27b4d21/Kernel/setSubstitutionSystem.m#L352)):*

FYI: This kind of could might benefit from using `Scope` and `UnpackOptions.. Doesn't have to be this PR of course, just a friendly reminder!
</blockquote></details>

Yes, I do want to work on it at some point: #402.
___
*[libSetReplace/Match.cpp, line 597 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmFhVv-dPe2uUjkrvC:-MHmY4Z_B3Dx4pLsvzUg:b6334tk) ([raw file](https://github.com/maxitg/setreplace/blob/a6057b0454d2f6d4618781b973fd97fae27b4d21/libSetReplace/Match.cpp#L597)):*

I'm sure the current code is correct, but why couldn't this just be the equivalent of numeric_limits<Atom>::max()? I don't really follow the logic of what is being done, but I assume there is only one connecting atom and its a dummy value. 
</blockquote></details>

Good idea. Except instead of `numeric_limits::max()` we should use a special constant because `numeric_limits::max()` might cause overflows which could easily lead to weeds.
___
*[libSetReplace/Match.cpp, line 669 at r1](https://reviewable.io/reviews/maxitg/setreplace/404#-MHmGn-Q2KznB0G9E5g_:-MHsNGe09kbVZY-ThPqY:b6549n1) ([raw file](https://github.com/maxitg/setreplace/blob/a6057b0454d2f6d4618781b973fd97fae27b4d21/libSetReplace/Match.cpp#L669)):*

Just a minor complaint: Not for this PR to fix, but I'd love to have *all* of these common functions have a simple comment about what they do and why they do it. It would make it much easier to review the code for correctness, which right now I really cannot do without taking a whole day to understand exactly what is happening. 
</blockquote></details>

Done for new functions in this PR. Tracked in #421 for the old functions.

taliesinb (approved):
:lgtm:

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/404#-:-MHwc6BaE2dd3FRIXFgr:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @aokellermann and @maxitg)

PR #437, opened 2020-10-08
C++ Profiling Tests

aokellermann:
## Changes
* Adds to GTest same tests as in [performanceTest.wls](performanceTest.wls)
* Adds CMake target for profiling tests

## Comments
* Related to #428

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)


maxitg (changes requested):
> Reviewed 2 of 4 files at r1, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)* status: all files reviewed, 3 unresolved discussions (waiting on @aokellermann)

___
*[CMakeLists.txt, line 46 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA1C35E-kc0dfYcNf4:-MJA1C35E-kc0dfYcNf5:b-rp9grx) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/CMakeLists.txt#L46)):*
> ```CMake
> 
> if(SET_REPLACE_PROFILE)
>   set(SET_REPLACE_PROFILE_OPTIONS "-O0")
> ```

I'm confused. Shouldn't it be the opposite, as for profiling purposes we want the performance to be as close to the release version as possible?
___
*[libSetReplace/test/profile_tests.cpp, line 44 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA8CjS9bmnR9rVF4KU:-MJA8CjS9bmnR9rVF4KV:bm0z7qy) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/libSetReplace/test/profile_tests.cpp#L44)):*
> ```cpp
> 
> TEST(Set, profileLargeRule) {
>   Set set({{{{-1, -2}, {-2, -1}, {-1, -3}, {-2, -3}, {-3, -1}, {-3, -2}},
> ```

It seems wrong to have these rules duplicated in two different places. Maybe we can write the rules to a separate file, and have both WL and C++ code read from it?
___
*[libSetReplace/test/profile_tests.cpp, line 231 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA6Zrd6jm5ggIVnyyd:-MJA6Zrd6jm5ggIVnyye:b-ur2vad) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/libSetReplace/test/profile_tests.cpp#L231)):*
> ```cpp
>                  {-12, -8, -8},
>                  {-14, -9, -9}}};
> ```

What about the other two rules? There are 10 total, 8 for each case of the previous row of cells, and two for extending the boundary on each side.

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)* status: all files reviewed, 3 unresolved discussions (waiting on @maxitg)

___
*[CMakeLists.txt, line 46 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA1C35E-kc0dfYcNf4:-MJAGczt2VsBljPNnZ5v:b-1fsirv) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/CMakeLists.txt#L46)):*

I'm confused. Shouldn't it be the opposite, as for profiling purposes we want the performance to be as close to the release version as possible?
</blockquote></details>

[This](https://www.jetbrains.com/help/clion/cpu-profiler.html#UsingTheProfiler) says that you should reduce optimization because of inlining.
___
*[libSetReplace/test/profile_tests.cpp, line 44 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA8CjS9bmnR9rVF4KU:-MJAH1RT-BWFWAylJovn:b-bangth) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/libSetReplace/test/profile_tests.cpp#L44)):*

It seems wrong to have these rules duplicated in two different places. Maybe we can write the rules to a separate file, and have both WL and C++ code read from it?
</blockquote></details>

It seems like more trouble than it's worth honestly, but it's up to you.
___
*[libSetReplace/test/profile_tests.cpp, line 231 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA6Zrd6jm5ggIVnyyd:-MJAHL0G1VMKCLO0CWwT:bb71nyg) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/libSetReplace/test/profile_tests.cpp#L231)):*

What about the other two rules? There are 10 total, 8 for each case of the previous row of cells, and two for extending the boundary on each side.
</blockquote></details>

Oops

maxitg (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)* status: all files reviewed, 1 unresolved discussion (waiting on @aokellermann)

___
*[CMakeLists.txt, line 46 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA1C35E-kc0dfYcNf4:-MJEoL_ECkQFfBLCn_l6:b-tztaoe) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/CMakeLists.txt#L46)):*

[This](https://www.jetbrains.com/help/clion/cpu-profiler.html#UsingTheProfiler) says that you should reduce optimization because of inlining.
</blockquote></details>

Fair enough, I guess different options are useful for different purposes. Do you think though it should be instead called `SET_REPLACE_WEEDWHACK_OPTIONS` because that's what one would want to use for weed whacking as well? It would also make it more clear what this option is going to do.
___
*[libSetReplace/test/profile_tests.cpp, line 44 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA8CjS9bmnR9rVF4KU:-MJEow-07aI4Qf51WkrY:b-b2q8bs) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/libSetReplace/test/profile_tests.cpp#L44)):*

It seems like more trouble than it's worth honestly, but it's up to you.
</blockquote></details>

Ok, we can keep them here. The problem is that we will want to add more tests in the future (e.g., for testing various evolution options like event selection function). And we don't want to duplicate them in two different places every time.

However, the right solution is probably to keep the libSetReplace tests here and call them directly from `./performanceTest.wls`, but that's too big of a task to do here, so I agree the best thing we can do now is to merge it as is.

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @maxitg)

___
*[CMakeLists.txt, line 46 at r3](https://reviewable.io/reviews/maxitg/setreplace/437#-MJA1C35E-kc0dfYcNf4:-MJEugQ54K8rOH7v3S8O:bi72cl3) ([raw file](https://github.com/maxitg/setreplace/blob/6975595fc39f9f2e941ee6c8a0e9b7413d92640b/CMakeLists.txt#L46)):*

Fair enough, I guess different options are useful for different purposes. Do you think though it should be instead called `SET_REPLACE_WEEDWHACK_OPTIONS` because that's what one would want to use for weed whacking as well? It would also make it more clear what this option is going to do.
</blockquote></details>

I decided to just remove it. No inlining is happening anyway, so it doesn't really help to add the option.

maxitg (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/437)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #449, opened 2020-10-11
Automated markdown image embedding 


taliesinb:
### The problem

Creating and maintaining markdown files that incorporate images from Mathematica is needlessly tedious, involving manual copy-pasting from a live notebook, exporting to an image file on disk, and then hyperlinking the file in to the markdown document. I think we can improve that with automated tooling. 

### Possible solution

The most basic technology would be:

1. A *render script* wolframscript that processes a markdown file, finds WL code cells within it that indicate they wish to 'embed their output', and then runs these code cells to produce the desired images. These images are processed, compressed (e.g. using PNG limited-palette compression when appropriate), and automatically placed into the correct folder and the corresponding hyperlink in the .md file updated to point to the new file. If the images already exist within the .md file, they are merely replaced or updated. 

2. A more sophisticated technology would involve a bidirectional, lossless conversion between (a subset of) Mathematica notebooks and markdown files. This is not crucial to solve this issue, but is a missing piece of technology that would it a practical reality to work with markdown-based research in a WYSIWYG fashion for those who wish to do so. This would make cross-publishing to Wolfram Physics bulletins a more friendly experience. 

I discuss some small elaborations in the point below.

### Alternative solutions

Continue doing what we currently do. Personally I refer working in semi-WYSIWYG editors since they allow rapid prototyping of the 'flow' of a post, and continuous minor tweaking of visual details and style. The current manual practice does not easily support this.

### Additional context

Some small elaborations of point 1 are described here that are not necessary for the minimal viable product:

1. An auxiliary script can perform garbage collection when outdated image files are no longer references by markdown files in the image folder.

2. If the image after compression is small enough that it can fit into some small number of base64 characters (e.g. 400), it can be directly embedded in the markdown notebook via the "data" URL scheme (on a single line), avoiding the need for a separate file. This has several advantages, such as allowing the default rendering behavior of .md files within Github's diffing tools to include such images. 

3. Pursuant to the earlier point, a heavily compressed, low-resolution version of the image could *always* be embedded via a "data" URL, to make diffs easier to evaluate and to allow a .md file to be self-contained in a way that makes it easier to share (albiet with lower visual quality for the associated images). Most editors are good at ignoring very long lines that contain embedded data in this way, but a definitely limit should be placed to ensure that if line-wrapping is enabled, the image does not contain more than e.g. a half-page of base64.

[...]

PR #450, opened 2020-10-11
EventOrderingFunction "Any"

### LINKED ISSUE #351: Nondeterministic evolution order
### The problem
A specific evolution order is a big problem for parallelization efforts. However, sometimes it is useful to generate an evolution with *any* event order even if it is not deterministic or reproducible. Allowing for such arbitrary evolution order makes parallelization, especially for large-scale distributed systems, much easier to implement.

Also note, that in the case of causal invariant or local multiway systems, the evolution order does not matter (except for event numbering), and these kinds are the most interesting systems, so it makes perfect sense to focus parallelization efforts on the arbitrary event ordering case.

### Possible solution
`"EventOrderingFunction" -> "Any"` would allow for any evolution order, which does not have to be seed-dependent, deterministic, or reproducible. This should be passed from `WolframModel` to the C++ code.

### Additional context
This does not yet require any changes on the C++ side, for a single-thread evolution,`"Any"` could behave the same way as random. To retain some level of determinism, it will be useful to have #146, although it is not strictly necessary.

### Subtasks
- [ ] Implement the `WolframModel` side.
- [ ] Implement the C++ side;

aokellermann:
## Changes
* Adds `"EventOrderingFunction"` option `"Any"` on WL side, and `Matcher::OrderingFunction::Any` on C++ side.
* Closes #351 

## Examples
Updated example added to the README is [here](https://github.com/maxitg/SetReplace/blob/b2473dc36be8c9f79f5b2d1fe57b14e98ead8ec1/READMEImages/AllEventOrderingFunctionPlots.png).

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/450)


maxitg (changes requested):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/450)* status: all files reviewed, 5 unresolved discussions (waiting on @aokellermann)

___
*[README.md, line 1503 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJNx6xi8Ww_KSS34GHH:-MJNx6xi8Ww_KSS34GHI:bia9yzp) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1503)):*
> ```Markdown
> 
> * `"Random"`: selects a single match uniformly at random. It is possible to do that efficiently because the C++ implementation of `WolframModel` (the only one that supports `"EventOrderingFunction"`) keeps track of all possible matches at any point during the evolution. `"Random"` is guaranteed to select a single match, so the remaining sorting criteria are ignored. It can also be omitted because the random event is always chosen if provided sorting criteria are insufficient. The seeding can be controlled with [`SeedRandom`](https://reference.wolfram.com/language/ref/SeedRandom.html). However, the result does depend on your platform (Mac/Linux/Windows) and the specific build (version) of *SetReplace*.
> * `"Any"`: the chosen match is undefined.
> ```

Add an empty line before.
___
*[README.md, line 1503 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJNx89q-Pm5_PhnxQjg:-MJNx89q-Pm5_PhnxQjh:b-eyegj1) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1503)):*
I think this is a bit obscure. I think it should say something like:

> could select any match leading to nondeterministic and undefined evolution order. In some cases, it has better performance than `"Random"`.

Otherwise, people would be confused about why they would ever want to use it over `"Random"`.
___
*[README.md, line 1521 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO--jG1yfb-nYN33iW:-MJO--jG1yfb-nYN33iX:b-i8vds0) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1521)):*
> ```Markdown
> ```
> 
> 
> ```

It looks like the image has a different format than the other ones (e.g., `Out[2]=` instead of `Out[.]=`). It might be that the Linux version of WL produces images in a different way. Did you copy it as Edit -> Copy As -> Bitmap? If not, could you try doing that instead? Also, if your display is not HDPI, it might produce an image with too low resolution. If that happens, use this image (and we really need to do #449):

___
*[libSetReplace/Match.cpp, line 149 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO2VxKAMMUU0XNNz29:-MJO2VxKAMMUU0XNNz2A:blg3a4v) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/libSetReplace/Match.cpp#L149)):*
> ```cpp
>   // dereferenced values in the corresponding classes above.
> 
>   OrderingSpec orderingSpec_;
> ```

This should go above the two comments above.
___
*[Tests/eventOrderingFunction.wlt, line 219 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO47J02c1gFITGuz_P:-MJO47J1EQ_gtwF2WPCr:b-6fqn2) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/Tests/eventOrderingFunction.wlt#L219)):*
> ```Mathematica
>       (* When "Any" is used the match is indeterminate, so just make sure it works. *)
>       VerificationTest[
>         WolframModel[{{1}} -> {{}}, Automatic, "EventOrderingFunction" -> "Any"]["FinalState"] == {{}}
> ```

The error message would be better if it's split in two lines (that checks for equality, or, more precisely, `SameQ`), i.e.,
```wl
VerificationTest[
  WolframModel[{{1}} -> {{}}, Automatic, "EventOrderingFunction" -> "Any"]["FinalState"],
  {{}}
]
```

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/450)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @maxitg)

___
*[README.md, line 1503 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJNx6xi8Ww_KSS34GHH:-MJO9y2W5MwGv-yUKFMf:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1503)):*

Add an empty line before.
</blockquote></details>

Done.
___
*[README.md, line 1503 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJNx89q-Pm5_PhnxQjg:-MJO9zGX820xeW8ew2KQ:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1503)):*

I think this is a bit obscure. I think it should say something like:

> could select any match leading to nondeterministic and undefined evolution order. In some cases, it has better performance than `"Random"`.

Otherwise, people would be confused about why they would ever want to use it over `"Random"`.
</blockquote></details>

Done.
___
*[README.md, line 1521 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO--jG1yfb-nYN33iW:-MJOA9Nw8NKdHSHiEMy2:bmxovvq) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/README.md#L1521)):*

It looks like the image has a different format than the other ones (e.g., `Out[2]=` instead of `Out[.]=`). It might be that the Linux version of WL produces images in a different way. Did you copy it as Edit -> Copy As -> Bitmap? If not, could you try doing that instead? Also, if your display is not HDPI, it might produce an image with too low resolution. If that happens, use this image (and we really need to do #449):

</blockquote></details>

Done. Mathematica on Linux doesn't seem to have that functionality, so I just copied your image.
___
*[libSetReplace/Match.cpp, line 149 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO2VxKAMMUU0XNNz29:-MJOAJP48zVjn1N7FUJ-:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/libSetReplace/Match.cpp#L149)):*

This should go above the two comments above.
</blockquote></details>

Done.
___
*[Tests/eventOrderingFunction.wlt, line 219 at r1](https://reviewable.io/reviews/maxitg/setreplace/450#-MJO47J02c1gFITGuz_P:-MJOAL9m66PWJK3V58qW:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/241b6b4ddc9665bf6d50df83dbeddd9deb2b53e6/Tests/eventOrderingFunction.wlt#L219)):*

The error message would be better if it's split in two lines (that checks for equality, or, more precisely, `SameQ`), i.e.,
```wl
VerificationTest[
  WolframModel[{{1}} -> {{}}, Automatic, "EventOrderingFunction" -> "Any"]["FinalState"],
  {{}}
]
```
</blockquote></details>

Done.

maxitg (approved):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/450)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #31, opened 2019-06-14
Heuristic to detect if C++ implementation is exponentially slow


maxitg:
C++ code is only fast with certain assumptions, in particular, it assumes that the vertex degrees are small. There is a simple counterexample in which it is not the case:
```
SetReplace[#[[1]], FromAnonymousRules[#[[2]]], 30, 
   Method -> 
    "C++"] &@{{{1}}, {{{1}} -> {{1}, {1}, {1}}, {{1}, {1}, {1}} -> \
{{1}}}}
```
This runs 3 seconds, and that increases exponentially with the number of steps. It also consumes exponential memory.

The current algorithm in C++ implementation fails completely in this case. Thus, there should be a heuristic that would detect it and use a different method (WolframLanguage implementation, for example, works in this case).

This is important because examples like these come up during enumeration, and they need to be handled automatically.

PR #453, opened 2020-10-12
Improve performance of encodeNestedLists

### LINKED ISSUE #46: Slow passing inputs to C++ code
Passing large sets to C++ code is very slow. I.e., evaluate:
```
rules = {{{a_, b_}, {a_, c_}, {a_, d_}} :> 
    Module[{$0, $1, $2}, {{$0, $1}, {$1, $2}, {$2, $0}, {$0, $2}, \
{$2, $1}, {$1, $0}, {$0, b}, {$1, c}, {$2, d}, {b, $2}, {d, $0}}]};
AbsoluteTiming[
  set = SetReplace[{{0, 0}, {0, 0}, {0, 0}}, rules, 100]] // First
AbsoluteTiming[SetReplace[set, rules]][[1]]
```
The two timings produced should be similar, however, the second one is much slower than the other, which means the slow down occurs when the large graph is passed to C++ code.

**Version:**
 - OS: macOS 10.14.5 (18F203)
 - Wolfram Language Version: 12.0.0 for Mac OS X x86 (64-bit) (April 7, 2019)
 - SetReplace git sha: https://github.com/maxitg/SetReplace/commit/7feedcadedb382eeb028be8a3b8d9cbc6fee28b1

**Additional:**
Unit test for `SetCases` will need to be updated with a smaller time after that's fixed.

daneelsan:
## Changes
* Closes #46.
* Improve performance of `encodeNestedLists` inside `setSubstitutionSystem$cpp.m` by not using `ReplaceRepeated`.

## Examples
```wl
In[]:= rules = {{{a_, b_}, {a_, c_}, {a_, d_}} :> 
    Module[{$0, $1, $2}, {{$0, $1}, {$1, $2}, {$2, $0}, {$0, $2}, \
{$2, $1}, {$1, $0}, {$0, b}, {$1, c}, {$2, d}, {b, $2}, {d, $0}}]};
AbsoluteTiming[set = SetReplace[{{0, 0}, {0, 0}, {0, 0}}, rules, 100]] // First
AbsoluteTiming[SetReplace[set, rules]][[1]]

Out[]= 0.052426
Out[]= 0.026598
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/453)


daneelsan:
This improves this example in the documentation, but I'm not sure how to rework that part:
```wl
In[]:= AbsoluteTiming[WolframModel[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> {{0}}}, {{0}}, <|"MaxEvents" -> 30|>, Method -> "LowLevel"]] // First
Out[]= 1.88309

In[]:= AbsoluteTiming[WolframModel[{{{0}} -> {{0}, {0}, {0}}, {{0}, {0}, {0}} -> {{0}}}, {{0}}, <|"MaxEvents" -> 30|>, Method -> "Symbolic"]] // First
Out[]= 0.019555
```

maxitg (changes requested):
This example from the documentation is about a completely separate issue (#31). I don't think it really needs changing at all, but we can update the timing since it has probably improved due to this and @aokellermann's optimizations.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/453#-:-MJTINxf7ofycIDskAoz:b-svg807)* status: all files reviewed, 2 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 80 at r1](https://reviewable.io/reviews/maxitg/setreplace/453#-MJTI6WN3Dpbql4hXkZt:-MJTI6WN3Dpbql4hXkZu:bjq4081) ([raw file](https://github.com/maxitg/setreplace/blob/7c96d3c7be94a80a9a917f34529e1451c1d162a2/Kernel/setSubstitutionSystem$cpp.m#L80)):*
> ```objc
> (* This format is used to pass both rules and set data into libSetReplace over LibraryLink *)
> 
> encodeNestedLists[l_List] := Flatten @ {Length @ l, encodeNestedLists /@ l}
> ```

Please use more descriptive variable names, e.g., `l -> list` and `a -> arg`.
___
*[Tests/performance.wlt, line 33 at r1](https://reviewable.io/reviews/maxitg/setreplace/453#-MJTHLMw-W3I8abq9utX:-MJTHLMw-W3I8abq9utY:bwxzotj) ([raw file](https://github.com/maxitg/setreplace/blob/317a96150672ea97f8e532815eaba2074ade4654/Tests/performance.wlt#L33)):*
> ```Mathematica
>               1000]],
>             List,
>             TimeConstraint -> 10,
> ```

This test actually has nothing to do with this (the initial state is small here). The reason the time constraint is so large here is that the test does not fail randomly due to fluctuations in performance and would only fail if it was catastrophically slower than before.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/453)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @maxitg)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 80 at r1](https://reviewable.io/reviews/maxitg/setreplace/453#-MJTI6WN3Dpbql4hXkZt:-MJTO7vf1jvDBoqEahFo:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/d51a051c317c47c3c1ff046f0233558d46b6924d/Kernel/setSubstitutionSystem$cpp.m#L80)):*

Please use more descriptive variable names, e.g., `l -> list` and `a -> arg`.
</blockquote></details>

Done.
___
*[Tests/performance.wlt, line 33 at r1](https://reviewable.io/reviews/maxitg/setreplace/453#-MJTHLMw-W3I8abq9utX:-MJTOBIP2pdcm6T-Av_v:b-kuu49y) ([raw file](https://github.com/maxitg/setreplace/blob/2278ed1d916d6e07be5365a8191538878166169c/Tests/performance.wlt#L33)):*

This test actually has nothing to do with this (the initial state is small here). The reason the time constraint is so large here is that the test does not fail randomly due to fluctuations in performance and would only fail if it was catastrophically slower than before.
</blockquote></details>

Reverted the changes I made there.

maxitg (approved):
Awesome! :rocket:

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/453#-:-MJTSgKp4z2V376COfxs:bqb8wix)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #352, opened 2020-05-14
Parallelize nondeterministic WolframModel on CPU


maxitg:
Requires #351.

### The problem
The other CPU issue #155 attempts to parallelize `WolframModel` without changing functionality. However, this limits our ability to parallelize because one has to make sure to evaluate the events in a certain order. In this issue, we drop that requirement and parallelize it with the goal of obtaining the result for any ordering. The output can even change each time we run `WolframModel`.

Note, this is not as bad as it sounds because once we have #146 and if the system we are running is a local multiway or a causal invariant system, the evaluation order does not matter.

### Possible solution
Each thread could simply:
1. Pick a match from [the queue](https://github.com/maxitg/SetReplace/blob/aa6885272de1cec6fdf4b56677d0fda0de6e25d4/libSetReplace/Match.cpp#L141).
2. Compute the outputs and index them for new matches.

One needs to be careful here to ensure matches are not lost/computed incorrectly because some edges got deleted/added simultaneously in a different thread.

### Additional context
This issue is for CPU parallelization specifically. There is a similar one for GPUs (TODO: add issue) and distributed systems (TODO: add issue).

PR #468, opened 2020-10-14
Parallelism manager

aokellermann:
## Changes
* Implements a thread-safe RAII system for managing reservation of threads for a specific type of hardware (currently only CPU is implemented). This is done so that we don't waste resources creating more software threads than there are hardware threads, when different classes are all trying to use threading. This PR makes that easier by having a centralized "pool".
* Includes tests to ensure thread-safety and correct implementation on various CPU thread counts.
* Includes documentation.

## Comments
* Prerequisite for #352.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468)


maxitg (changes requested):
Can you comment on the end goal here? What is the strategy for parallelizing the evolution algorithm? Do we expect the GPU implementation to use the same class for managing threads?

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468#-:-MJcOAAY9bYR7odsAU0E:bgaz398)* status: all files reviewed, 16 unresolved discussions (waiting on @aokellermann)

___
*[libSetReplace/Match.cpp, line 218 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSQbO5EZ78aydfxy2:-MJcSQbO5EZ78aydfxy3:b-kjrmw6) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Match.cpp#L218)):*
> ```cpp
>     };
> 
>     {
> ```

Why an extra set of braces?
___
*[libSetReplace/Match.cpp, line 221 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSAu-0Yf1yWZCSWWA:-MJcSAu-0Yf1yWZCSWWB:bmviy9e) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Match.cpp#L221)):*
> ```cpp
>       // Only create threads if there is more than one rule
>       const auto token = Parallelism::acquire(Parallelism::HardwareType::StdCpu, rules_.size());
>       const uint64_t numThreadsToUse = rules_.size() > 1 ? token->numThreads() : 0;
> ```

Do we need to attempt to acquire anything at all if `rules_.size() == 1`?
___
*[libSetReplace/Parallelism.hpp, line 22 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcYJw-BEcLQ1z46XSQ:-MJcYJw-BEcLQ1z46XSR:bqpyspi) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.hpp#L22)):*
> ```cpp
> class ThreadAcquisitionToken {
>  public:
>   ThreadAcquisitionToken(const HardwareType& type, const int64_t& requestedNumThreads);
> ```

Will it be possible to keep the threads themselves here. So that we only create them once, run the evolution, and then join them. This way, we will avoid the thread-creating overhead, which might give us a 5-10% performance boost right away looking at your profiling results.
___
*[libSetReplace/Parallelism.hpp, line 26 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcX9gW6MUh9FBS9Kpa:-MJcX9gW6MUh9FBS9Kpb:b77a9qo) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.hpp#L26)):*
> ```cpp
>   /** @brief Returns the number of threads successfully reserved.
>    */
>   [[nodiscard]] int64_t numThreads() const noexcept;
> ```

Interesting, I didn't know about `[[nodiscard]]`. Is there any reason all `const` functions would not have it if they don't take any pointers as arguments?
___
*[libSetReplace/Parallelism.hpp, line 38 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcZpLc130q5bdVZVOu:-MJcZpLc130q5bdVZVOv:brhkaye) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.hpp#L38)):*
> ```cpp
>  */
> inline ThreadAcquisitionTokenPtr acquire(const HardwareType& type, const int64_t& requestedNumThreads) {
>   return std::make_shared<ThreadAcquisitionTokenPtr::element_type>(type, requestedNumThreads);
> ```

Can we move the definition to the `.cpp` file?
___
*[libSetReplace/Parallelism.hpp, line 45 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJchwoK6Mc5W6tNUe01:-MJchwoK6Mc5W6tNUe02:b2uk60k) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.hpp#L45)):*
> ```cpp
> bool isAvailable(const HardwareType& type);
> 
> namespace Testing {
> ```

I don't think this should be public, but I don't really have a much better solution to this. We could make it private and then use `FRIEND_TEST`. Although in that case, we will need to make sure it can still compile without gtest. But I think we should at least use a macro that would disable this if not building with tests.
___
*[libSetReplace/Parallelism.cpp, line 55 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJciSS12628dCa37LSy:-MJciSS12628dCa37LSz:bg9lb8g) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L55)):*
> ```cpp
>   if (requestedNumThreads <= 1) return 0;
>   if (type == HardwareType::StdCpu) return cpuParallelism.acquireThreads(requestedNumThreads);
>   throw std::runtime_error("Invalid Parallelism::Type");
> ```

Should it be "Invalid Parallelism::HardwareType"?
___
*[libSetReplace/Parallelism.cpp, line 61 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcimid3BbzkUgeHKXu:-MJcimid3BbzkUgeHKXv:b-6se5bg) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L61)):*
> ```cpp
>  */
> void releaseThreads(const HardwareType& type, const int64_t& numThreadsToReturn) {
>   if (numThreadsToReturn <= 1) return;
> ```

I think we should be allowed to release one thread. Currently, it fails silently in this case, which will cause a "thread leak". I realize it cannot happen in the current implementation, but it might cause problems in the future.
___
*[libSetReplace/Parallelism.cpp, line 63 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJciyd71kPlQtCrJdhc:-MJciyd71kPlQtCrJdhd:boforeu) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L63)):*
> ```cpp
>   if (numThreadsToReturn <= 1) return;
>   if (type == HardwareType::StdCpu) return cpuParallelism.releaseThreads(numThreadsToReturn);
>   throw std::runtime_error("Invalid Parallelism::Type");
> ```

Same, should it be "Invalid Parallelism::HardwareType"?
___
*[libSetReplace/Parallelism.cpp, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJckuw93n5tJZ6zUJc_:-MJckuw93n5tJZ6zUJca:b3u9cu1) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L88)):*
> ```cpp
> bool isAvailable(const HardwareType& type) {
>   if (type == HardwareType::StdCpu) return cpuParallelism.isAvailable();
>   throw std::runtime_error("Invalid Parallelism::Type");
> ```

Parallelism::HardwareType
___
*[libSetReplace/Parallelism.cpp, line 94 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcjsmCDCcjrjx_Fcm0:-MJcjsmCDCcjrjx_Fcm1:bl1dum) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L94)):*
> ```cpp
> void overrideNumHardwareThreads(const HardwareType& type, const int64_t& numThreads) {
>   if (type == HardwareType::StdCpu) cpuParallelism.overrideNumHardwareThreads(numThreads);
>   else
> ```

Add `{` and `}` after `else`.
___
*[libSetReplace/Parallelism.cpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJckveZAASrMouc2C2l:-MJckveZAASrMouc2C2m:b3u9cu1) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/Parallelism.cpp#L95)):*
> ```cpp
>   if (type == HardwareType::StdCpu) cpuParallelism.overrideNumHardwareThreads(numThreads);
>   else
>     throw std::runtime_error("Invalid Parallelism::Type");
> ```

Parallelism::HardwareType
___
*[libSetReplace/test/Parallelism_tests.cpp, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcmo7EDa3BYgQ5yv0X:-MJcmo7F3YrH0XIYKjD_:b-jz8d4y) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/test/Parallelism_tests.cpp#L47)):*
> ```cpp
>         int reserved = threadsRemaining - newThreadsRemaining;
> 
>         if (reserved <= 1) {
> ```

I'm completely confused by this. Should it be `reserved >= 1`?
___
*[libSetReplace/test/Parallelism_tests.cpp, line 48 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcnqNaC3KT35ioBilB:-MJcnqNaC3KT35ioBilC:b1kkwew) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/test/Parallelism_tests.cpp#L48)):*
> ```cpp
> 
>         if (reserved <= 1) {
>           auto token = acquire(cpu, reserved);
> ```

`const`
___
*[libSetReplace/test/Parallelism_tests.cpp, line 65 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcmy5o1mxaAMX57yZS:-MJcmy5o1mxaAMX57yZT:b-yqu059) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/test/Parallelism_tests.cpp#L65)):*
> ```cpp
> // Ensures there are no race conditions in thread acquisition.
> TEST(Parallelism, CpuThreadSafety) {
>   int n = 10000;
> ```

`constexpr`
___
*[libSetReplace/test/Parallelism_tests.cpp, line 70 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcoJQv5fa4NG2SRu5x:-MJcoJQv5fa4NG2SRu5y:basz9w7) ([raw file](https://github.com/maxitg/setreplace/blob/8fa4ffca8b7854c123e1c24170166485d670c71a/libSetReplace/test/Parallelism_tests.cpp#L70)):*
> ```cpp
>   std::vector<std::thread> threads(n / 2);
> 
>   const auto lambda = [&tokens](int i) {
> ```

Rename `lambda` -> `acquireTwoThreads`?

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468)* status: 2 of 6 files reviewed, 16 unresolved discussions (waiting on @maxitg)

___
*[libSetReplace/Match.cpp, line 218 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSQbO5EZ78aydfxy2:-MJcvO2a8p8Oz8-9Pit7:b-z2yv89) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Match.cpp#L218)):*

Why an extra set of braces?
</blockquote></details>

It's a critical section.
___
*[libSetReplace/Match.cpp, line 221 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSAu-0Yf1yWZCSWWA:-MJcvTHA4OjzPjkbwhLJ:bvjwlo6) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Match.cpp#L221)):*

Do we need to attempt to acquire anything at all if `rules_.size() == 1`?
</blockquote></details>

Acquire returns 0 if 1 thread is requested.
___
*[libSetReplace/Parallelism.hpp, line 22 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcYJw-BEcLQ1z46XSQ:-MJcvjRaDykPGclUBeU8:bamhths) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L22)):*

Will it be possible to keep the threads themselves here. So that we only create them once, run the evolution, and then join them. This way, we will avoid the thread-creating overhead, which might give us a 5-10% performance boost right away looking at your profiling results.
</blockquote></details>

Threads aren't designed to be reused. They execute the entry function upon construction. You could do a more complex task in the spirit of what you're talking about, but that code wouldn't live in this file and is far beyond the scope of this PR.
___
*[libSetReplace/Parallelism.hpp, line 26 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcX9gW6MUh9FBS9Kpa:-MJcwmgs-h3tjXs1-zMT:bwyecjx) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L26)):*

Interesting, I didn't know about `[[nodiscard]]`. Is there any reason all `const` functions would not have it if they don't take any pointers as arguments?
</blockquote></details>

Nope. 
___
*[libSetReplace/Parallelism.hpp, line 38 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcZpLc130q5bdVZVOu:-MJcxFM975kCAojtrapF:b4p59z7) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L38)):*

Can we move the definition to the `.cpp` file?
</blockquote></details>

We could, but inlining is much faster for a tiny function like this. Let me know what you think.
___
*[libSetReplace/Parallelism.hpp, line 45 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJchwoK6Mc5W6tNUe01:-MJcz7bX-ao1e5T1hAOV:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L45)):*

I don't think this should be public, but I don't really have a much better solution to this. We could make it private and then use `FRIEND_TEST`. Although in that case, we will need to make sure it can still compile without gtest. But I think we should at least use a macro that would disable this if not building with tests.
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 55 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJciSS12628dCa37LSy:-MJczRkq8_6XNg2G_Uqw:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L55)):*

Should it be "Invalid Parallelism::HardwareType"?
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 61 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcimid3BbzkUgeHKXu:-MJczTCnC2qxuzbbFJVK:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L61)):*

I think we should be allowed to release one thread. Currently, it fails silently in this case, which will cause a "thread leak". I realize it cannot happen in the current implementation, but it might cause problems in the future.
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 63 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJciyd71kPlQtCrJdhc:-MJczTlD7Ot_z0lCxjE0:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L63)):*

Same, should it be "Invalid Parallelism::HardwareType"?
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJckuw93n5tJZ6zUJc_:-MJczUA062PxQKPhQ2AC:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L88)):*

Parallelism::HardwareType
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 94 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcjsmCDCcjrjx_Fcm0:-MJczYTh7D5oxKQM-v8h:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L94)):*

Add `{` and `}` after `else`.
</blockquote></details>

Done.
___
*[libSetReplace/Parallelism.cpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJckveZAASrMouc2C2l:-MJczYl14Nvk0k4tjOu8:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.cpp#L95)):*

Parallelism::HardwareType
</blockquote></details>

Done.
___
*[libSetReplace/test/Parallelism_tests.cpp, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcmo7EDa3BYgQ5yv0X:-MJd-lu26yx2jW-xbczB:byfiiq0) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/test/Parallelism_tests.cpp#L47)):*

I'm completely confused by this. Should it be `reserved >= 1`?
</blockquote></details>

Wow, that's a bad weed.
___
*[libSetReplace/test/Parallelism_tests.cpp, line 48 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcnqNaC3KT35ioBilB:-MJd-waXEVhW1uYy4LnW:b-opdtso) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/test/Parallelism_tests.cpp#L48)):*

`const`
</blockquote></details>

It can't be moved if it's const.
___
*[libSetReplace/test/Parallelism_tests.cpp, line 65 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcmy5o1mxaAMX57yZS:-MJd05p2-skdkpcMSdFR:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/test/Parallelism_tests.cpp#L65)):*

`constexpr`
</blockquote></details>

Done.
___
*[libSetReplace/test/Parallelism_tests.cpp, line 70 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcoJQv5fa4NG2SRu5x:-MJd09Q-AwJdLpvlKJum:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/test/Parallelism_tests.cpp#L70)):*

Rename `lambda` -> `acquireTwoThreads`?
</blockquote></details>

Done.

aokellermann:
> Can you comment on the end goal here? What is the strategy for parallelizing the evolution algorithm? Do we expect the GPU implementation to use the same class for managing threads?

I was thinking that any utilities for parallelism management could go here. Why do you want to know about the strategy here? It's not related to this issue specifically. I designed this code so that it is easy to add different hardware implementations by using template specialization. I think it could be useful for GPUs as well, but maybe someone who has more experience with GPU acceleration will contradict me on that.

maxitg (changes requested):
Ok, fair enough. I was just wondering about how and where specifically it will be used.

> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468#-:-MJdc7a2CnIJkn_iol09:bub8hgo)* status: all files reviewed, 1 unresolved discussion (waiting on @aokellermann)

___
*[libSetReplace/Match.cpp, line 221 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSAu-0Yf1yWZCSWWA:-MJddffB2_v6DOwFRsL3:bl7iy9s) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Match.cpp#L221)):*

Acquire returns 0 if 1 thread is requested.
</blockquote></details>

Well, that just means we don't need `rules_.size() > 1` condition at all. I was wondering why check for the rule count after trying to acquire threads instead of before?
___
*[libSetReplace/Parallelism.hpp, line 22 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcYJw-BEcLQ1z46XSQ:-MJdcj2s3HDx8s2mI3Wm:b-uf22sn) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L22)):*

Threads aren't designed to be reused. They execute the entry function upon construction. You could do a more complex task in the spirit of what you're talking about, but that code wouldn't live in this file and is far beyond the scope of this PR.
</blockquote></details>

Sure, certainly not in this PR.
___
*[libSetReplace/Parallelism.hpp, line 38 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcZpLc130q5bdVZVOu:-MJdeJzZ6_THaku3JBkX:bi61x4h) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Parallelism.hpp#L38)):*

We could, but inlining is much faster for a tiny function like this. Let me know what you think.
</blockquote></details>

Ok, fair enough, I did not realize `inline` requires it to be in the header.

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @maxitg)

___
*[libSetReplace/Match.cpp, line 221 at r1](https://reviewable.io/reviews/maxitg/setreplace/468#-MJcSAu-0Yf1yWZCSWWA:-MJdm6J112WEeYKk2fEU:br0df3s) ([raw file](https://github.com/maxitg/setreplace/blob/a2bf340a0288dd7b15c21980158bf564dedfc039/libSetReplace/Match.cpp#L221)):*

Well, that just means we don't need `rules_.size() > 1` condition at all. I was wondering why check for the rule count after trying to acquire threads instead of before?
</blockquote></details>

Ah, i misread.

maxitg (approved):
:lgtm:

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/468#-:-MJdo_zW1zTobDD1Jb-7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #490, opened 2020-10-20
Add HypergraphToGraph function

### LINKED ISSUE #488: HypergraphToGraph
### The problem
Implement the `HypergraphToGraph` function. 

### Possible solution
One possible approach is the one used in the [WFR](https://resources.wolframcloud.com/FunctionRepository/resources/HypergraphToGraph):
```wl
HypergraphToGraph[list : {___List}, opts : OptionsPattern[Graph]] :=
  Block[{i, j, l},
    Graph[
      Flatten @ Table[
        DirectedEdge[
          l[[j]], l[[i]]],
          {l, list},
          {i, Length[l]},
          {j, i - 1}],
      opts]
  ]
```

### Comment
@maxitg said that there are multiple ways to do it, and that a second argument would specify said method.

For the implementation above, this could be: `HypergraphToGraph[hg_, Method -> "SameDistanceMatrix"]`

See "convert the hypergraph into a normal Graph preserving structure" inside [hypergraphAutomorphismGroup](https://github.com/maxitg/Kernel/HypergraphAutomorphismGroup.m):
```wl
Graph[Catenate[toStructurePreservingBinaryEdges /@ e]]
```
Perhaps: `HypergraphToGraph[hg_, Method -> "SameStructure"]`?

daneelsan:
## Changes
* Closes #488.
* `HypergraphToGraph[hgraph_, method_]` uses `method` to convert the hypergraph `hg` to a `Graph` object. Current "methods"  include: `"DirectedDistancePreserving"`, `"UndirectedDistancePreserving"` and `"StructurePreserving"`.

## Comments
* Don't know if  "method" the appropriate name for the second argument.

## Examples

```wl
In[]:= Map[
  <|
    "Hypergraph" -> "HG"[#],
    "HypergraphPlot" -> WolframModelPlot[#, VertexLabels -> Automatic],
    "HypergraphToGraph" -> Association@Table[
      m -> HypergraphToGraph[#, m, VertexLabels -> Automatic, ImageSize -> Medium],
     {m, {"DirectedDistancePreserving", "StructurePreserving", "UndirectedDistancePreserving"}}]|> &,
  {(*{},{{}},*){{1}}, {{1, 1}}, {{1, 1, 1}}, {{1, 2, 1}}, {{1, 2, 3}}, {{1, 1, 2, 3}}, {{1, 3, 2}, {2, 1, 3}, {1, 1}}}] // Dataset
```
image

#### `IsomorphicHypergraphQ`
```wl
isomorphicHypergraphQ[hgraph1_, hgraph2_] :=
 IsomorphicGraphQ[
  HypergraphToGraph[hgraph1, "StructurePreserving"], 
  HypergraphToGraph[hgraph2, "StructurePreserving"]]

In[]:= isomorphicHypergraphQ[{}, {{}}]
Out[]= False

In[]:= isomorphicHypergraphQ[{{}}, {{}}]
Out[]= True

In[]:= isomorphicHypergraphQ[{{}, {x}}, {{}, {1}}]
Out[]= True

In[]:= isomorphicHypergraphQ[{{}, {x, x, y, z}, {z, w}}, {{}, {{x}, {x}, 2, 3}, {3, 4}}]
Out[]= True
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)


daneelsan:
This is a bug, right? (should be `Graph[{1}, {}]`)
```wl
In[]:= ResourceFunction["HypergraphToGraph"][{{1}}] // InputForm
Out[]//InputForm= Graph[{}, {}]
```

maxitg (commented):
Yes, for this particular converting method, I think it should be `Graph[{1}, {}]`, and it certainly should be a valid `Graph` object. The way I understand it, as long as all vertices are there and the distances are preserved, it's correct.

Also, as far as I understand WL conventions, `Method` refers to a way to obtain the answer but not the answer itself, so it seems like it's not the right name here. Actually, I think it should not even be an option because it's really important. In fact, I'm not even sure it should have a default (e.g., this particular method loses information).

Also, do you think it will be better to call it `"DistancePreserving"` as this will directly explain what is it doing?

So, in short, I suggest the following syntax:
```wl
In[] := HypergraphToGraph[{{1, {3}, 2}, {2, 1, {3}}, {1, 1}}, "DistancePreserving", VertexLabels -> Automatic]
```

By the way, in case you are wondering, CI is complaining about the lack of tests.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MK2woVTChh5jZkTJ6XB:bmhqimc)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @daneelsan)

daneelsan:
The thing you said about `Method` makes total sense, and it indeed having an `Automatic` "method" is not ideal.
I will add the second argument, then.
About the tests, yes, incoming---

maxitg (changes requested):
I thought about it a bit more, and I think we should call it `"DirectedDistancePreserving"` (or something like that).

This is because the "orderedness" in the hypergraph does not necessarily have to correspond to directedness in distances (e.g., we don't expect the particle motion to respect these directions). We can then also have `"UndirectedDistancePreserving"` which would just make complete subgraphs out of every hyperedge.

Also, can we add syntax autocompletion so that we don't need to remember these names?
https://github.com/maxitg/SetReplace/blob/e78a32c085c78458792409fb77c960e96d956865/Kernel/WolframModel.m#L324-L325

And don't forget the documentation.

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MK5a5jF922qUurxjDCr:bv5knmu)* status: all files reviewed, 3 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/HypergraphToGraph.m, line 12 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5ix1U-tKMNQM23_P7:-MK5ix1U-tKMNQM23_P8:btkysqc) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L12)):*
> ```objc
> 
> (* Options *)
> Options[HypergraphToGraph] = Join[{}, Options @ Graph];
> ```

Why join an empty list?
___
*[Kernel/HypergraphToGraph.m, line 35 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5k1NZ5Z9YFZGbq9Zs:-MK5k1N_4x7_aQx9dEcP:b26586o) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L35)):*
> ```objc
> HypergraphToGraph::invalidMethod = StringJoin[{
>   "The argument at position 2 in `1` is not ",
>   Replace[ToString[#, InputForm] & /@ $validMethods, {{a__, b_} :> {Riffle[{a}, ", "], " or ", b}}],
> ```

That's a cool message constructor!
___
*[Kernel/HypergraphToGraph.m, line 50 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5kGaF82UMjZhItdKt:-MK5kGaF82UMjZhItdKu:b37i3ez) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L50)):*
> ```objc
>       2],
>     FilterRules[{opts}, Options @ Graph]]
> ```

Add an error message for unrecognized options, like

```wl
In[] := WolframModel[1 -> 2, {1}, "invalid" -> "option"]
```
___
*[Tests/HypergraphToGraph.wlt, line 81 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5ksyY9NAO88UMbz-9:-MK5ksyY9NAO88UMbz-A:b-a4r3et) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Tests/HypergraphToGraph.wlt#L81)):*
> ```Mathematica
>             DirectedEdge[1, 1]}]
>       ]
>     }
> ```

Will also be good to test `{{1, 1}}`, `{{1, 1, 1}}`, and maybe `{{1, 2, 1}}`.

daneelsan (commented):
Added argument completition.
Should "StructuredPreserving" (the method used in "HypergraphAutomorphismGroup") be called "DirectedStructuredPreserving"?

I posted examples of the 3 methods so we can spot if these are not working as intended.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MK7UqOc3DWqkDSt6Fhv:bp25pso)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/HypergraphToGraph.m, line 12 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5ix1U-tKMNQM23_P7:-MK7UiOY9EE6dNbwC9vo:b-ld7cbm) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L12)):*

Why join an empty list?
</blockquote></details>

Forgot to remove the list. Done.
___
*[Kernel/HypergraphToGraph.m, line 35 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5k1NZ5Z9YFZGbq9Zs:-MK7Vi9pDFLX29vAn2sT:bdw0g5o) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L35)):*

That's a cool message constructor!
</blockquote></details>

Thanks :sunglasses: 
___
*[Kernel/HypergraphToGraph.m, line 50 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5kGaF82UMjZhItdKt:-MK7VmRw41SesTLxNUIp:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Kernel/HypergraphToGraph.m#L50)):*

Add an error message for unrecognized options, like

```wl
In[] := WolframModel[1 -> 2, {1}, "invalid" -> "option"]
```
</blockquote></details>

Done.
___
*[Tests/HypergraphToGraph.wlt, line 81 at r2](https://reviewable.io/reviews/maxitg/setreplace/490#-MK5ksyY9NAO88UMbz-9:-MK7XvCCDvBCnGE2yfcv:bebrt1w) ([raw file](https://github.com/maxitg/setreplace/blob/eb390e0bbff45ecbd9b13c030aee8ff7d0aa6915/Tests/HypergraphToGraph.wlt#L81)):*

Will also be good to test `{{1, 1}}`, `{{1, 1, 1}}`, and maybe `{{1, 2, 1}}`.
</blockquote></details>

Will do

maxitg (changes requested):
I think `"StructurePreserving”` is ok because ordered hyperedges are the default. Note that this is a different situation than `”DirectedDistancePreserving”` because if we drop “Directed”, it would also assume how to compute distances on ordered hypergraphs. But if you want to make it more explicit, the name should be `”OrderedStructurePreserving”`.

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MK7mNcd0Ok8Kn_KKVZw:b-3s2co4)* status: all files reviewed, 6 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/HypergraphToGraph.m, line 53 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7ZbQn3VOb3e0skBqY:-MK7ZbQo82tSsQBRy7v6:b-5ok7ji) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L53)):*
> ```objc
> graphJoin[graphs : {__Graph}, opts___] := With[{
>     vertices = Sort @ Union @ Catenate[VertexList /@ graphs],
>     edges = Sort@ Catenate[EdgeList /@ graphs]},
> ```

Add a space before `@`.
___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MK7a-sv-g7hVDNqeZDk:b-46gv0p) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*
> ```objc
> (* Structure preserving *)
> hyperedgeToGraph$StructurePreserving[hyperedge_, opts___] := With[{
>     edgeVertices = Table[Unique["v", {Temporary}], Length @ hyperedge]},
> ```

Why `Temporary`? Also, do you think it would make sense to use a more descriptive name? Like, `{“Hyperedge”, hyperedgeIndex, vertexPositionIndex}`? Although if we do that, we will have to rename vertex-vertices to `{“Vertex”, vertexName}` to avoid conflicts. I kind of like this approach because it makes coloring/manipulating vertices easy. But if we don't use descriptive names, we can also use integers instead of new symbols.
___
*[Kernel/HypergraphToGraph.m, line 98 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7m1sh1ZLiQPWu6FFi:-MK7m1sh1ZLiQPWu6FFj:b-5ujsgp) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L98)):*
> ```objc
>   Graph[hyperedge, UndirectedEdge @@@ Subsets[hyperedge, {2}], opts]
> 
> hypergraphToGraph[_, hgraph_ ? hypergraphQ, "UndirectedDistancePreserving", opts : OptionsPattern[]] :=
> ```

This can be combined together with `”DirectedDistancePreserving”` to a single function.
___
*[Tests/HypergraphToGraph.wlt, line 78 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7r8q90aNC8ZKcZluO:-MK7r8q90aNC8ZKcZluP:b-wqx263) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L78)):*
> ```Mathematica
>       VerificationTest[
>         HypergraphToGraph[{Range[4]}, "DirectedDistancePreserving"],
>         Graph[{1, 2, 3, 4}, {{{1, 2}, {1, 3}, {1, 4}, {2, 3}, {2, 4}, {3, 4}}, Null}]
> ```

I think this is an undocumented syntax. Why not use `->`?
___
*[Tests/HypergraphToGraph.wlt, line 116 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7rq1d3T-4n1fO41dE:-MK7rq1d3T-4n1fO41dF:btw0bau) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L116)):*
> ```Mathematica
>         Graph[{1}, {UndirectedEdge[1, 1], UndirectedEdge[1, 2], UndirectedEdge[2, 1]}]
>       ]
>     }
> ```

Add a test with multiple edges for undirected case, and also add tests for the structure preserving method.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MK8R25b8CZGyY83tG4l:b5hsy53) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*

Why `Temporary`? Also, do you think it would make sense to use a more descriptive name? Like, `{“Hyperedge”, hyperedgeIndex, vertexPositionIndex}`? Although if we do that, we will have to rename vertex-vertices to `{“Vertex”, vertexName}` to avoid conflicts. I kind of like this approach because it makes coloring/manipulating vertices easy. But if we don't use descriptive names, we can also use integers instead of new symbols.
</blockquote></details>

`Temporary` was just a remnant from the code I copied, it makes no sense here.
I like the idea of naming the vertices explicitly (before I used `AnnotationRules`). It made the code simpler.
So instead of `{"Hyperedge", ___}` I used `"Hyperedge"[1, 2]`, let me know what you think:

```
HypergraphToGraph[{{x, x, y, z}}, "StructurePreserving", VertexLabels -> Automatic]
```

___
*[Kernel/HypergraphToGraph.m, line 98 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7m1sh1ZLiQPWu6FFi:-MK8RevSElUD0BZ9PN2I:b-urwtjm) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L98)):*

This can be combined together with `”DirectedDistancePreserving”` to a single function.
</blockquote></details>

True. I merged the internal^2 function.
___
*[Tests/HypergraphToGraph.wlt, line 78 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7r8q90aNC8ZKcZluO:-MK8RwyQ2eT3VTR-C1NZ:bf4h4o1) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L78)):*

I think this is an undocumented syntax. Why not use `->`?
</blockquote></details>

I might have looked at the `InputForm` and copied the result. I will change this.
___
*[Tests/HypergraphToGraph.wlt, line 116 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7rq1d3T-4n1fO41dE:-MK8S1A05VDZL951dZnr:bebrt1w) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L116)):*

Add a test with multiple edges for undirected case, and also add tests for the structure preserving method.
</blockquote></details>

Will do

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/HypergraphToGraph.m, line 53 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7ZbQn3VOb3e0skBqY:-MK8S6---RWPNIOT16r-:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L53)):*

Add a space before `@`.
</blockquote></details>

Done.

maxitg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MKARor-6kZPazSmMjEM:bvh1pst) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*

`Temporary` was just a remnant from the code I copied, it makes no sense here.
I like the idea of naming the vertices explicitly (before I used `AnnotationRules`). It made the code simpler.
So instead of `{"Hyperedge", ___}` I used `"Hyperedge"[1, 2]`, let me know what you think:

```
HypergraphToGraph[{{x, x, y, z}}, "StructurePreserving", VertexLabels -> Automatic]
```


</blockquote></details>

That’s an interesting way of doing it. What bothers me though is that it looks very similar to `Hyperedge[...]` which will have a different meaning. Plus, we want to keep it consistent with `”ExpressionsEventsGraph”`.
___
*[Kernel/HypergraphToGraph.m, line 66 at r5](https://reviewable.io/reviews/maxitg/setreplace/490#-MKAQVaC8G4H5fw-21Id:-MKAQVaC8G4H5fw-21Ie:b44yvz6) ([raw file](https://github.com/maxitg/setreplace/blob/f25a01a1b77d4283e493a6785033e5a5e023e826/Kernel/HypergraphToGraph.m#L66)):*
> ```objc
>   ]
> 
> hypergraphToGraph[_, hgraph_ ? hypergraphQ, "UndirectedDistancePreserving", opts : OptionsPattern[]] :=
> ```

Why not put `”UndirectedDistancePreserving” | “DirectedDistancePreserving”` here and a `Replace` or `If` in the body? That will make it even simpler.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @maxitg)

___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MKAu4b7C6c897DM_OH7:b-ikurj6) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*

That’s an interesting way of doing it. What bothers me though is that it looks very similar to `Hyperedge[...]` which will have a different meaning. Plus, we want to keep it consistent with `”ExpressionsEventsGraph”`.
</blockquote></details>

Agree with both points.

Should `VertexLabels -> Automatic` do something special?
```
labelF[{"Hyperedge", x_, y_}] := "h"[x, y]
labelF[{"Vertex", x_}] := x

GraphicsRow@{
  HypergraphToGraph[{{x, x, y, z}, {y, z, w}}, "StructurePreserving", VertexLabels -> Automatic],
  HypergraphToGraph[{{x, x, y, z}, {y, z, w}}, "StructurePreserving", VertexLabels -> Placed["Name", Below, labelF]]
  }
```

___
*[Kernel/HypergraphToGraph.m, line 98 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7m1sh1ZLiQPWu6FFi:-MKAuTXK7fyehC-D2JU5:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L98)):*

True. I merged the internal^2 function.
</blockquote></details>

Done.
___
*[Kernel/HypergraphToGraph.m, line 66 at r5](https://reviewable.io/reviews/maxitg/setreplace/490#-MKAQVaC8G4H5fw-21Id:-MKAuZgF4DJOakZGhZcB:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/f25a01a1b77d4283e493a6785033e5a5e023e826/Kernel/HypergraphToGraph.m#L66)):*

Why not put `”UndirectedDistancePreserving” | “DirectedDistancePreserving”` here and a `Replace` or `If` in the body? That will make it even simpler.
</blockquote></details>

Done.
___
*[Tests/HypergraphToGraph.wlt, line 78 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7r8q90aNC8ZKcZluO:-MKAuWhtCCBpJBOQYrW_:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L78)):*

I might have looked at the `InputForm` and copied the result. I will change this.
</blockquote></details>

Done.
___
*[Tests/HypergraphToGraph.wlt, line 116 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7rq1d3T-4n1fO41dE:-MKAuX8n66BeLseU3sP7:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L116)):*

Will do
</blockquote></details>

Done.

maxitg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 5 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MKCP8Lu1yJW8xdWz8el:bvvvtvs) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*

Agree with both points.

Should `VertexLabels -> Automatic` do something special?
```
labelF[{"Hyperedge", x_, y_}] := "h"[x, y]
labelF[{"Vertex", x_}] := x

GraphicsRow@{
  HypergraphToGraph[{{x, x, y, z}, {y, z, w}}, "StructurePreserving", VertexLabels -> Automatic],
  HypergraphToGraph[{{x, x, y, z}, {y, z, w}}, "StructurePreserving", VertexLabels -> Placed["Name", Below, labelF]]
  }
```


</blockquote></details>

It can. It’s probably not as important here as for expressions-events graph since I imagine these will mostly be used in code as an adapter for Graph-functions, but there is certainly no harm in having that, and it’s useful for documentation purposes.
___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKCSzra4BUFr7e-7Prs:b-gyucbs) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*
> ```objc
>       hyperedgeGraphs,
>       opts,
>       VertexStyle -> {{"Hyperedge", _, _} -> LightBlue},
> ```

Custom styles should go to `style.m`, and be used similarly to expressions-events graph. Sorry, I forgot about it the last time.
___
*[Tests/HypergraphToGraph.wlt, line 116 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7rq1d3T-4n1fO41dE:-MKCR_yDE6G-LGBg_IRQ:b-aff8xv) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L116)):*

Done.
</blockquote></details>

Add multi-edge structure preserving test.
___
*[Tests/HypergraphToGraph.wlt, line 135 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCQ-s02DgF8benyo3A:-MKCQ-s02DgF8benyo3B:b-h4q0gk) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L135)):*
> ```Mathematica
>       VerificationTest[
>         HypergraphToGraph[{{}}, "StructurePreserving"],
>         Graph[{}, {}]
> ```

This is an interesting case. If it’s truly structure preserving, it should be different from an empty graph. Maybe add an isolated edge-vertex for each empty hyperedge? Unless you can think of a better solution.
___
*[Tests/HypergraphToGraph.wlt, line 139 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCRgWb2tyI41ph8roB:-MKCRgWb2tyI41ph8roC:b-rgpjfd) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L139)):*
> ```Mathematica
> 
>       VerificationTest[
>         HypergraphToGraph[{{1}}, "StructurePreserving", VertexStyle -> Automatic, EdgeStyle -> Automatic],
> ```

`Automatic` is the default value for style options, and yet if you pass it as `Automatic` it changes the output. That doesn’t seem right. I think `Automatic` should make custom styles in this case (we cannot change the default option value because it’s shared with other methods).
___
*[Tests/HypergraphToGraph.wlt, line 146 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSSngC0KIY_pSK1H8:-MKCSSngC0KIY_pSK1H9:b-1di2q) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L146)):*
> ```Mathematica
> 
>       VerificationTest[
>         HypergraphToGraph[{{1, 1}}, "StructurePreserving", VertexStyle -> Automatic, EdgeStyle -> Automatic],
> ```

Also, if we are using custom styles, we should test them. Not the specific colors/styles, but the fact that they are different for different vertex/edge types.

Same if we are using custom vertex labels.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/HypergraphToGraph.m, line 68 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7a-sv-g7hVDNqeZDj:-MKQfhZoA7brIKCZ1D16:b-a2zfh8) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Kernel/HypergraphToGraph.m#L68)):*

It can. It’s probably not as important here as for expressions-events graph since I imagine these will mostly be used in code as an adapter for Graph-functions, but there is certainly no harm in having that, and it’s useful for documentation purposes.
</blockquote></details>

Decided against it. The `VertexStyle` and `EdgeStyle` should provide enough information.
___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKQg0N0E5fB1ZU4GQlX:bhsofov) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*

Custom styles should go to `style.m`, and be used similarly to expressions-events graph. Sorry, I forgot about it the last time.
</blockquote></details>

Moved them to `style.m`. Should I add them to `$styleNames`, too?
___
*[Tests/HypergraphToGraph.wlt, line 116 at r4](https://reviewable.io/reviews/maxitg/setreplace/490#-MK7rq1d3T-4n1fO41dE:-MKQg7CI3F-mq085S0P8:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57a73ea4fd72bc4b29752049e4d03d34a0644c15/Tests/HypergraphToGraph.wlt#L116)):*

Add multi-edge structure preserving test.
</blockquote></details>

Done.
___
*[Tests/HypergraphToGraph.wlt, line 135 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCQ-s02DgF8benyo3A:-MKQg8JTESS9AIB_VBCm:b-x4ll4z) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L135)):*

This is an interesting case. If it’s truly structure preserving, it should be different from an empty graph. Maybe add an isolated edge-vertex for each empty hyperedge? Unless you can think of a better solution.
</blockquote></details>

Decided to add a `{"Hyperedge", index_, 0}` at the start of every "Hyperedge" vertex. Let me know what you think.

```wl
HypergraphToGraph[
 {{x, x, y, z}, {z, w}},
 "StructurePreserving",
 VertexLabels -> 
  Placed["Name", Automatic, 
   Replace[#, {{"Hyperedge", x_, y_} :> 
       Subscript["t", x, y], {"Vertex", v_} :> "vertex"[v]}] &],
 GraphLayout -> "SpringEmbedding"]
```

___
*[Tests/HypergraphToGraph.wlt, line 139 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCRgWb2tyI41ph8roB:-MKQgMe-BLxSepdKw_bq:bhj8yjc) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L139)):*

`Automatic` is the default value for style options, and yet if you pass it as `Automatic` it changes the output. That doesn’t seem right. I think `Automatic` should make custom styles in this case (we cannot change the default option value because it’s shared with other methods).
</blockquote></details>

Yes, it was a mess. Fixed it.
```wl
In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, 
  "StructurePreserving"] === 
 HypergraphToGraph[{{x, x, y, z}, {z, w}}, "StructurePreserving", 
  VertexStyle -> Automatic, EdgeStyle -> Automatic]

Out[]= True
```
___
*[Tests/HypergraphToGraph.wlt, line 146 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSSngC0KIY_pSK1H8:-MKQgrcV9wE9OtcaaHPH:b-xyvtnk) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L146)):*

Also, if we are using custom styles, we should test them. Not the specific colors/styles, but the fact that they are different for different vertex/edge types.

Same if we are using custom vertex labels.
</blockquote></details>

So I added a simple test, but I'm unsure whether it works as you asked.

daneelsan:
I'm unsure whether the algorithm for `"UndirectedDistancePreserving"` is correct. @maxitg you mentioned that it basically meant that each hyperedge maps to a complete subgraph. If the hyperedge was `{1, 1, 2, 3}`, what would be the correct `Graph`?

Is it 
```wl
In[]:= Graph[{1 -> 1, 1 -> 2, 1 -> 3, 2 -> 3}, DirectedEdges -> False]
```
image

or
```wl
Graph[{1 -> 2, 1 -> 3, 2 -> 3}, DirectedEdges -> False]
```

maxitg (changes requested):
I think it should be either what `HypergraphToGraph[{{1, 1, 2, 3}}, "UndirectedDistancePreserving"]` currently does or the second option above. We just need to keep it predictable and as consistent as possible with `"DirectedDistancePreserving"`.

> Reviewed 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKR-D0-DNaFUhZQ9KLE:b4cgbfk)* status: all files reviewed, 6 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[README.md, line 2033 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSuhex5i89cn4LGDxl:-MKSuheyAIGZQg2G6Mad:bclc005) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2033)):*
> ```Markdown
> ### HypergraphToGraph
> 
> **`HypergraphToGraph`** converts a hypergraph to a `Graph` object. There are (currently) 3 ways to perform this transformation:
> ```

`Graph` should link to [WL documentation](https://reference.wolfram.com/language/ref/Graph.html).
___
*[README.md, line 2042 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKR0-RJ1ysXNCtBwp6p:-MKR0-RJ1ysXNCtBwp6q:bz80by7) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2042)):*
> ```Markdown
> In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, "DirectedDistancePreserving", VertexLabels -> Automatic, GraphLayout -> "SpringElectricalEmbedding"]
> ```
> ```

We need the output (as a picture).

Here is the picture:


___
*[README.md, line 2048 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwp8E6ZyY8p1Zy9sd:-MKSwp8E6ZyY8p1Zy9se:b-jkch3i) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2048)):*
> ```Markdown
> 
> ```wl
> In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, "UndirectedDistancePreserving", VertexLabels -> Automatic, GraphLayout -> "SpringElectricalEmbedding"]
> ```

Here is the picture:


___
*[README.md, line 2053 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwTegCjaHO_TGmfJP:-MKSwTegCjaHO_TGmfJQ:b-ui4jrj) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2053)):*
> ```Markdown
> #### **"StructurePreserving"**
> 
> Convert a hypergraph to a graph by preserving its structure:
> ```

This needs clarification that this kind of conversion does not lose any information, and it's possible to unambiguously convert back.
___
*[README.md, line 2056 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwqEd9KFSsnCaQ5aR:-MKSwqEe3yCx5b2MWh0V:bvris54) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2056)):*
> ```Markdown
> 
> ```wl
> In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, "StructurePreserving", VertexLabels -> Automatic]
> ```

Here is the picture:


___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKSxkH9ChYibtooSAxB:b-wx7i4c) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*

Moved them to `style.m`. Should I add them to `$styleNames`, too?
</blockquote></details>

Yes, that will be helpful.
___
*[Tests/HypergraphToGraph.wlt, line 135 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCQ-s02DgF8benyo3A:-MKSz4w08AawiocWSQPF:b-xhs920) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L135)):*

Decided to add a `{"Hyperedge", index_, 0}` at the start of every "Hyperedge" vertex. Let me know what you think.

```wl
HypergraphToGraph[
 {{x, x, y, z}, {z, w}},
 "StructurePreserving",
 VertexLabels -> 
  Placed["Name", Automatic, 
   Replace[#, {{"Hyperedge", x_, y_} :> 
       Subscript["t", x, y], {"Vertex", v_} :> "vertex"[v]}] &],
 GraphLayout -> "SpringEmbedding"]
```


</blockquote></details>

Looks good to me. Worth trying `IsomorphicHypergraphQ` though with and without it to see how much performance is affected.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 1 of 7 files reviewed, 6 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[README.md, line 2033 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSuhex5i89cn4LGDxl:-MKWL99GAbNPx9R3dHXR:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2033)):*

`Graph` should link to [WL documentation](https://reference.wolfram.com/language/ref/Graph.html).
</blockquote></details>

Done.
___
*[README.md, line 2053 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwTegCjaHO_TGmfJP:-MKWLNL842xvuAjK8L20:b-gk1pbx) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2053)):*

This needs clarification that this kind of conversion does not lose any information, and it's possible to unambiguously convert back.
</blockquote></details>

Added the comment.
___
*[README.md, line 2056 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwqEd9KFSsnCaQ5aR:-MKWLGbW-42nJU-_Yjf1:b1ocey) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2056)):*

Here is the picture:


</blockquote></details>

Thanks! Forgot to mention that I needed help with these.
I need the updated image for this, as I changed how empty hyperedges are treated.
___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKWLQ7V4ObQL-3ABpxW:bf4jlrw) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*

Yes, that will be helpful.
</blockquote></details>

I'm unsure how to do this as the custom style is only applicable when calling `"StructurePreserving"`.
___
*[Tests/HypergraphToGraph.wlt, line 135 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCQ-s02DgF8benyo3A:-MKWLXUvAbo5uRnx9kup:b-5kqzw0) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Tests/HypergraphToGraph.wlt#L135)):*

Looks good to me. Worth trying `IsomorphicHypergraphQ` though with and without it to see how much performance is affected.
</blockquote></details>

Retracted from that idea. Empty hyperedges are just isolated vertices in the graph. 
```wl
HypergraphToGraph[{{x, x, y, z}, {}, {z, w}}, "StructurePreserving", VertexLabels -> Automatic]
```

maxitg (changes requested):
> Reviewed 6 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 4 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[README.md, line 2056 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwqEd9KFSsnCaQ5aR:-MKaYl2Z3uTa5EPYTcoO:b3k1jh5) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2056)):*

Thanks! Forgot to mention that I needed help with these.
I need the updated image for this, as I changed how empty hyperedges are treated.
</blockquote></details>

Here is the update:

___
*[README.md, line 2040 at r8](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaWyaN3jHOqBrJoalA:-MKaWyaN3jHOqBrJoalB:brppkiv) ([raw file](https://github.com/maxitg/setreplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/README.md#L2040)):*
> ```Markdown
> 
> ```wl
> In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, "DirectedDistancePreserving", VertexLabels -> Automatic, GraphLayout -> "SpringElectricalEmbedding"]
> ```

This line is too long (150 chars) and does not fit in one line on GitHub. Could you split it into two lines (the lines should not exceed 120 chars and should be possible to see on a full-size desktop GitHub page without scrolling):

___
*[README.md, line 2050 at r8](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaZOahAaTbxrxDPk5n:-MKaZOahAaTbxrxDPk5o:b4gfnae) ([raw file](https://github.com/maxitg/setreplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/README.md#L2050)):*
> ```Markdown
> 
> ```wl
> In[]:= HypergraphToGraph[{{x, x, y, z}, {z, w}}, "UndirectedDistancePreserving", VertexLabels -> Automatic, GraphLayout -> "SpringElectricalEmbedding"]
> ```

This line does not fit as well.
___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKaZmev-0maz-b_keGt:b3my0gi) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*

I'm unsure how to do this as the custom style is only applicable when calling `"StructurePreserving"`.
</blockquote></details>

Well, we can add a new object type, for example, `"HypergraphStructurePreservingGraph"` (or a better name if you have one). Any custom styles will then go in it. Similar to, e.g., `"BranchialGraph"`: https://github.com/maxitg/SetReplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/Kernel/style.m#L165-L169

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[README.md, line 2056 at r7](https://reviewable.io/reviews/maxitg/setreplace/490#-MKSwqEd9KFSsnCaQ5aR:-MKalMqq-0RTW0UHr_ul:b-ucfkn4) ([raw file](https://github.com/maxitg/setreplace/blob/25cf11f033a9a3ff7765b390daa5f7c0996455d4/README.md#L2056)):*

Here is the update:

</blockquote></details>

Thanks.
___
*[README.md, line 2040 at r8](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaWyaN3jHOqBrJoalA:-MKaksaf10DQd8lpFwgW:b-gb7eiv) ([raw file](https://github.com/maxitg/setreplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/README.md#L2040)):*

This line is too long (150 chars) and does not fit in one line on GitHub. Could you split it into two lines (the lines should not exceed 120 chars and should be possible to see on a full-size desktop GitHub page without scrolling):

</blockquote></details>

Roger that.
___
*[README.md, line 2050 at r8](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaZOahAaTbxrxDPk5n:-MKalLqQ6Xw5-kfa2T18:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/README.md#L2050)):*

This line does not fit as well.
</blockquote></details>

Done.
___
*[Kernel/HypergraphToGraph.m, line 92 at r6](https://reviewable.io/reviews/maxitg/setreplace/490#-MKCSzra4BUFr7e-7Prr:-MKalTvy8wUcHoe_zKUc:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/c75f607f9994fc3588b34a108565f8a87b78c93d/Kernel/HypergraphToGraph.m#L92)):*

Well, we can add a new object type, for example, `"HypergraphStructurePreservingGraph"` (or a better name if you have one). Any custom styles will then go in it. Similar to, e.g., `"BranchialGraph"`: https://github.com/maxitg/SetReplace/blob/0763819472b545b0bf1747cfa44384b6b302dc2a/Kernel/style.m#L165-L169
</blockquote></details>

Done.

maxitg (changes requested):
I just have one suggestion remaining (sorry, missed it last time).

> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKasJagCtlP1QE-DfZ4:bnbc6vr)* status: all files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/style.m, line 307 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaqZsOEZpglyKKIkG8:-MKaqZsOEZpglyKKIkG9:bx6y66e) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L307)):*
> ```objc
> 
>   (* HypergraphToGraph Structure Preserving *)
>   $structurePreservingVertexStyle -> {{"Hyperedge", _, _} -> LightBlue},
> ```

Sorry, just noticed: this should only have styles as values. E.g., `$structurePreservingHyperedgeVertexStyle -> LightBlue`, to be consistent with `"ExpressionsEventsGraph"`, and to make it easier to get to the styles directly. The vertex patterns should only appear in the function's code.

Note that
```wl
combinedOptionsProperties[Graph][
  VertexStyle -> $structurePreservingVertexStyle,
  EdgeStyle -> $structurePreservingEdgeStyle]
```
won't work anymore in that case, but I don't really think it's necessary here.
___
*[Kernel/style.m, line 308 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKarWxk-42JSloMuafZ:-MKarWxk-42JSloMuaf_:brhumzt) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L308)):*
> ```objc
>   (* HypergraphToGraph Structure Preserving *)
>   $structurePreservingVertexStyle -> {{"Hyperedge", _, _} -> LightBlue},
>   $structurePreservingEdgeStyle -> {DirectedEdge[{"Hyperedge", _, _}, {"Hyperedge", _, _}] -> Dashed}
> ```

Same here.

taliesinb (changes requested):
A question about the name "Undirected" and "Directed". Is it clear we want to use this terminology, and we are sure what exactly it means? What is a directed versus an undirected hyperedge? As opposed to an ordered and unordered hyperedge? If all hyperedges are length 2, then there isn't any scope for the symmetry groups (either trivial or Z₂), but for larger hyperedges, and in particular for hypergraphs that contain mixtures of different length of hyperedge, there is a *lot* of freedom. Do we have a naming scheme that will scale to these different cases (and are those different cases clearly enumerated somewhere?). We don't have to figure this all out now, but it would be nice to know that the naming we pick is likely to scale.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKb0vSa5ulDcc6huDSZ:bhhgpag)* status: all files reviewed, 3 unresolved discussions (waiting on @daneelsan)

___
*[README.md, line 2037 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKb0UHh0xW-KZ3lrmAV:-MKb0UHi1MgRIsFFVl_9:bc73228) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/README.md#L2037)):*
> ```Markdown
> #### **"DirectedDistancePreserving"**
> 
> Convert a hypergraph to a directed graph with the same distance matrix:
> ```

This explanation is very sparse, in that it doesn't describe *how* the embedding occurs, e.g. that vertices and hyperedges in the hypergraph both get mapped to two different kinds of vertex in the ordinary graph. Should we explain that?

maxitg (commented):
We are using "ordered" and "unordered" for hypergraphs/hyperedges, and "directed" and "undirected" for ordinary binary graphs (as is the convention there). The only other name we have at the moment is "cyclic" which corresponds to the cyclic group.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKb2LQ0D_10nx0_zBuX:b8s28r9)* status: all files reviewed, 3 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[README.md, line 2037 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKb0UHh0xW-KZ3lrmAV:-MKb39vy-5iYBcaihz-F:b-hocmkg) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/README.md#L2037)):*

This explanation is very sparse, in that it doesn't describe *how* the embedding occurs, e.g. that vertices and hyperedges in the hypergraph both get mapped to two different kinds of vertex in the ordinary graph. Should we explain that? 
</blockquote></details>

`"DirectedDistancePreserving"` does not have the two types of vertices, and it is not a one-to-one map, it only preserves distances. Only `"StructurePreserving"` has two kinds of vertices. But I agree it would make sense to explain it there.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/style.m, line 307 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaqZsOEZpglyKKIkG8:-MKekmLl4ud_eUIjkukp:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L307)):*

Sorry, just noticed: this should only have styles as values. E.g., `$structurePreservingHyperedgeVertexStyle -> LightBlue`, to be consistent with `"ExpressionsEventsGraph"`, and to make it easier to get to the styles directly. The vertex patterns should only appear in the function's code.

Note that
```wl
combinedOptionsProperties[Graph][
  VertexStyle -> $structurePreservingVertexStyle,
  EdgeStyle -> $structurePreservingEdgeStyle]
```
won't work anymore in that case, but I don't really think it's necessary here.
</blockquote></details>

Done.
___
*[Kernel/style.m, line 308 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKarWxk-42JSloMuafZ:-MKekn-wBaE9JhGzkXGx:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L308)):*

Same here.
</blockquote></details>

Done.

daneelsan (commented):
Perhaps that convention ("Directed" & "Undirected" vs "Ordered" & "Unordered") should be explained in the Glossary?

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKerXDGA4gG_swBYCa7:b-yqrn9l)* status: 4 of 11 files reviewed, 3 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[README.md, line 2037 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKb0UHh0xW-KZ3lrmAV:-MKerPgaEByyr0AIsjCs:b-1cnfsy) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/README.md#L2037)):*

`"DirectedDistancePreserving"` does not have the two types of vertices, and it is not a one-to-one map, it only preserves distances. Only `"StructurePreserving"` has two kinds of vertices. But I agree it would make sense to explain it there.
</blockquote></details>

I expanded the explanation of `"StructurePreserving"`.

maxitg (changes requested):
> Reviewed 2 of 2 files at r11, 1 of 1 files at r12, 4 of 4 files at r13.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/style.m, line 307 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaqZsOEZpglyKKIkG8:-MKemZAs5nn5K2j1Hvq-:bxcq5fv) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L307)):*

Done.
</blockquote></details>

But it's not all of the edges and vertices. It should mention the kind of edges/vertices in the name (and in the association keys).

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/style.m, line 307 at r10](https://reviewable.io/reviews/maxitg/setreplace/490#-MKaqZsOEZpglyKKIkG8:-MKfRZtn9zxKmiHnQSEg:b-2dmjfd) ([raw file](https://github.com/maxitg/setreplace/blob/0321acd7e66d385ac9626e9115cea673febf4346/Kernel/style.m#L307)):*

But it's not all of the edges and vertices. It should mention the kind of edges/vertices in the name (and in the association keys).
</blockquote></details>

Didn't notice you wrote `$structurePreservingHyperedgeVertexStyle`.
Done.

maxitg (dismissed):
> Reviewed 2 of 2 files at r14.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 1 unresolved discussion (waiting on @taliesinb)

taliesinb (commented):
I'm somewhat unsure about the string method names but will discuss on SetReplace rather.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKfVYJVDMDw83485P67:b8ed59o)* status: all files reviewed, 1 unresolved discussion (waiting on @daneelsan)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md, line 23 at r14](https://reviewable.io/reviews/maxitg/setreplace/490#-MKfUb-DDuZpn4-kIfYC:-MKfUb-DDuZpn4-kIfYD:b7u0vdb) ([raw file](https://github.com/maxitg/setreplace/blob/a62871172e0b22877efce00f2d316b7234d8d25e/Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md#L23)):*
> ```Markdown
> ## **"UndirectedDistancePreserving"**
> 
> Convert a hypergraph to an undirected graph with the same distance matrix, that is, each hyperedge is mapped to a complete subgraph:
> ```

Great. One last thing, I should have made this comment before as well, but it wasn't as clear to me then what was going on. 

So my understanding is that for "UndirectedDistancePreserving" the vertices are the same between the hypergraph and the graph? E.g. VertexList[hypergraph] == VertexList[graph]? And a single hyperedge {1, 2, 3} gets converted into the set of edges {1, 2}, {1, 3}, {2, 3}? Whereas I imagine with "DirectedDistancePreserving" it gets converted into the set of edges {1, 2}, {2, 3}? Again, it's easy for us to say that and wasn't obvious to me.

Again, can we not just say this in the documentation? It's simple enough to spell out... Not going to insist on it but I do think it's worth doing.

taliesinb (approved):
Ah, I did not see the extensive commentary about naming already, my apologies. I suck at Reviewable. I've missed the boat there, so just going to approve (my request for slightly expanded documentation is totally optional).

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKfXdR72M8p6cfZjRP_:b-luokdd)* status: all files reviewed, 1 unresolved discussion (waiting on @daneelsan)

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 1 unresolved discussion (waiting on @taliesinb)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md, line 23 at r14](https://reviewable.io/reviews/maxitg/setreplace/490#-MKfUb-DDuZpn4-kIfYC:-MKfbMXc88KuxuzJdJ20:b31c8tj) ([raw file](https://github.com/maxitg/setreplace/blob/a62871172e0b22877efce00f2d316b7234d8d25e/Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md#L23)):*

Great. One last thing, I should have made this comment before as well, but it wasn't as clear to me then what was going on. 

So my understanding is that for "UndirectedDistancePreserving" the vertices are the same between the hypergraph and the graph? E.g. VertexList[hypergraph] == VertexList[graph]? And a single hyperedge {1, 2, 3} gets converted into the set of edges {1, 2}, {1, 3}, {2, 3}? Whereas I imagine with "DirectedDistancePreserving" it gets converted into the set of edges {1, 2}, {2, 3}? Again, it's easy for us to say that and wasn't obvious to me.

Again, can we not just say this in the documentation? It's simple enough to spell out... Not going to insist on it but I do think it's worth doing. 
</blockquote></details>

No, in the directed case, it gets converted to `{{1, 2}, {1, 3}, [2, 3}}`. Directed distance is assumed to be one if there is only one hyperedge that needs to be traversed. And I agree, it would make sense to mention this in the documentation.

maxitg (changes requested):
> Reviewed 1 of 1 files at r15.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: all files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md, line 9 at r16](https://reviewable.io/reviews/maxitg/setreplace/490#-MKqV8Tm08b5zGIiWubC:-MKqV8Tm08b5zGIiWubD:b-7ehi1z) ([raw file](https://github.com/maxitg/setreplace/blob/c21c512b5b6424e874ab50ca2e4d12b9a96e6ee7/Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md#L9)):*
> ```Markdown
> ## **"DirectedDistancePreserving"**
> 
> Converts a hypergraph to a directed graph with the same distance matrix, where distance between two vertices `v1` and `v2` in a hypergraph is defined as the minimum length of a path which connects `v1` and `v2` (i.e. 1 if both vertices belong to the same hyperedge):
> ```

Either change to "Converts" -> "Convert" or change "Convert" -> "Converts" below.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md, line 9 at r16](https://reviewable.io/reviews/maxitg/setreplace/490#-MKqV8Tm08b5zGIiWubC:-MKqW3Lj1NBw6nUgdRLD:boj1qgt) ([raw file](https://github.com/maxitg/setreplace/blob/c21c512b5b6424e874ab50ca2e4d12b9a96e6ee7/Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md#L9)):*

Either change to "Converts" -> "Convert" or change "Convert" -> "Converts" below.
</blockquote></details>

Good catch. Done.

maxitg (approved):
:lgtm:

> Reviewed 1 of 1 files at r17.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/490#-:-MKqWQ0B2cSMPhB0YkIF:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @taliesinb)

PR #506, opened 2020-11-02
Rasterization functions to produce inputs & outputs

taliesinb:
This change introduces rasterization functions to produce input, output, and input output pairs, including the little `In[*]`, `Out[*]` cell labels. These are produced at high DPI resolutions.

Here's a screenshot of the intended behavior:

PixelSnap 2020-11-02 at 21 25 37@2x

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -465,17 +465,19 @@ code.
 #### Markdown
 We are using GitHub-flavored Markdown for documentation and research notes.
 
-Images (e.g., of output cells) should be made by selecting the relevant cells in the Front End, copying them as bitmaps, and saving them as .png files to [Documentation/Images](/Documentation/Images) (in the documentation) or to the Images directory of the corresponding research note. They should then be inserted using the code similar to this:
+Images of input and/or output cells can be easily produced using Mathematica by running the command 
+`RasterizePreviousCell[]` in the cell that follows the output you want to capture. These images should be saved 
+to [Documentation/Images](/Documentation/Images) as PNG files. This can be done totally automatically using 
+the function `ExportImageForEmbedding["name", image]`, where "name" should be a CamelCaseDescriptiveName that 
+describes the content and intent of the image. The following code captures the previous cell and saves it in one go:
 
-  ```html
-  <img src="/Documentation/Images/image.png" width="xxx">
-  ```
+```wl
+ExportImageForEmbedding["AGoodNameGoesHere", RasterizePreviousCell[]]
+```
 
-  where the `width` should be computed as
-
-  ```wl
-  Round[0.6 First @ Import["$RepoRoot/Documentation/Images/image.png", "ImageSize"]]
-  ```
+This function will return markdown-compatible code that you can paste straight into the relevant readme or research
+note. In fact, the function will copy this code straight into your clipboard so that you can simply paste it into
+the markdown file. If you wish to capture an output without the Out[] tag, and without the corresponding input cell, you can use `RasterizeExpression[output]` in place of `RasterizePreviousCell[]`.
 
 # Research
 
--- a/Kernel/init.m
+++ b/Kernel/init.m
@@ -3,7 +3,11 @@ Unprotect["SetReplace`*"];
 ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
 
 Block[
-  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {}},
+  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
+      HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
+	    HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
+	    HoldPattern[ReturnFailure[msg_String, args___]] :> ReturnFailure[MessageName[$LHSHead, msg], args]
+  }},
   Get[FileNameJoin[{FileNameDrop[$InputFileName], "usageString.m"}]];
 ];
 
--- a/Kernel/init.m
+++ b/Kernel/init.m
@@ -5,7 +5,7 @@ ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
 SetReplace`$SetReplaceBaseDirectory = FileNameDrop[$InputFileName, -2];
 
 Block[
-  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
+  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {
       HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
 	    HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
 	    HoldPattern[ReturnFailure[msg_String, args___]] :> ReturnFailure[MessageName[$LHSHead, msg], args]
--- a/Kernel/init.m
+++ b/Kernel/init.m
@@ -2,13 +2,15 @@ Unprotect["SetReplace`*"];
 
 ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
 
+SetReplace`$SetReplaceBaseDirectory = FileNameDrop[$InputFileName, -2];
+
 Block[
   {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
       HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
 	    HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
 	    HoldPattern[ReturnFailure[msg_String, args___]] :> ReturnFailure[MessageName[$LHSHead, msg], args]
   }},
-  Get[FileNameJoin[{FileNameDrop[$InputFileName], "usageString.m"}]];
+  Get[FileNameJoin[{SetReplace`$SetReplaceBaseDirectory, "Kernel", "usageString.m"}]];
 ];
 
 SetAttributes[#, {Protected, ReadProtected}] & /@ Evaluate @ Names @ "SetReplace`*";
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -6,8 +6,6 @@ PackageExport["RasterizeAsOutput"]
 PackageExport["RasterizeAsInput"]
 PackageExport["RasterizeAsInputOutputPair"]
 
-ClearAll[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair]
-
 SetRelatedSymbolGroup[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair]
 
 cellToExportPacket[cell_] := ExportPacket[cell,
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -8,15 +8,17 @@ PackageExport["RasterizeAsInputOutputPair"]
 
 SetRelatedSymbolGroup[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair]
 
-cellToExportPacket[cell_] := ExportPacket[cell,
+rasterizeCell[cell_] := bitmapToImage @ MathLink`CallFrontEnd @ ExportPacket[cell,
   "BitmapPacket", ColorSpace -> RGBColor, "AlphaChannel" -> False,
-  "DataCompression" -> True, ImageResolution -> 144
+  "DataCompression" -> True, ImageResolution -> 144 (* 144, being 2x72, is the resolution for Retina displays *)
 ];
 
 bitmapToImage[System`ConvertersDump`Bitmap[rawString_, {width_, height_, depth_}, ___]] := Scope[
   bytes = NumericArray[Developer`RawUncompress @ rawString, "Byte"];
   Internal`ArrayReshapeTo[bytes, {height, width, depth}];
-  Image[Image`ReverseNumericArray[bytes, False], Interleaving -> True, Magnification -> 0.5]
+  (* we use magnification of 1/2 to compensate for the double-resolution of retina display,
+  ensuring the image *displays* at the normal size in the front end *)
+  Image[Image`ReverseNumericArray[bytes, False], Interleaving -> True, Magnification -> 1/2]
 ]
 
 toCell[expr_, type_, isize_] := Cell[
@@ -26,10 +28,8 @@ toCell[expr_, type_, isize_] := Cell[
   GraphicsBoxOptions -> {ImageSize -> isize}, Graphics3DBoxOptions -> {ImageSize -> isize}
 ];
 
-rasterizeCell[cell_Cell] := 
-	bitmapToImage @ MathLink`CallFrontEnd @ cellToExportPacket @ cell;
-
-fastRasterize[expr_, type_:"Output", size_:Medium] := rasterizeCell @ toCell[Unevaluated @ expr, type, size];
+rasterizeExpr[expr_, type_:"Output"] := 
+  rasterizeCell @ toCell[Unevaluated @ expr, type, Small];
 
 cellLabelToImage[text_] := With[{img = rasterizeCell @ Cell[text, "CellLabel", "CellLabelExpired"]}, 
 	ImagePad[img, {{80 - ImageDimensions[img][[1]], 15}, {0, 10}}, White]];
@@ -37,7 +37,8 @@ cellLabelToImage[text_] := With[{img = rasterizeCell @ Cell[text, "CellLabel", "
 $outputCellLabel := $outputCellLabel = cellLabelToImage["Out[\:f759\:f363]="];
 $inputCellLabel := $inputCellLabel = cellLabelToImage["In[\:f759\:f363]:="];
 
-assembleWithLabel[label_, cell_] := ImageAssemble[{{label, cell}}, "Fit", Background -> White];
+assembleWithLabel[label_, cell_] := 
+  ImageAssemble[{{label, cell}}, "Fit", Background -> White, Magnification -> 1/2];
 
 SetUsage @ "
 RasterizeAsOutput[expr$] creates an Image of expr$, formatted graphically as an \"Output\" cell.
@@ -47,7 +48,8 @@ RasterizeAsOutput[expr$] creates an Image of expr$, formatted graphically as an 
 
 SyntaxInformation[RasterizeAsOutput] = {"ArgumentsPattern" -> {_}};
 
-RasterizeAsOutput[expr_] := assembleWithLabel[$outputCellLabel, fastRasterize[expr, "Output"]];
+RasterizeAsOutput[expr_] := 
+  assembleWithLabel[$outputCellLabel, rasterizeExpr[expr, "Output"]];
 
 SetUsage @ "
 RasterizeAsInput[expr$] creates an Image of expr$, formatted textually as an \"Input\" cell.
@@ -59,11 +61,12 @@ RasterizeAsInput[expr$] creates an Image of expr$, formatted textually as an \"I
 SyntaxInformation[RasterizeAsInput] = {"ArgumentsPattern" -> {_}};
 
 SetAttributes[{RasterizeAsInput, RasterizeAsInputOutputPair}, HoldFirst];
-RasterizeAsInput[expr_] := assembleWithLabel[$inputCellLabel, fastRasterize[Unevaluated @ expr, "Input"]];
+RasterizeAsInput[expr_] := 
+  assembleWithLabel[$inputCellLabel, rasterizeExpr[Unevaluated @ expr, "Input"]];
 
-imageColumn[images_] := With[
+imageColumn[images_, opts___Rule] := With[
   {width = Max[First /@ ImageDimensions /@ images]}, 
-  ImageAssemble[{ImageCrop[#, {width, Full}, Left, Padding -> White]}& /@ images]
+  ImageAssemble[{ImageCrop[#, {width, Full}, Left, Padding -> White]}& /@ images, opts]
 ];
 
 SetUsage @ "
@@ -77,4 +80,5 @@ and the corresponding resulting output expression.
 
 SyntaxInformation[RasterizeAsInputOutputPair] = {"ArgumentsPattern" -> {_}};
 
-RasterizeAsInputOutputPair[expr_] := imageColumn[{RasterizeAsInput[expr], RasterizeAsOutput[expr]}];
+RasterizeAsInputOutputPair[expr_] := 
+  imageColumn[{RasterizeAsInput[expr], RasterizeAsOutput[expr]}, Magnification -> 1/2];
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -73,8 +73,8 @@ SetUsage @ "
 RasterizeAsInputOutputPair[expr$] creates an Image of expr$ formatted as both an input, \
 and the corresponding resulting output expression.
 * RasterizeAsInputOutputPair Holds expr$, preventing it from evaluating.
-* RasterAsInput[expr$] is used to create the Image of the input.
-* RasterAsOutput[expr$] is used to create the Image of the result.
+* RasterizeAsInput[expr$] is used to create the Image of the input.
+* RasterizeAsOutput[expr$] is used to create the Image of the result.
 * The Image is produced at high DPI, suitable for Retina displays.
 "
 
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -45,6 +45,8 @@ RasterizeAsOutput[expr$] creates an Image of expr$, formatted graphically as an 
 * The Image is produced at high DPI, suitable for Retina displays.
 "
 
+SyntaxInformation[RasterizeAsOutput] = {"ArgumentsPattern" -> {_}};
+
 RasterizeAsOutput[expr_] := assembleWithLabel[$outputCellLabel, fastRasterize[expr, "Output"]];
 
 SetUsage @ "
@@ -53,6 +55,8 @@ RasterizeAsInput[expr$] creates an Image of expr$, formatted textually as an \"I
 * RasterizeAsInput uses PrettyForm to automatically format the given input code.
 * The Image is produced at high DPI, suitable for Retina displays.
 "
+
+SyntaxInformation[RasterizeAsInput] = {"ArgumentsPattern" -> {_}};
 
 SetAttributes[{RasterizeAsInput, RasterizeAsInputOutputPair}, HoldFirst];
 RasterizeAsInput[expr_] := assembleWithLabel[$inputCellLabel, fastRasterize[Unevaluated @ expr, "Input"]];
@@ -70,5 +74,7 @@ and the corresponding resulting output expression.
 * RasterAsOutput[expr$] is used to create the Image of the result.
 * The Image is produced at high DPI, suitable for Retina displays.
 "
+
+SyntaxInformation[RasterizeAsInputOutputPair] = {"ArgumentsPattern" -> {_}};
 
 RasterizeAsInputOutputPair[expr_] := imageColumn[{RasterizeAsInput[expr], RasterizeAsOutput[expr]}];
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -186,21 +186,44 @@ correctly render on e.g. GitHub but may not render in e.g. VSCode.
 * The resulting markdown is placed on the system clipboard, ready to be pasted into a file.
 "
 
-SyntaxInformation[ExportImageForEmbedding] = {"ArgumentsPattern" -> {_, _}};
+SyntaxInformation[ExportImageForEmbedding] = {"ArgumentsPattern" -> {_, _, OptionsPattern[]}};
 
 ExportImageForEmbedding::nodot = "The name \"``\" should not contain a file extension, since one will be added automatically.";
 ExportImageForEmbedding::notcap = "The name \"``\" should start with a capital letter and not contain any spaces.";
 ExportImageForEmbedding::noexport = "The image could not be exported.";
+ExportImageForEmbedding::paclet = "You have apparently loaded an installed, pacletized version of SetReplace (located at ``).
+This means that ExportImageForEmbedding does not know where your development directory is, and hence where to save images.
+Either load SetReplace directly from the Git repo, using e.g. Get[\"~/git/SetReplace/Kernel/init.m\"], or provide the directory \
+of your checkout of the Git repository manually using the \"DevelopmentDirectory\" option.";
+ExportImageForEmbedding::baddevdir = "The provided development directory `` does not exist or is not a string."
+ExportImageForEmbedding::noimgpath = "The directory in which to save images was expected to exist at \"``\", but was not found.";
 
-$ImagesDirectory = FileNameJoin[{$SetReplaceBaseDirectory, "Documentation", "Images"}];
 $ImageMarkdownTemplate = StringTemplate["<img src=\"/Documentation/Images/``\" width=\"``\">"];
+
+Options[ExportImageForEmbedding] = {
+  "DevelopmentDirectory" -> Automatic,
+  CompressionLevel -> 1.0,
+  "ColorMapLength" -> Automatic
+}
 
 ExportImageForEmbedding[name_String, image_Image, opts:OptionsPattern[]] := Scope[
   If[StringContainsQ[name, "."], ReturnFailed["nodot", name]];
   If[name === "" || !UpperCaseQ[StringTake[name, 1]] || StringContainsQ[name, " "], ReturnFailed["notcap", name]];
   filename = name <> ".png";
-  path = FileNameJoin[{$ImagesDirectory, filename}];
-  result = Export[path, image, opts, CompressionLevel -> 1.0];
+  UnpackOptions[developmentDirectory, compressionLevel, colorMapLength];
+  SetAutomatic[developmentDirectory, $SetReplaceBaseDirectory];
+  If[!StringQ[developmentDirectory] || !DirectoryQ[developmentDirectory],
+    ReturnFailed["baddevdir", developmentDirectory]
+  ];
+  imagesDirectory = FileNameJoin[{developmentDirectory, "Documentation", "Images"}];
+  If[!DirectoryQ[imagesDirectory],
+    If[StringStartsQ[developmentDirectory, {$UserBasePacletsDirectory, $BasePacletsDirectory}],
+      ReturnFailed["paclet", $SetReplaceBaseDirectory],
+      ReturnFailed["noimgpath", imagesDirectory]
+    ]
+  ];
+  path = FileNameJoin[{imagesDirectory, filename}];
+  result = Export[path, image, CompressionLevel -> compressionLevel, "ColorMapLength" -> colorMapLength];
   If[!StringQ[result], ReturnFailed["noexport"]];
   width = First @ ImageDimensions[image];
   markdown = $ImageMarkdownTemplate[filename, width / 2.];
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -24,7 +24,7 @@ SetAttributes[MakeFormattedCodeBoxes, HoldFirst];
 MakeFormattedCodeBoxes[expr_, n_:60] := Block[
   {GeneralUtilities`Formatting`PackagePrivate`$prettyFormWidth = n,
    GeneralUtilities`PackageScope`$enableReflow = False},
-	MakeFormattedBoxes[expr] /. symbol_String /;
+  MakeFormattedBoxes[expr] /. symbol_String /;
     StringMatchQ[symbol, ("GeneralUtilities`" | "SetReplace`") ~~ Except["`"]..] :>
       StringExtract[symbol, "`" -> 2]
   (* MakeFormattedBoxes fully qualifies all symbols, we strip off these two since
@@ -83,7 +83,7 @@ rasterizeExpr[expr_, type_:"Output"] :=
 
 cellLabelToImage[text_] := With[
   {img = RasterizeCell @ Cell[text, "CellLabel", "CellLabelExpired"]},
-	ImagePad[img, {{80 - ImageDimensions[img][[1]], 15}, {0, 10}}, White]];
+  ImagePad[img, {{80 - ImageDimensions[img][[1]], 15}, {0, 10}}, White]];
 
 $outputCellLabel := $outputCellLabel = cellLabelToImage["Out[\:f759\:f363]="];
 $inputCellLabel := $inputCellLabel = cellLabelToImage["In[\:f759\:f363]:="];
@@ -154,13 +154,13 @@ image and copy a markdown-compatible <img> tag to the clipboard that displays th
 "
 
 RasterizePreviousCell[] := Scope[
-	prev1 = PreviousCell[]; prev2 = PreviousCell @ prev1;
-	prev1 = NotebookRead[prev1]; prev2 = NotebookRead[prev2];
-	If[Head[prev1] =!= Cell, Return[$Failed]];
-	If[cellType[prev1] === "Output" && cellType[prev2] === "Input",
-		RasterizeCell[{prev2, prev1}],
-		RasterizeCell[prev1]
-	]
+  prev1 = PreviousCell[]; prev2 = PreviousCell @ prev1;
+  prev1 = NotebookRead[prev1]; prev2 = NotebookRead[prev2];
+  If[Head[prev1] =!= Cell, Return[$Failed]];
+  If[cellType[prev1] === "Output" && cellType[prev2] === "Input",
+    RasterizeCell[{prev2, prev1}],
+    RasterizeCell[prev1]
+  ]
 ];
 
 SetUsage @ "
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -49,7 +49,7 @@ RasterizeAsOutput[expr$] creates an Image of expr$, formatted graphically as an 
 
 SyntaxInformation[RasterizeAsOutput] = {"ArgumentsPattern" -> {_}};
 
-RasterizeAsOutput[expr_] := 
+RasterizeAsOutput[expr_] :=
   assembleWithLabel[$outputCellLabel, rasterizeExpr[expr, "Output"]];
 
 SetUsage @ "
@@ -81,7 +81,7 @@ and the corresponding resulting output expression.
 
 SyntaxInformation[RasterizeAsInputOutputPair] = {"ArgumentsPattern" -> {_}};
 
-RasterizeAsInputOutputPair[expr_] := 
+RasterizeAsInputOutputPair[expr_] :=
   imageColumn[{RasterizeAsInput[expr], RasterizeAsOutput[expr]}, Magnification -> 1/2];
 
 SetUsage @ "
--- a/Kernel/rasterization.m
+++ b/Kernel/rasterization.m
@@ -5,8 +5,9 @@ PackageImport["GeneralUtilities`"]
 PackageExport["RasterizeAsOutput"]
 PackageExport["RasterizeAsInput"]
 PackageExport["RasterizeAsInputOutputPair"]
+PackageExport["ExportImageForEmbedding"]
 
-SetRelatedSymbolGroup[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair]
+SetRelatedSymbolGroup[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair, ExportImageForEmbedding]
 
 rasterizeCell[cell_] := bitmapToImage @ MathLink`CallFrontEnd @ ExportPacket[cell,
   "BitmapPacket", ColorSpace -> RGBColor, "AlphaChannel" -> False,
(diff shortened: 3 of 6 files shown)


daneelsan (changes requested):
The builds are failing because:
1) There are no `SyntaxInformation` for the functions, and
2) There are no tests (though I'm not sure how to test these)

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506#-:-ML9eUgJAqq7H2MyE3ej:beau7yd)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[Kernel/rasterization.m, line 9 at r1](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9eFm40mx2K6S5Hj74:-ML9eFm40mx2K6S5Hj75:bk06ay1) ([raw file](https://github.com/maxitg/setreplace/blob/5bd7c2bcadd953774b5475ed19cc6065299e9448/Kernel/rasterization.m#L9)):*
> ```objc
> PackageExport["RasterizeAsInputOutputPair"]
> 
> ClearAll[RasterizeAsOutput, RasterizeAsInput, RasterizeAsInputOutputPair]
> ```

I think this is not needed (the init file already does some `ClearAll`).

taliesinb (commented):
I've added SyntaxInformation for the 3 functions, but I don't think we can reasonably test these functions, since the output of rasterization varies platform-to-platform, but also because it would require a frontend, which isn't available when running the CI.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506#-:-ML9fdN400fXKHsUdENU:bxoazur)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @daneelsan and @maxitg)

___
*[Kernel/rasterization.m, line 9 at r1](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9eFm40mx2K6S5Hj74:-ML9fbzlCzig69uBUBh7:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/5bd7c2bcadd953774b5475ed19cc6065299e9448/Kernel/rasterization.m#L9)):*

I think this is not needed (the init file already does some `ClearAll`).
</blockquote></details>

Done.

maxitg (changes requested):
The frontend is available (I was surprised as well, but it somehow works with WolframScript). Check [`graphicsQ`](https://github.com/maxitg/SetReplace/blob/master/Kernel/testUtilities.m) function.

We cannot reasonably test whether the pictures are correct. But we should test that the functions evaluate and produce an image.

I think another thing we can do is make a grayscale image and check there are no colors (to check there are no pink background in the output). Also, we can check that the image size is large enough to fit In[]/Out[] at a correct resolution.

Beyond that, I don't think we can do anything with a reasonable amount of effort.

Also, [contributing](https://github.com/maxitg/SetReplace/blob/master/.github/CONTRIBUTING.md#markdown) should be updated to tell people to use these functions.

Otherwise, this is great! Very useful.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506#-:-ML9gVOe5h9aOAi1ZigC:b-r3rg2g)* status: all files reviewed, 11 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/rasterization.m, line 13 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9n5gt3eRLpNDm1bwz:-ML9n5gt3eRLpNDm1bx-:b-ngmd6s) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L13)):*
> ```objc
> cellToExportPacket[cell_] := ExportPacket[cell,
>   "BitmapPacket", ColorSpace -> RGBColor, "AlphaChannel" -> False,
>   "DataCompression" -> True, ImageResolution -> 144
> ```

Nit: Maybe add a comment mentioning this is the Retina-default resolution?
___
*[Kernel/rasterization.m, line 19 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9nO5V8Ysz8Nla7l03:-ML9nO5V8Ysz8Nla7l04:blobpzi) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L19)):*
> ```objc
>   bytes = NumericArray[Developer`RawUncompress @ rawString, "Byte"];
>   Internal`ArrayReshapeTo[bytes, {height, width, depth}];
>   Image[Image`ReverseNumericArray[bytes, False], Interleaving -> True, Magnification -> 0.5]
> ```

Nit: Add a comment about `Magnification`?
___
*[Kernel/rasterization.m, line 25 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ne4IDg2832-kWCJy:-ML9ne4IDg2832-kWCJz:b-jb8c1m) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L25)):*
> ```objc
>   BoxData[If[type === "Input", PrettyFormBoxes @ expr, ToBoxes[Unevaluated @ expr]]],
>   type, ShowCellBracket -> False, Background -> Automatic, CellMargins -> 0,
>   CellFrameMargins -> 0, CellContext -> "Global`", 
> ```

Remove the trailing space.
___
*[Kernel/rasterization.m, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ntnLBE0PMXZRsBC_:-ML9ntnLBE0PMXZRsBCa:b-jb8c1m) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L29)):*
> ```objc
> ];
> 
> rasterizeCell[cell_Cell] := 
> ```

Remove the trailing space.
___
*[Kernel/rasterization.m, line 30 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9nun8CuZngAra1FtW:-ML9nun8CuZngAra1FtX:bgg0m2n) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L30)):*
> ```objc
> 
> rasterizeCell[cell_Cell] := 
> 	bitmapToImage @ MathLink`CallFrontEnd @ cellToExportPacket @ cell;
> ```

Tabs -> spaces
___
*[Kernel/rasterization.m, line 34 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9oJ6a5-3z40-ZjmHK:-ML9oJ6a5-3z40-ZjmHL:b-jb8c1m) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L34)):*
> ```objc
> fastRasterize[expr_, type_:"Output", size_:Medium] := rasterizeCell @ toCell[Unevaluated @ expr, type, size];
> 
> cellLabelToImage[text_] := With[{img = rasterizeCell @ Cell[text, "CellLabel", "CellLabelExpired"]}, 
> ```

Remove the trailing space.
___
*[Kernel/rasterization.m, line 35 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9oKAvCWLWD574snsy:-ML9oKAvCWLWD574snsz:bckqmhb) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L35)):*
> ```objc
> 
> cellLabelToImage[text_] := With[{img = rasterizeCell @ Cell[text, "CellLabel", "CellLabelExpired"]}, 
> 	ImagePad[img, {{80 - ImageDimensions[img][[1]], 15}, {0, 10}}, White]];
> ```

Tabs -> spaces.
___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-ML9l41z6XF_bDaaYYnO:b-np5gkw) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*
> ```objc
> RasterizeAsInput[expr$] creates an Image of expr$, formatted textually as an \"Input\" cell.
> * RasterizeAsInput Holds expr$, preventing it from evaluating.
> * RasterizeAsInput uses PrettyForm to automatically format the given input code.
> ```

It should not disable custom boxes. E.g., this should not print the inside of the evolution object as we use inputs like this in the documentation (even though I admit the formatting looks nice in this case):

___
*[Kernel/rasterization.m, line 66 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9osf-8NC3Q0dntsUm:-ML9osf-8NC3Q0dntsUn:b-jpgcig) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L66)):*
> ```objc
> imageColumn[images_] := With[
>   {width = Max[First /@ ImageDimensions /@ images]}, 
>   ImageAssemble[{ImageCrop[#, {width, Full}, Left, Padding -> White]}& /@ images]
> ```

Add a space before `&` (for consistency).
___
*[Kernel/rasterization.m, line 80 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ovEAD_4t2NUJ4ba0:-ML9ovEBEe6B51n8E246:b-p0tqwv) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L80)):*
> ```objc
> SyntaxInformation[RasterizeAsInputOutputPair] = {"ArgumentsPattern" -> {_}};
> 
> RasterizeAsInputOutputPair[expr_] := imageColumn[{RasterizeAsInput[expr], RasterizeAsOutput[expr]}];
> ```

Add an empty line at the end of the file (for consistency).

taliesinb (commented):
I've added a new function called ExportImageForEmbedding which more directly solves the annoying process of creating images and embedding them in markdown. You use it in combination with the existing rasterization functions, for example:



> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506#-:-MLCkVPU12-YlWmACD7M:bcx84gq)* status: 0 of 5 files reviewed, 10 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/rasterization.m, line 19 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9nO5V8Ysz8Nla7l03:-MLCQVBgAJFS7LMMMa-g:b-ja5vyh) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L19)):*

Nit: Add a comment about `Magnification`?
</blockquote></details>

It turns out the Magnification code wasn't actually even working properly! I've commented and fixed it. The example I pasted in the PR now looks 'normal size' thanks to the fix.
___
*[Kernel/rasterization.m, line 25 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ne4IDg2832-kWCJy:-MLCQUM6Ec4lRr--wlo5:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L25)):*

Remove the trailing space.
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ntnLBE0PMXZRsBC_:-MLCQTvc28cfhIgKIljW:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L29)):*

Remove the trailing space.
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 30 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9nun8CuZngAra1FtW:-MLCQTiD-mAMB5x9J5-S:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L30)):*

Tabs -> spaces
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 34 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9oJ6a5-3z40-ZjmHK:-MLCQTCD9vSikF2JJOFD:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L34)):*

Remove the trailing space.
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 35 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9oKAvCWLWD574snsy:-MLCQStG1F7E8vppdSwg:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L35)):*

Tabs -> spaces.
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-MLCPTAI-2GSaaxdsS1O:bi6b0ww) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*

It should not disable custom boxes. E.g., this should not print the inside of the evolution object as we use inputs like this in the documentation (even though I admit the formatting looks nice in this case):

</blockquote></details>

I don't see how this can be done consistently, since if I render the expression as StandardForm, expressions like Hold[Graphics[{}]] will incorrectly end up rendering the inner graphics (which may pink itself etc because they aren't correct). Another way of saying this is that there is technically *more* information available in an actual cell in an actual notebook then can be captured by an expression sent to RasterizeAsInput -- specifically the box structure, which in your example contains the typesetting. I can't faithfully reconstruct that from just an expression. 

So if you want to make a raster of an *actual* input + output cell, that should probably be a separate function that targets a particular notebook and makes rasters of its inputs and outputs. I can do this but I'd suggest it be done in a separate PR.
___
*[Kernel/rasterization.m, line 66 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9osf-8NC3Q0dntsUm:-MLCQRkiCSmR3VwpgKdb:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L66)):*

Add a space before `&` (for consistency).
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 80 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9ovEAD_4t2NUJ4ba0:-MLCQSHd-jQBpGxhqOMW:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L80)):*

Add an empty line at the end of the file (for consistency).
</blockquote></details>

Done.

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 5 files reviewed, 10 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-MLD7LDCAI2lhXAB7jB8:b-e1aelq) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*

I don't see how this can be done consistently, since if I render the expression as StandardForm, expressions like Hold[Graphics[{}]] will incorrectly end up rendering the inner graphics (which may pink itself etc because they aren't correct). Another way of saying this is that there is technically *more* information available in an actual cell in an actual notebook then can be captured by an expression sent to RasterizeAsInput -- specifically the box structure, which in your example contains the typesetting. I can't faithfully reconstruct that from just an expression. 

So if you want to make a raster of an *actual* input + output cell, that should probably be a separate function that targets a particular notebook and makes rasters of its inputs and outputs. I can do this but I'd suggest it be done in a separate PR.
</blockquote></details>

Ok, I've actually implemented this now, as the function `RasterizePreviousCell[]`. Here's how it works in a notebook:


/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-03 at 14.43.31@2x.png

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 5 files reviewed, 10 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-MLDBWQ8-CyIWsTs_4s2:bkxugqs) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*

Ok, I've actually implemented this now, as the function `RasterizePreviousCell[]`. Here's how it works in a notebook:


/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-03 at 14.43.31@2x.png
</blockquote></details>

I'm now correctly matching spacing:

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLE7CRCBdR0nnN7vBw9:bskk660) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*
> ```Markdown
> 
> ```wl
> ExportImageForEmbedding["AGoodNameGoesHere", RasterizePreviousCell[]]
> ```

This is awesome!
Having this trouble, though:
```wl
In[]:= ExportImageForEmbedding["AGoodNameGoesHere", RasterizePreviousCell[]]
```


The problem might be how `$SetReplaceBaseDirectory` is obtained:
```wl
In[]:= $SetReplaceBaseDirectory
Out[]= "/home/daniels/.Mathematica/Paclets/Repository/SetReplace-0.3.84"
```

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLECfzT8XBAuCjQqn5J:b-m2am4j) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*

This is awesome!
Having this trouble, though:
```wl
In[]:= ExportImageForEmbedding["AGoodNameGoesHere", RasterizePreviousCell[]]
```


The problem might be how `$SetReplaceBaseDirectory` is obtained:
```wl
In[]:= $SetReplaceBaseDirectory
Out[]= "/home/daniels/.Mathematica/Paclets/Repository/SetReplace-0.3.84"
```

</blockquote></details>

Can you try run load SetReplace from the repo? E.g. use `Get["~/git/SetReplace"]`? 

I realize though that if people try to run the ExportImageForEmbedding from an installed paclet, it won't know where to find the documentation, since that is a development-only thing. I should at least detect this case and give a sensible error message. Thanks!

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLEEbzh6Lher-Vp7QiD:be4im1j) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*

Can you try run load SetReplace from the repo? E.g. use `Get["~/git/SetReplace"]`? 

I realize though that if people try to run the ExportImageForEmbedding from an installed paclet, it won't know where to find the documentation, since that is a development-only thing. I should at least detect this case and give a sensible error message. Thanks!
</blockquote></details>

Yep, that works :+1: 


Perhaps supporting a `File` as the first argument would make it more flexible for these cases.

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 0 of 6 files reviewed, 12 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[Kernel/rasterization.m, line 168 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEFhGg5qMcFl3j4v3b:-MLEFhGg5qMcFl3j4v3c:bprnxiu) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L168)):*
> ```objc
> "
> 
> RasterizePreviousCell[] := Scope[
> ```

Missing `SyntaxInformation`.

maxitg (changes requested):
> Reviewed 2 of 5 files at r3, 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 3 of 6 files reviewed, 20 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[test.wls, line 28 at r3](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDO-JS2I8qRkeP_IOL:-MLDO-JT9tNGIqXGgAI8:b-jzxx4a) ([raw file](https://github.com/maxitg/setreplace/blob/155ac1198886ece87ef6bd3b9b3bba9a6dca5045/test.wls#L28)):*
> ```
>   CloseKernels[];
>   LaunchKernels[];
>   With[{path = $localSetReplaceEntryPoint}, ParallelEvaluate[Get[path]]]
> ```

This looks like a breaking change to the testing script. This cannot possibly be in the same PR as the rasterization functions, because it has absolutely nothing to do with them. It needs to be discussed separately.

It shouldn't be necessary to change `test.wls` in this PR at all, and there are in fact three(!) extra PRs here all mixed up into one: testing without installing (which should still require building as some functions won't work otherwise), `requiresFrontEnd` (which should add itself to existing graphics-testing tests as well), and `-np` (I believe the reason tests take so long to start is because the kernel tries to use the Internet connection on parallel kernels launch, it would be better if can fix that instead of disabling parallelization outright).
___
*[.github/CONTRIBUTING.md, line 468 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDoYzh2WDsK6bODJN7:-MLDoYzh2WDsK6bODJN8:bevx2pg) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L468)):*
> ```Markdown
> We are using GitHub-flavored Markdown for documentation and research notes.
> 
> Images of input and/or output cells can be easily produced using Mathematica by running the command 
> ```

Remove trailing spaces.
___
*[.github/CONTRIBUTING.md, line 468 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDrNo6AuP_nBc4_uHQ:-MLDrNo6AuP_nBc4_uHR:b-fujcf3) ([raw file](https://github.com/maxitg/setreplace/blob/64ef6e8ee30bf4dd54ca82bc1750e4235f681ed3/.github/CONTRIBUTING.md#L468)):*
> ```Markdown
> We are using GitHub-flavored Markdown for documentation and research notes.
> 
> Images (e.g., of output cells) should be made by selecting the relevant cells in the Front End, copying them as bitmaps, and saving them as .png files to [Documentation/Images](/Documentation/Images) (in the documentation) or to the Images directory of the corresponding research note. They should then be inserted using the code similar to this:
> ```

"or to the Images directory of the corresponding research note" is important and should not be lost.
___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLEDQLh9kAFOMpZbY2v:b-ai8tc) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*

Can you try run load SetReplace from the repo? E.g. use `Get["~/git/SetReplace"]`? 

I realize though that if people try to run the ExportImageForEmbedding from an installed paclet, it won't know where to find the documentation, since that is a development-only thing. I should at least detect this case and give a sensible error message. Thanks!
</blockquote></details>

This is a very special use case. I wouldn't object if this function could detect it's loaded this way and handle it correctly, but the full image path should still be supported, and I don't think we should recommend people to load WL code directly because it's going to cause all kind of issues with missing functions and inconsistent C++/WL code. We should work on optimizing the build script instead.

Also, note that not all images go to `/Documentation/Images`, some of them will go to specific research notes. So, I think this should take at least a *relative* path as an argument.
___
*[.github/CONTRIBUTING.md, line 479 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDqmA04RiSc3lC2hTh:-MLDqmA04RiSc3lC2hTi:b-xo64or) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L479)):*
> ```Markdown
> 
> This function will return markdown-compatible code that you can paste straight into the relevant readme or research
> note. In fact, the function will copy this code straight into your clipboard so that you can simply paste it into
> ```

If it copies something to the clipboard, it should make it obvious in the name because it is a potentially destructive operation.
___
*[Kernel/init.m, line 8 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDr_MS0Vb4jOf6pLRt:-MLDr_MT0p46guqVv3KD:bn455uf) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/init.m#L8)):*
> ```objc
> 
> Block[
>   {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
> ```

Remove the trailing tab.
___
*[Kernel/init.m, line 9 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDx3o5-GlmcYQytaL9:-MLDx3o5-GlmcYQytaLA:b35yp4g) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/init.m#L9)):*
> ```objc
> Block[
>   {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
>       HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
> ```

What is `$Unreachable`. I don't see it used anywhere.
___
*[Kernel/init.m, line 10 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDs9l2CzDiTeZB8gmq:-MLDs9l2CzDiTeZB8gmr:bhvqj69) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/init.m#L10)):*
> ```objc
>   {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {	
>       HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
> 	    HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
> ```

Tabs -> spaces (I suggest enabling invisible characters in the editor until we have a code formatter because there are a lot of these, it's also usually possible to set up the editor to emit spaces instead of tabs).
___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-MLED1gL5C5JaDd9U7jk:bbdt1rg) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*

I'm now correctly matching spacing:



</blockquote></details>

This seems to generate the wrong vertical position for the input cell:


Also, why wouldn't `RasterizeAsInput` do that?

I don't see why would we ever want to generate the image without custom boxes, because if it's just text, it should stay as text in Markdown.
___
*[Kernel/rasterization.m, line 4 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE6pQG-iphA_JU_C-V:-MLE6pQG-iphA_JU_C-W:b3twcae) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L4)):*
> ```objc
> 
> PackageImport["GeneralUtilities`"]
> ```

All of these functions need complete argument checking if they are exposed.
___
*[Kernel/rasterization.m, line 6 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE6RbLCCsAct3Q24R-:-MLE6RbM-mW5voy2Esng:b-cgmt66) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L6)):*
> ```objc
> 
> PackageExport["MakeFormattedCodeBoxes"]
> PackageExport["RasterizeCell"]
> ```

What is the use case here for what we cannot already do with `RasterizeExpressionAs*`?
___
*[Kernel/rasterization.m, line 7 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE16wn7Cir6IeBLHZX:-MLE16wn7Cir6IeBLHZY:b-o5mcep) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L7)):*
> ```objc
> PackageExport["MakeFormattedCodeBoxes"]
> PackageExport["RasterizeCell"]
> PackageExport["RasterizeExpression"]
> ```

Is it reimplementing `Rasterize`? If so, we should not expose it without a good reason. Given that the purpose of these is to generate single images for the documentation, `Rasterize` is sufficiently fast for that. If this does something else that `Rasterize` does not do, it should mention it in the name.
___
*[Kernel/rasterization.m, line 15 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDxVgc5ZYxc83GHoNO:-MLDxVgc5ZYxc83GHoNP:b-okkgq5) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L15)):*
> ```objc
> 
> SetUsage @ "
> MakeFormattedCodeBoxes[expr$] generates boxes representing expr$ as 'nicely formatted code'.
> ```

Why would we want to have a boxes expression?

```wl
In[] := MakeFormattedCodeBoxes[Graph[{1 -> 2}]]
Out[] = RowBox[{"Graph", "[", 
  RowBox[{"{", RowBox[{"1", " ", "\[Rule]", " ", "2"}], "}"}], "]"}]
```

Also, if it's a part of the code formatter, we will need to figure out the style first (because `PrettyForm` is not consistent with the current SetReplace code style) before we can expose it.
___
*[Kernel/rasterization.m, line 25 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDzQ0FCROwVwJpyRpQ:-MLDzQ0FCROwVwJpyRpR:b-dpvow5) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L25)):*
> ```objc
> SetAttributes[MakeFormattedCodeBoxes, HoldFirst];
> 
> MakeFormattedCodeBoxes[expr_, n_:60] := Block[
> ```

Why 60? The limit for ordinary code is 120. For md files, it should be a bit smaller because the page width does not fit 120 characters, we should probably measure what it is and use it here.
___
*[Kernel/rasterization.m, line 25 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE77t4ClnQ-qibJacF:-MLE77t4ClnQ-qibJacG:b-ljtgj7) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L25)):*
> ```objc
> SetAttributes[MakeFormattedCodeBoxes, HoldFirst];
> 
> MakeFormattedCodeBoxes[expr_, n_:60] := Block[
> ```

I don't think

```wl
In[] := MakeFormattedCodeBoxes[Graph[{1 -> 2}], -\[Pi]]
```

will do the right thing here.
___
*[Kernel/rasterization.m, line 162 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDnhGo3ze55wb8uAc1:-MLDnhGo3ze55wb8uAc2:b-monja4) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L162)):*
> ```objc
> RasterizePreviousCell[] will read the previous cell(s) from the current notebook \
> and attempt to rasterize them.
> * An input cell followed by an output cell will be rasterized together, similar to RasterizeExpressionAsInputOutputPair.
> ```

We will typically only want to rasterize the output cell, as the input cell would be pure text (it should be copyable). Also, the name says `Cell` (singular).
___
*[Kernel/rasterization.m, line 206 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEHjWs2456Mh8uyItF:-MLEHjWs2456Mh8uyItG:b6u2wr8) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L206)):*
> ```objc
>   If[!StringQ[result], ReturnFailed["noexport"]];
>   width = First @ ImageDimensions[image];
>   markdown = $ImageMarkdownTemplate[filename, width / 2.];
> ```

I agree that the value of 1/2 makes more sense here than 0.6. However, consistency is important. In order to change this, we will need to change all of the existing pictures as well (in a separate PR). I suggest keeping it at 0.6 for now (unless you want to merge that other PR first).
___
*[Kernel/usageString.m, line 18 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEIELtEWm8morPZv8m:-MLEIELtEWm8morPZv8n:b-toktft) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/usageString.m#L18)):*
> ```objc
> 
> PackageExport["$SetReplaceBaseDirectory"]
> (* this is actually set in Kernel/init.m, but needs to be exported from somewhere *)
> ```

Put the comment before the line it's commenting about.

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 3 of 6 files reviewed, 20 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[test.wls, line 28 at r3](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDO-JS2I8qRkeP_IOL:-MLElpw0CKB-J492fHjS:bq4nxm3) ([raw file](https://github.com/maxitg/setreplace/blob/155ac1198886ece87ef6bd3b9b3bba9a6dca5045/test.wls#L28)):*

This looks like a breaking change to the testing script. This cannot possibly be in the same PR as the rasterization functions, because it has absolutely nothing to do with them. It needs to be discussed separately.

It shouldn't be necessary to change `test.wls` in this PR at all, and there are in fact three(!) extra PRs here all mixed up into one: testing without installing (which should still require building as some functions won't work otherwise), `requiresFrontEnd` (which should add itself to existing graphics-testing tests as well), and `-np` (I believe the reason tests take so long to start is because the kernel tries to use the Internet connection on parallel kernels launch, it would be better if can fix that instead of disabling parallelization outright).
</blockquote></details>

I'll make another PR for the testing script changes, per our discussion.
___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLEc0aKDfjB4hE0TTBb:b3g9ocm) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*

This is a very special use case. I wouldn't object if this function could detect it's loaded this way and handle it correctly, but the full image path should still be supported, and I don't think we should recommend people to load WL code directly because it's going to cause all kind of issues with missing functions and inconsistent C++/WL code. We should work on optimizing the build script instead.

Also, note that not all images go to `/Documentation/Images`, some of them will go to specific research notes. So, I think this should take at least a *relative* path as an argument.
</blockquote></details>

About the location of the images: that is a good point. One solution might be to support two shortcut syntaxes and then a fully qualified path:

"Foo.png" → "~/git/SetReplace/Documentation/Images/Foo.png"
"Research/SubFolder/Images/Foo.png" → "~/git/SetReplace/Research/SubFolder/Images/Foo.png"
"~/git/SetReplace/Documentation/Images/Foo.png" → as given. 

You would only be able to use the first two if you loaded the paclet from your git repo -- you'd get an error message otherwise.
___
*[.github/CONTRIBUTING.md, line 479 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDqmA04RiSc3lC2hTh:-MLEbWim62_nyKrow4sk:bpwyf1s) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L479)):*

If it copies something to the clipboard, it should make it obvious in the name because it is a potentially destructive operation.
</blockquote></details>

I'll remove the feature, since putting it in the name would make it a very ugly named function.
___
*[Kernel/init.m, line 9 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDx3o5-GlmcYQytaL9:-MLEVetH9Bey7koTyc1u:b-g53dhr) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/init.m#L9)):*

What is `$Unreachable`. I don't see it used anywhere.
</blockquote></details>

It's a GeneralUtilities symbol that indicates some code is unreachable. It causes a Panic if it is ever reached that helps you debug stuff.
___
*[Kernel/init.m, line 10 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDs9l2CzDiTeZB8gmq:-MLEVmimEksRVY3STkQ4:b-j9jdyo) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/init.m#L10)):*

Tabs -> spaces (I suggest enabling invisible characters in the editor until we have a code formatter because there are a lot of these, it's also usually possible to set up the editor to emit spaces instead of tabs).
</blockquote></details>

Thought I caught all of these but apparently not.
___
*[Kernel/rasterization.m, line 55 at r2](https://reviewable.io/reviews/maxitg/setreplace/506#-ML9l41z6XF_bDaaYYnN:-MLEXYY-1jG7wQZWIc9Y:b-i37h56) ([raw file](https://github.com/maxitg/setreplace/blob/62b58983247da8a45846642b21296d72903c4cf4/Kernel/rasterization.m#L55)):*

This seems to generate the wrong vertical position for the input cell:


Also, why wouldn't `RasterizeAsInput` do that?

I don't see why would we ever want to generate the image without custom boxes, because if it's just text, it should stay as text in Markdown.
</blockquote></details>

To solve this properly I'd have to measure the baseline of the input image and match it up. It's possible, but complicated and I've spent too much time on this PR already. Can we live with it for now? 
___
*[Kernel/rasterization.m, line 6 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE6RbLCCsAct3Q24R-:-MLEYzuk-xw-HMac_fr7:brm6x0w) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L6)):*

What is the use case here for what we cannot already do with `RasterizeExpressionAs*`?
</blockquote></details>

Fine I'll unexpose. 
___
*[Kernel/rasterization.m, line 7 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE16wn7Cir6IeBLHZX:-MLEVtNd5RHlvUOpg36N:b-1b4maw) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L7)):*

Is it reimplementing `Rasterize`? If so, we should not expose it without a good reason. Given that the purpose of these is to generate single images for the documentation, `Rasterize` is sufficiently fast for that. If this does something else that `Rasterize` does not do, it should mention it in the name.
</blockquote></details>

Ah, you're right. I'll avoid exposing it. 
___
*[Kernel/rasterization.m, line 15 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDxVgc5ZYxc83GHoNO:-MLEYqnb7V4ZJZgHjpV1:b-jcd89v) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L15)):*

Why would we want to have a boxes expression?

```wl
In[] := MakeFormattedCodeBoxes[Graph[{1 -> 2}]]
Out[] = RowBox[{"Graph", "[", 
  RowBox[{"{", RowBox[{"1", " ", "\[Rule]", " ", "2"}], "}"}], "]"}]
```

Also, if it's a part of the code formatter, we will need to figure out the style first (because `PrettyForm` is not consistent with the current SetReplace code style) before we can expose it.
</blockquote></details>

I'll unexpose it.
___
*[Kernel/rasterization.m, line 25 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDzQ0FCROwVwJpyRpQ:-MLEWskA3uzs4UdFslFl:b5hteoe) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L25)):*

Why 60? The limit for ordinary code is 120. For md files, it should be a bit smaller because the page width does not fit 120 characters, we should probably measure what it is and use it here.
</blockquote></details>

The reasoning is that it is expensive to have very wide inputs, since they will cause a large amount of extra white pixels to be generated (since the *output cell* raster will be padded to the same width as the input so they can be joined into one image). That just seems like unnecessary bloating of the git repo. If you were writing the input cell *by hand* in a notebook you would definitely line break it sooner than 120 characters for this precise reason (well, at least, I would).
___
*[Kernel/rasterization.m, line 25 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE77t4ClnQ-qibJacF:-MLEXCsi7irRlfsV0Ov2:b-hq4cp5) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L25)):*

I don't think

```wl
In[] := MakeFormattedCodeBoxes[Graph[{1 -> 2}], -\[Pi]]
```

will do the right thing here.
</blockquote></details>

It sounds like you don't want this function anyway, so I'll just unexpose it.
___
*[Kernel/rasterization.m, line 162 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDnhGo3ze55wb8uAc1:-MLEXj0k3DYJfo0wneyf:b-gdhymd) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L162)):*

We will typically only want to rasterize the output cell, as the input cell would be pure text (it should be copyable). Also, the name says `Cell` (singular).
</blockquote></details>

Want to suggest a better name? RasterizePreviousResult?
___
*[Kernel/rasterization.m, line 168 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEFhGg5qMcFl3j4v3b:-MLEam6k63i_fC92wZ83:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L168)):*

Missing `SyntaxInformation`.
</blockquote></details>

Done.
___
*[Kernel/rasterization.m, line 206 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEHjWs2456Mh8uyItF:-MLEXlzF69PfsOYBZ8MX:bw793gs) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L206)):*

I agree that the value of 1/2 makes more sense here than 0.6. However, consistency is important. In order to change this, we will need to change all of the existing pictures as well (in a separate PR). I suggest keeping it at 0.6 for now (unless you want to merge that other PR first).
</blockquote></details>

The value of 1/2 makes sense because that is exactly the right value to produce retina-exact rendering -- values that show up as identical to the retina output on retina displays, and ordinary on ordinary displays. It feels wrong to me to break that for slight scaling consistency that no-one will notice -- in other words, why make new images look slightly worse so they all look *consistently* slightly worse? 
___
*[Kernel/usageString.m, line 18 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEIELtEWm8morPZv8m:-MLEa5NWEKCJSjIEJyoX:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/usageString.m#L18)):*

Put the comment before the line it's commenting about.
</blockquote></details>

Done.

maxitg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/506)* status: 3 of 6 files reviewed, 20 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[.github/CONTRIBUTING.md, line 475 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLE7CRB7jMlPo5B-H-_:-MLJoLqeBkE-51BVZ0QY:b-kqbqnv) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/.github/CONTRIBUTING.md#L475)):*

About the location of the images: that is a good point. One solution might be to support two shortcut syntaxes and then a fully qualified path:

"Foo.png" → "~/git/SetReplace/Documentation/Images/Foo.png"
"Research/SubFolder/Images/Foo.png" → "~/git/SetReplace/Research/SubFolder/Images/Foo.png"
"~/git/SetReplace/Documentation/Images/Foo.png" → as given. 

You would only be able to use the first two if you loaded the paclet from your git repo -- you'd get an error message otherwise.
</blockquote></details>

I think the shorthand syntax might be useful even if the paclet is built with a build script. This is because if it's built this way (instead of the multi-platform way for releases), it is most likely built by a developer which means the repo directory probably did not change as well. I suggest we just expose a new global constant `$SetReplaceRepositoryRoot` (similar to `$SetReplaceGitSHA`, etc.), which would be defined either by loading the unbuilt source or by using the build script.

I also suggest removing both the constant above and the `ExportImageForEmbedding` function altogether from the release builds.
___
*[Kernel/rasterization.m, line 162 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLDnhGo3ze55wb8uAc1:-MLJn6ET8nQNC6BOODpi:b9sdpi7) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L162)):*

Want to suggest a better name? RasterizePreviousResult?
</blockquote></details>

No, I suggest it to only rasterize a single previous cell.

The only reason we would ever want to rasterize the input cell is if the input cell itself contains non-textual boxes, which is quite rare. Almost always, the input would be a text code block, and the output would be an image. So, this function should only rasterize the output cell (unless the previous cell is something else).
___
*[Kernel/rasterization.m, line 206 at r4](https://reviewable.io/reviews/maxitg/setreplace/506#-MLEHjWs2456Mh8uyItF:-MLJngx14zO9C_laspMX:bo4uq8r) ([raw file](https://github.com/maxitg/setreplace/blob/43789b9790c36f1bb1c580baf43d108ccd659ce5/Kernel/rasterization.m#L206)):*

The value of 1/2 makes sense because that is exactly the right value to produce retina-exact rendering -- values that show up as identical to the retina output on retina displays, and ordinary on ordinary displays. It feels wrong to me to break that for slight scaling consistency that no-one will notice -- in other words, why make new images look slightly worse so they all look *consistently* slightly worse? 
</blockquote></details>

Because consistency is more important. Someone might need to use this code to regenerate existing images, in which case they would look different from surrounding images and create inconsistencies and confusion.

This is a somewhat moot point because we need to change existing images anyway. I'm just suggesting to change this value to 0.5 in that other PR instead of this one.

PR #518, opened 2020-11-06
Improve libSetReplace loading

taliesinb:
Try to find the libSetReplace in the same part of the filesystem that the source was loaded from. If that fails, look for a possible Build directory that contains a staged build to use. If that fails, give up.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/518)

PR #519, opened 2020-11-06
Use a temporary build directory

taliesinb:
Use a temporary build directory, rather than the /Build subdir of the root dir. This is friendlier to IDEs, find-in-files, etc. This will have to be rebased if #518 lands to update the temporary directory used *there* as well.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/519)

### DIFF
--- a/build.wls
+++ b/build.wls
@@ -9,6 +9,7 @@ Check[
 $success = True;
 
 Check[
+  Print["Temporary build directory is ", $buildDirectory];
   $builtLibSetReplaceHash = builtLibSetReplaceHash[];
   $libSetReplaceSourceHash = libSetReplaceSourceHash[];
   If[$builtLibSetReplaceHash =!= $libSetReplaceSourceHash,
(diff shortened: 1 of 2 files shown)


maxitg (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/519#-:-ML_ttPN1d4ZVo97Lo6l:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

taliesinb:
@maxitg : I have actually implemented the more ambitious goal we discussed, about doing inline builds on developer machines so that there is no separate Build directory. I'll open a PR for that, but let's keep this PR around in case we reject that PR.

maxitg:
Why don't we merge it now while we work on that other PR? It does not make things worse, and it's done already. I don't see any reason to hold it.

taliesinb (commented):
Ok!

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/519#-:-MLgaZKY90GUJ9vbzLyI:b2nhlkw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @taliesinb)

PR #520, opened 2020-11-09
Load GeneralUtilities in Kernel/init.m, and add back some simple macro sugar

### LINKED ISSUE #508: Build fails if SetReplace is imported in kernel init without GeneralUtilities
### The problem
Set `Kernel/init.m` to
```wl
<< SetReplace`;
```

Run the build script. The build fails:
```
> ./build.wls

ReplaceRepeated::reps: {GeneralUtilities`Control`PackagePrivate`$DesugaringRules} is neither a list of replacement rules nor a valid dispatch table, and so cannot be used for replacing.

ReplaceRepeated::reps: {GeneralUtilities`Control`PackagePrivate`$DesugaringRules} is neither a list of replacement rules nor a valid dispatch table, and so cannot be used for replacing.

ReplaceRepeated::reps: {GeneralUtilities`Control`PackagePrivate`$DesugaringRules} is neither a list of replacement rules nor a valid dispatch table, and so cannot be used for replacing.

General::stop: Further output of ReplaceRepeated::reps will be suppressed during this calculation.
```

### Expected behavior
The build should not fail.

### Version
```wl
```
image

```wl
In[] := $SetReplaceGitSHA
```
64ef6e8ee30bf4dd54ca82bc1750e4235f681ed3

### Additional context
It appears that kernel init is evaluated before *GeneralUtilities* is loaded. We will need to import it manually in *SetReplace*'s `init.m` without adding *GeneralUtilities* to `$ContextPath`.

taliesinb:
## Changes

* Fixes #508.
* Use `BeginPackage` and `Begin` to do proper context isolation to make loading GU safer, which we now do.
* Also include some useful sugar that isn't funky syntax which the previous override of `$DesugaringRules` unintentionally removed -- I will use this in a lot of code I write, including a pending PR (that will be resurrected soon!).
* Define the package global `$SetReplaceBaseDirectory`, which other PRs will end up using in various places.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520)

### DIFF
--- a/Kernel/globals.m
+++ b/Kernel/globals.m
@@ -1,0 +1,12 @@
+Package["SetReplace`"]
+
+PackageImport["GeneralUtilities`"]
+
+(* Note: this is actually set in Kernel/init.m, but it needs to be exported from a new-style package so that
+the new-style package loader understands that this is a global symbol *)
+PackageExport["$SetReplaceRootDirectory"]
+
+SetUsage @ "
+$SetReplaceRootDirectory contains the path of the root of the SetReplace package that was loaded. \
+This corresponds to the directory that containts the Kernel/ directory.
+"
--- a/Kernel/init.m
+++ b/Kernel/init.m
@@ -2,9 +2,31 @@ Unprotect["SetReplace`*"];
 
 ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
 
+(* Make sure that we don't effect $ContextPath by getting GU, and that we are isolated from
+any user contexts *)
+BeginPackage["SetReplace`"];
+
+(* This is useful for various purposes, like loading libraries from the 'same place' as the
+paclet, and also knowing *where* the loaded code even came from. *)
+$SetReplaceRootDirectory = FileNameDrop[$InputFileName, -2];
+
+Needs["GeneralUtilities`"];
+
+(* ensure private symbols we use below don't show up on Global, etc *)
+Begin["SetReplace`Private`"]
+
 Block[
-  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {}},
-  Get[FileNameJoin[{FileNameDrop[$InputFileName], "usageString.m"}]];
+  (* Temporarily overrule some of the more exotic features of the macro system.
+  TODO: Fix this upstream when GU is open sourced. *)
+  {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {
+    HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
+    HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
+    HoldPattern[ReturnFailure[msg_String, args___]] :> ReturnFailure[MessageName[$LHSHead, msg], args]
+  }},
+  Get[FileNameJoin[{$SetReplaceRootDirectory, "Kernel", "usageString.m"}]];
 ];
+
+End[];
+EndPackage[];
 
 SetAttributes[#, {Protected, ReadProtected}] & /@ Evaluate @ Names @ "SetReplace`*";
--- a/Kernel/init.m
+++ b/Kernel/init.m
@@ -2,7 +2,7 @@ Unprotect["SetReplace`*"];
 
 ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
 
-(* Make sure that we don't effect $ContextPath by getting GU, and that we are isolated from
+(* Make sure that we don't affect $ContextPath by getting GU, and that we are isolated from
 any user contexts *)
 BeginPackage["SetReplace`"];
 
--- a/Tests/buildData.wlt
+++ b/Tests/buildData.wlt
@@ -1,4 +1,12 @@
 <|
+  "$SetReplaceRootDirectory" -> <|
+    "tests" -> {
+      VerificationTest[
+        FileExistsQ @ $SetReplaceRootDirectory
+      ]
+    }
+  |>,
+
   "$SetReplaceGitSHA" -> <|
     "tests" -> {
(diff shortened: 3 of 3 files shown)


maxitg:
The CI fails because the usage and tests for `$SetReplaceBaseDirectory` are missing.

taliesinb:
@maxitg This is ready for review.

daneelsan (commented):
Looks good to me.
So is it okay (going forward) to use `ReturnFail*` and `$Unreachable` in the codebase?

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520#-:-MMq1EUjEJfKSG8fFhTP:bismbs2)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @daneelsan and @maxitg)

daneelsan (commented):
Also, is this correct? Maybe I'm confused but I thought this variable was supposed to point to the git directory.
```
In[] := $SetReplaceRootDirectory
Out[] = "/home/daniels/.Mathematica/Paclets/Repository/SetReplace-0.3.93"
```

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520#-:-MMq1qyzAOGkLBJ-X_Sl:bf7c3k3)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @daneelsan and @maxitg)

maxitg (changes requested):
@daneelsan, it points to the directory the code was loaded from. It will point to the git directory if you `Get` *SetReplace* directly from it, which will be fully supported once all pending PRs are merged.

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520#-:-MMq_6AW1A9sEsFfNQFj:bfyvdak)* status: all files reviewed, 2 unresolved discussions (waiting on @taliesinb)

___
*[Kernel/init.m, line 5 at r1](https://reviewable.io/reviews/maxitg/setreplace/520#-MMq_QvYD8ScfooYYaxj:-MMq_QvYD8ScfooYYaxk:bnudyvo) ([raw file](https://github.com/maxitg/setreplace/blob/ae36ca191aa42c85ddf35b9052185844d4c881d5/Kernel/init.m#L5)):*
> ```objc
> ClearAll @@ (# <> "*" & /@ Contexts["SetReplace`*"]);
> 
> (* Make sure that we don't effect $ContextPath by getting GU, and that we are isolated from
> ```

effect -> affect.
___
*[Kernel/init.m, line 23 at r1](https://reviewable.io/reviews/maxitg/setreplace/520#-MMqaFTdDOCnacntMYrj:-MMqaFTdDOCnacntMYrk:b-gexy59) ([raw file](https://github.com/maxitg/setreplace/blob/ae36ca191aa42c85ddf35b9052185844d4c881d5/Kernel/init.m#L23)):*
> ```objc
>   {GeneralUtilities`Control`PackagePrivate`$DesugaringRules = {
>     HoldPattern[$Unreachable] :> Unreachable[$LHSHead],
>     HoldPattern[ReturnFailed[msg_String, args___]] :> ReturnFailed[MessageName[$LHSHead, msg], args],
> ```

It seems like this will never be used because:
1. If we are in a public function, we will want to return unevaluated instead of `$Failed`.
2. If we are not in a public function, we will want to use the public function's head for the message.

Is there something I'm missing here?

---

On a side note, It would be cool to have a function that would:
1. Exit all functions until it finds one called from the outside of the package.
2. Return that one unevaluated.
3. Print a message using the same head (of the unevaluated function).

Right now, we are doing this by:
1. Defining the public function as:

  ```wl
  expr : Func[args___] := With[{res = Catch[func[Func, args]]},
    res /; res =!= $Failed
  ]
  ```

2. Passing the head or `expr` into the internal function (because it could be a package scope function called from different places).
3. Using `Throw` to exit all the way back to the original function `Func`.
4. If we need to use one public function from another, we use the internal counterpart instead.

It would be nice to simplify this if possible (obviously, not in this PR).

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520)* status: all files reviewed, 2 unresolved discussions (waiting on @maxitg)

___
*[Kernel/init.m, line 5 at r1](https://reviewable.io/reviews/maxitg/setreplace/520#-MMq_QvYD8ScfooYYaxj:-MMqo8m45YhPiXcT1vWr:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/ae36ca191aa42c85ddf35b9052185844d4c881d5/Kernel/init.m#L5)):*

effect -> affect.
</blockquote></details>

Done.
___
*[Kernel/init.m, line 23 at r1](https://reviewable.io/reviews/maxitg/setreplace/520#-MMqaFTdDOCnacntMYrj:-MMqo9i1E8uOV3ppgBH7:basemw4) ([raw file](https://github.com/maxitg/setreplace/blob/ae36ca191aa42c85ddf35b9052185844d4c881d5/Kernel/init.m#L23)):*
> If we are in a public function, we will want to return unevaluated instead of $Failed.

This isn't necessarily true: some System functions *do* return Failed, rather than stay unevaluated. Specifically this occurs when there is an effectful function that cannot produce the desired effect, e.g. `CreateDirectory[9]`. That's what I intend to use ReturnFailed for. I use this idiom extensively in DevUtils, which in some cases is loaded *within* SetReplace, and so needs this sugar to exist to behave in the proper way.

> On a side note, It would be cool to have a function that would:

This is what I created `ThrowFailure` for! It allows one to throw a failure with an error message from some internal "driver code", and have it caught by some "top level" code which will produce the desired message on the symbol associated with that top-level code. We can discuss further on Discord.

maxitg (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/520#-:-MMs-u3P5xn-AsJrs6u-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #531, opened 2020-11-23
Rasterization in DevUtils`

taliesinb:
## Changes
* This change depends on PR #529, since the code lives in `` DevUtils` ``
* This replaces #506, which is considerably more complicated
* It introduces a single function, `ExportPreviousInputOutputToMarkdown`, which rasterizes an input+output pair, and produces markdown + image as a result. If possible the input is pure Mathematica, while the output is an image.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/531)


taliesinb:
This is roughly how it works:

PixelSnap 2020-11-23 at 22 37 25@2x

taliesinb:
I have now documented the workflow in Contributing.

PR #535, opened 2020-11-26
Introduce rasterization functions to DevUtils

taliesinb:
These are:

* RasterizePreviousInputOutputAndExportToMarkdown
* RasterizePreviousOutputAndExportToMarkdown
* RasterizeExpressionAndExportToMarkdown

They have a variety of options (including the max width of image after
which to line break, the magnification, and details about compression).

The rasterization workflow is documented in CONTRIBUTING.md.

Here's an example of `RasterizePreviousInputOutputAndExportToMarkdown` being used (with the DryRun option given to make the example easier to see):

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.56.50@2x.png

This shows the behavior when the input cannot be stringified: 

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.56.12@2x.png

This shows the behavior of `RasterizePreviousOutputAndExportToMarkdown`:

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.57.23@2x.png

This shows the behavior of `RasterizeExpressionAndExportToMarkdown`:

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.58.11@2x.png

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)


daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[.github/CONTRIBUTING.md, line 545 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4O9lK2Oj3SoC6MPFj:-MN4O9lK2Oj3SoC6MPFk:b-fwm0z6) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/.github/CONTRIBUTING.md#L545)):*
> ```Markdown
> 
> ```wl
> Get["~/git/SetReplace/DevUtils/Kernel.init"];
> ```

This should be `"~/git/SetReplace/DevUtils/init.m"`.
___
*[DevUtils/Rasterization.m, line 12 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4PLoi8BG6DzGui28C:-MN4PLoi8BG6DzGui28D:bwaochb) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L12)):*
> ```objc
> 
> $usageSuffix = "
> * 'path$' should be a path, releative to the repository root, with a CamelCased filename ending in '.png'.
> ```

releative -> relative
___
*[DevUtils/Rasterization.m, line 16 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4PecAEuGGg_AIWbdl:-MN4PecAEuGGg_AIWbdm:bob09bx) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L16)):*
> ```objc
> * The resulting markdown will contain an absolute path relative to the root of the repository. It will \
> correctly render on e.g. GitHub but may not render in e.g. VSCode.
> * The option 'MaxWidth' controls the desird maximum width of the resulting raster. If possible, the expression \
> ```

desird -> desired
___
*[DevUtils/Rasterization.m, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4QdqM6jbjLo9GOoeR:-MN4QdqM6jbjLo9GOoeS:bomhl43) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L47)):*
> ```objc
> 
> SetUsage @ Evaluate["
> RasterizePreviousOutputAndExportToMarkdown['path$', expr$] ill read the previous output cell from the current \
> ```

ill read -> will read
___
*[DevUtils/Rasterization.m, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eLArEuDNV_FTuUuq:-MN4eLArEuDNV_FTuUur:b-9h29vv) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L47)):*
> ```
> RasterizePreviousOutputAndExportToMarkdown['path$', expr$]
> ```

The implementation only accepts a path, not an expression.
___
*[DevUtils/Rasterization.m, line 50 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4Ql11D5EpUvXJNEg4:-MN4Ql11D5EpUvXJNEg5:b-nfv46u) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L50)):*
> ```objc
> notebook, rasterize it, write the result to 'path$', and return an HTML  tag that can be pasted \
> directly into a markdown file.
> HTML  tag that can be pasted directly into a markdown file.
> ```

I think this line was duplicated.
___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MN4eUEwAFM0WkKoEuVt:b-uykfzc) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*
> ```objc
> directly into a markdown file.
> HTML  tag that can be pasted directly into a markdown file.
> * The resulting image WILL include an attached Out[]= label." <> $usageSuffix];
> ```

I'm not able to see the Out[]= label, perhaps this is why `RasterizePreviousOutputAndExportToMarkdown` is also not outputing "Out"?
___
*[DevUtils/Rasterization.m, line 69 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4R9Aw-XWC0w2QEL0s:-MN4R9Aw-XWC0w2QEL0t:b-ml1mna) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L69)):*
> ```objc
> SetUsage @ Evaluate["
> RasterizePreviousInputOutputAndExportToMarkdown['path$'] will read the previous input and output cell from the \
> current notebook, rasterize the output, write the result to 'path$', and return an markdown code block containing \
> ```

an markdown -> a markdown
___
*[DevUtils/Rasterization.m, line 70 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4bSXG51NMOtyCXhg6:-MN4bSXG51NMOtyCXhg7:bbveros) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L70)):*
> ```objc
> RasterizePreviousInputOutputAndExportToMarkdown['path$'] will read the previous input and output cell from the \
> current notebook, rasterize the output, write the result to 'path$', and return an markdown code block containing \
> the input and an HTML  tag containing the output rasterize, that can be pasted directly into a markdown file.
> ```

perhaps "the rasterized output"?
___
*[DevUtils/Rasterization.m, line 73 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4dktQ2yb8v_qOeo4g:-MN4dktR-WIIORaeKIyU:b-n853h3) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L73)):*
> ```objc
> * If the input cell does not contain purely textual boxes, it cannot be faithfully represented as text, and so \
> it will be included in the rasterized image instead.
> * The option 'RasterizeInput' -> True will force the input to be rasterized, and will not create a markdown \
> ```

Since this is the main function, `RasterizePreviousInputOutputAndExportToMarkdown` should also 
___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MN4cE1D0fF2qAMhohq5:bb7f718) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*
> ```objc
>   If[TrueQ[rasterizeInput],
>     icellString = $Failed;
>   ,
> ```

Are the spaces for this comma deliberate? I would assume the spacing should be the same as the if/else branch. E.g:
```
If[True,
  a
  ,
  b
]
```
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MN4cjy27IUO3QoYkZVJ:bcrk0y8) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*
> ```objc
>   ];
>   If[StringQ[icellString],
>     mdPrefix = "```wl\n" <> icellString <> "\n```\n\n";
> ```

I notice there is no `In[] := `. Maybe I missed the discussion but I thought this was mandatory?

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[.github/CONTRIBUTING.md, line 545 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4O9lK2Oj3SoC6MPFj:-MN5UvTIAZprqhZAotJO:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/.github/CONTRIBUTING.md#L545)):*

This should be `"~/git/SetReplace/DevUtils/init.m"`.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 12 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4PLoi8BG6DzGui28C:-MN5V-Uy27PDpkGROAj2:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L12)):*

releative -> relative
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 16 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4PecAEuGGg_AIWbdl:-MN5V1cUE3gA5mjU6ysV:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L16)):*

desird -> desired
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4QdqM6jbjLo9GOoeR:-MN5V4Ig-19ytBvwWlld:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L47)):*

ill read -> will read
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 47 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eLArEuDNV_FTuUuq:-MN5V8eA8AWNUYoAUF_V:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L47)):*

> ```
> RasterizePreviousOutputAndExportToMarkdown['path$', expr$]
> ```

The implementation only accepts a path, not an expression.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 50 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4Ql11D5EpUvXJNEg4:-MN5VBXi5FIE6-Gd6co8:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L50)):*

I think this line was duplicated.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MN5VZk31HXlXe3hwrji:b6rjdxa) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

I'm not able to see the Out[]= label, perhaps this is why `RasterizePreviousOutputAndExportToMarkdown` is also not outputing "Out"?
</blockquote></details>

Just to be clear, you are NOT getting the following behavior? If so, which platform are you on?


/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-26 at 23.43.17@2x.png
___
*[DevUtils/Rasterization.m, line 69 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4R9Aw-XWC0w2QEL0s:-MN5Vv66BfD_Fuv8Fy_I:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L69)):*

an markdown -> a markdown
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 70 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4bSXG51NMOtyCXhg6:-MN5W-3BBd0At8uSEr9k:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L70)):*

perhaps "the rasterized output"?
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 73 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4dktQ2yb8v_qOeo4g:-MN5Vziw4rXtLVLzMWpV:b-g6hns6) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L73)):*

Since this is the main function, `RasterizePreviousInputOutputAndExportToMarkdown` should also 
</blockquote></details>

I don't think you finished the sentence?
___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MN5WBYPEP7_zIs2ex0o:bi8cv4g) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

Are the spaces for this comma deliberate? I would assume the spacing should be the same as the if/else branch. E.g:
```
If[True,
  a
  ,
  b
]
```

</blockquote></details>

Yeah, they are deliberate. I put the comma at the level of indentation If statement that started it and the ] that finish it. I think it's clearer.
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MN5WTPD81asMNmOnYgk:bignh3) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

I notice there is no `In[] := `. Maybe I missed the discussion but I thought this was mandatory?
</blockquote></details>

Yeah, I'm not sure. I thought about that but it wouldn't do the right thing for multi-line statements. Also, if you tried to copy-paste it and run it you'd get:

```
In[] := Print["HELLO"];
SetDelayed::write: Tag In in In[] is Protected.
```

Which seems pretty confusing.

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MN5__VHBdWJiVkpQXET:bmn9a0w) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

Just to be clear, you are NOT getting the following behavior? If so, which platform are you on?


/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-26 at 23.43.17@2x.png
</blockquote></details>

It doesn't. Should be Linux's fault:



```
In[14]:= $Version

Out[14]= "12.2.0 for Linux x86 (64-bit) (October 14, 2020)"
```

By the way `Range` was not being properly rasterized for some reason:

___
*[DevUtils/Rasterization.m, line 73 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4dktQ2yb8v_qOeo4g:-MN5_g7o-2eu1wl_hn6l:b-w0q6ot) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L73)):*

I don't think you finished the sentence?
</blockquote></details>

:( I was referencing the lack of `Out[] :=`, but you already addressed it in other comment.
___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MN5_wTA-p-vElB2bC4z:b-m4w8md) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

Yeah, they are deliberate. I put the comma at the level of indentation If statement that started it and the ] that finish it. I think it's clearer.
</blockquote></details>

Interesting.
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MN5a2iq-0-Uld_SBxvw:bhyjb0t) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

Yeah, I'm not sure. I thought about that but it wouldn't do the right thing for multi-line statements. Also, if you tried to copy-paste it and run it you'd get:

```
In[] := Print["HELLO"];
SetDelayed::write: Tag In in In[] is Protected.
```

Which seems pretty confusing.
</blockquote></details>

In multi-line statements the `In[] :=` appears in the first line:
```
In[12]:= 1 + 1;
Range[5]

Out[13]= {1, 2, 3, 4, 5}
```

It is indeed confusing, but that was the convention we were using for documentation. Do you suggest to add the In[]'s manually?

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MN5c09n2gsy8GB0K6O7:b2zktb8) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

It doesn't. Should be Linux's fault:



```
In[14]:= $Version

Out[14]= "12.2.0 for Linux x86 (64-bit) (October 14, 2020)"
```

By the way `Range` was not being properly rasterized for some reason:


</blockquote></details>

Wait, I'm confused: all of your *screenshots* are missing the In[] and Out[] as well. Why is that? How did you take those screenshots? Also 12.2 isn't released yet. Have you tried this on 12.1? 
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MN5ccNS4zz5532QqDQb:bpcxi8e) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

In multi-line statements the `In[] :=` appears in the first line:
```
In[12]:= 1 + 1;
Range[5]

Out[13]= {1, 2, 3, 4, 5}
```

It is indeed confusing, but that was the convention we were using for documentation. Do you suggest to add the In[]'s manually?
</blockquote></details>

We should probably get some more opinions on this, including @maxitg . Personally I would say that the ideal here is to rasterize the input as well, since it makes the font, text size etc consistent with what the output looks like, but add javascript "onclick" event handler to copy the underlying code (minus the In[]) to your clipboard when you click the image. Copy-paste ability is the main reason to prefer code blocks here over rasterization (that and a small improvement in file size, I suppose).

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MN5hMFB4i3lrILsSE56:b4mepc0) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

Wait, I'm confused: all of your *screenshots* are missing the In[] and Out[] as well. Why is that? How did you take those screenshots? Also 12.2 isn't released yet. Have you tried this on 12.1? 
</blockquote></details>

Sorry, I cropped the image really bad:


This is:
```
In[6]:= $Version

Out[6]= "12.1.1 for Linux x86 (64-bit) (June 19, 2020)"
```
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MN5hteC3GGzc2B02SUX:bonwmli) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

We should probably get some more opinions on this, including @maxitg . Personally I would say that the ideal here is to rasterize the input as well, since it makes the font, text size etc consistent with what the output looks like, but add javascript "onclick" event handler to copy the underlying code (minus the In[]) to your clipboard when you click the image. Copy-paste ability is the main reason to prefer code blocks here over rasterization (that and a small improvement in file size, I suppose). 
</blockquote></details>

I like the idea of having a "copy to clipboard" button.

maxitg (changes requested):
Please add examples (with pictures) to the PR message.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535#-:-MNOBvhlCh1XxjWgfEu5:b-ql4qh6)* status: 0 of 2 files reviewed, 16 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MNOCm2p229y-Ii0MCwo:b-dpub09) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

Interesting.
</blockquote></details>

The comma is an "else", so I agree with @taliesinb, it makes sense for the same reason it makes sense in C++. I also agree that it's better to put the comma on a new line because otherwise, it's very hard to spot where the first part of the `If` ends, and the "else" starts. I think we should standardize this for the rest of the codebase.
___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MNODeFXAc7ug6hC3qKI:b-s6nqu8) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

I like the idea of having a "copy to clipboard" button.
</blockquote></details>

Ok, first, a general comment. If we are going to change any convention, that needs to be done in a separate PR including the changes to all ***existing*** code. Therefore, this PR has to include `In[]` and `Out[]` labels unless there is a preceding PR changing the convention.

The convention right now is to have `In[]` to match the `Out[]`. If there is a single-input multi-line code like this:

```wl
In[] := WolframModel[1 -> 2,
                     {1},
                     5]
```

`In[]` goes in the beginning. If there is other code that needs to be run before the input, it would look like this (otherwise it would look awekward if there are function definitions etc.):

```wl
x = 2 + 2;
In[] := x + 4
Out[] = 8
```

I admit click-to-copy would be ideal, but until we have that, let's keep `In[]` and `Out[]` labels. If we remove the `In[]`, we will have to remove `Out[]` as well for consistency, and then it will be hard to tell what's going on.
___
*[DevUtils/Rasterization.m, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO5Fmf5xCEd7QEI-8h:-MNO5Fmg0vJOa-uSMgEe:bllfmai) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L38)):*
> ```objc
> Options[RasterizeExpressionAndExportToMarkdown] = $exportOptions;
> 
> RasterizeExpressionAndExportToMarkdown[relativePath_, expr_, opts:OptionsPattern[]] := CatchFailureAsMessage @ Scope[
> ```

Check that `relativePath` is actually a string.
___
*[DevUtils/Rasterization.m, line 75 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNOBF9E4ZGUG34BNIzj:-MNOBF9E4ZGUG34BNIzk:b-cjc235) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L75)):*
> ```objc
> code block." <> $usageSuffix];
> 
> Options[RasterizePreviousInputOutputAndExportToMarkdown] = Append[$exportOptions, "RasterizeInput" -> False];
> ```

It would be useful to have some kind of a "dry run" option, where it produces the output, but not writes any files. It's useful for generating images for PR messages and comments where one needs to know the markdown code but doesn't need any files.
___
*[DevUtils/Rasterization.m, line 101 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO7pZB2gAjJ_gr8adp:-MNO7pZB2gAjJ_gr8adq:br0o1r7) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L101)):*
> ```objc
> 
> RasterizePreviousInputOutputAndExportToMarkdown[relativePath_, opts:OptionsPattern[]] := CatchFailureAsMessage @ Scope[
>   ocell = PreviousCell[]; icell = PreviousCell @ ocell;
> ```

If it has `InputOutput` in the name, it should actually search for input and output cells. Otherwise, it does not make sense in cases like this:

```wl
In[] := WolframModel[1 -> 2, {1}]["2"]

In[] := RasterizePreviousInputOutputAndExportToMarkdown["Test.png"]
```



Why don't we do something like this instead:

1. Allow the user to manually specify which cells they want to rasterize by specifying a list of cell objects.
2. Alternatively, one can pass a negative number specifying how many previous cells should be rasterized (e.g., `-2` would be similar to this function).

Doing this will eliminate the need for 2 functions instead of one.
___
*[DevUtils/Rasterization.m, line 110 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO9Th25H_6rdFD7uFg:-MNO9Th25H_6rdFD7uFh:bcouhkf) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L110)):*
> ```objc
>     icellString = cellToString[icell];
>   ];
>   If[StringQ[icellString],
> ```

The decision of which cell should be rasterized should be made based on the cell's contents, not on whether it's an input or an output.

For example, in this case both input and output should ***not*** be rasterized:

```wl
In[] := 2 + 2
Out[] = 4
```

In this case, the input should be rasterized, but the output should not be:



```wl
Out[] = 1
```

I.e., we need to figure out if the input/output have any non-trivial boxes to decide if they should be rasterized.

taliesinb (commented):
Here's an example of `RasterizePreviousInputOutputAndExportToMarkdown` being used (with the DryRun option given to make the example easier to see):

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.56.50@2x.png

This shows the behavior when the input cannot be stringified: 


/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.56.12@2x.png

This shows the behavior of `RasterizePreviousOutputAndExportToMarkdown`:

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.57.23@2x.png

This shows the behavior of `RasterizeExpressionAndExportToMarkdown`:

/Users/taliesinb/Library/Application Support/PixelSnap/media/PixelSnap 2020-11-30 at 16.58.11@2x.png
 

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535#-:-MNOdajf8kzAL1KCBlWY:b-ls8xi8)* status: 0 of 2 files reviewed, 16 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cjy27IUO3QoYkZVI:-MNO_rBk0a8CvhxpNPXS:b-z5d0sa) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L112)):*

Ok, first, a general comment. If we are going to change any convention, that needs to be done in a separate PR including the changes to all ***existing*** code. Therefore, this PR has to include `In[]` and `Out[]` labels unless there is a preceding PR changing the convention.

The convention right now is to have `In[]` to match the `Out[]`. If there is a single-input multi-line code like this:

```wl
In[] := WolframModel[1 -> 2,
                     {1},
                     5]
```

`In[]` goes in the beginning. If there is other code that needs to be run before the input, it would look like this (otherwise it would look awekward if there are function definitions etc.):

```wl
x = 2 + 2;
In[] := x + 4
Out[] = 8
```

I admit click-to-copy would be ideal, but until we have that, let's keep `In[]` and `Out[]` labels. If we remove the `In[]`, we will have to remove `Out[]` as well for consistency, and then it will be hard to tell what's going on.
</blockquote></details>

I'll try to automatically insert an `In[]` before the final expression in the input string.
___
*[DevUtils/Rasterization.m, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO5Fmf5xCEd7QEI-8h:-MNOYBFy8YYqY9848gxK:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L38)):*

Check that `relativePath` is actually a string.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 75 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNOBF9E4ZGUG34BNIzj:-MNOYfxGEaJ5F2xBx7rx:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L75)):*

It would be useful to have some kind of a "dry run" option, where it produces the output, but not writes any files. It's useful for generating images for PR messages and comments where one needs to know the markdown code but doesn't need any files.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 101 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO7pZB2gAjJ_gr8adp:-MNOZD6T7QAHQquHaFtU:b-iqb3jy) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L101)):*

If it has `InputOutput` in the name, it should actually search for input and output cells. Otherwise, it does not make sense in cases like this:

```wl
In[] := WolframModel[1 -> 2, {1}]["2"]

In[] := RasterizePreviousInputOutputAndExportToMarkdown["Test.png"]
```



Why don't we do something like this instead:

1. Allow the user to manually specify which cells they want to rasterize by specifying a list of cell objects.
2. Alternatively, one can pass a negative number specifying how many previous cells should be rasterized (e.g., `-2` would be similar to this function).

Doing this will eliminate the need for 2 functions instead of one.
</blockquote></details>

Idea 1 will not be friendly to use for the typical case (which *is* just an input/output pair), because the way of getting the two previous cells would require you to run something like this:

```
output = PreviousCell[];
input = PreviousCell[output];
ExportCellsToMarkdown["mypath.png", {input, output}];
```

That's seems needlessly complicated for the common case. 

While I like the idea behind 2, this seems to be general enough that we start to run into problems: if you pass e.g. n=5, but some of the cells are textual and others are not, and the textual and non-textual 'runs' do not split into just two contiguous groups, e.g. the cells are {text, text, image, text, image}, then it will we will need to make *two* images rather than one, and it is not clear then how to then name those images, since the user gave a *single* file name.  

Stepping back, it feels like this suggestion is pushing these functions to become a general-purpose notebook-to-markdown conversion tool. I do think this is a highly valuable technology to have (especially if it can be made bidirectional), but I think we will need to rethink the conventions of how images are stored, and move to a content-based scheme. It's out of scope for this PR, which was always meant to be a temporary solution to help us until a more general technology is ready.

To deal with your example, I'm happy to make this function simply fail if the two cells it picks up are not an input followed by an output. That should make it consistent with the name. 
___
*[DevUtils/Rasterization.m, line 110 at r2](https://reviewable.io/reviews/maxitg/setreplace/535#-MNO9Th25H_6rdFD7uFg:-MNO_cIQACp52CcDZECp:b-catt7m) ([raw file](https://github.com/maxitg/setreplace/blob/5a681021b7abea4d765560b02c32e167f84a4284/DevUtils/Rasterization.m#L110)):*

The decision of which cell should be rasterized should be made based on the cell's contents, not on whether it's an input or an output.

For example, in this case both input and output should ***not*** be rasterized:

```wl
In[] := 2 + 2
Out[] = 4
```

In this case, the input should be rasterized, but the output should not be:



```wl
Out[] = 1
```

I.e., we need to figure out if the input/output have any non-trivial boxes to decide if they should be rasterized.
</blockquote></details>

This connects with my earlier point about generality: if both cells are purely textual, then the relative path argument ends up being completely ignored. This tells us that such generality is inappropriate for this function, which is designed around the idea of exporting a single image, corresponding to an input. 

I agree that there is a more powerful and general technology which converts between notebooks and markdown, which can completely subsume these functions, but if you want to wait for that, we should just reject this PR and implement that general technology instead. That might take weeks to land, and in the meantime we'll still be making images by hand in the old-fashioned way, so I don't think that's the right strategy. We can always remove these functions later when the general technology is ready.

maxitg (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 89 at r3](https://reviewable.io/reviews/maxitg/setreplace/535#-MNPBO0qABeZFOFdmeI6:-MNPBO0qABeZFOFdmeI7:bmjx8k4) ([raw file](https://github.com/maxitg/setreplace/blob/a57b20e41ef818647e37e09d4e59726762e961db/DevUtils/Rasterization.m#L89)):*
> ```objc
> 
> $maxWidthCompensation = 0.9;
> rasterize[maxWidth_, epxr_Image] := expr;
> ```

`epxr` -> `expr`.
___
*[DevUtils/Rasterization.m, line 148 at r3](https://reviewable.io/reviews/maxitg/setreplace/535#-MNPHheKEGDIjfRUdJfe:-MNPHheKEGDIjfRUdJff:b-1umnr5) ([raw file](https://github.com/maxitg/setreplace/blob/a57b20e41ef818647e37e09d4e59726762e961db/DevUtils/Rasterization.m#L148)):*
> ```objc
>   filename = FileNameTake[relativePath];
>   relativeDir = FileNameDrop[relativePath];
>   If[Or[filename === "",!UpperCaseQ[StringTake[filename, 1]],
> ```

Nit: Put the arguments to the `Or` one per line to make it more readable.

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNTqRHr1iaMmqlj7Ky4:b-dzemyk) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

Sorry, I cropped the image really bad:


This is:
```
In[6]:= $Version

Out[6]= "12.1.1 for Linux x86 (64-bit) (June 19, 2020)"
```
</blockquote></details>

@daneelsan, can you try running the following, and tell me what you get?

First,

```wl
In[] := Graph[1 -> 2]
```

Then, on the next cell,

```wl
In[] := Rasterize[
 Notebook[{Cell[BoxData[""], "Input"], NotebookRead[PreviousCell[]]}]]
```

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNTsX315y3gCnSDVCYo:bnwmzc0) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

@daneelsan, can you try running the following, and tell me what you get?

First,

```wl
In[] := Graph[1 -> 2]
```

Then, on the next cell,

```wl
In[] := Rasterize[
 Notebook[{Cell[BoxData[""], "Input"], NotebookRead[PreviousCell[]]}]]
```
</blockquote></details>


___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MNTsDry5IQdZmSzb-D3:b-irtzlj) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

The comma is an "else", so I agree with @taliesinb, it makes sense for the same reason it makes sense in C++. I also agree that it's better to put the comma on a new line because otherwise, it's very hard to spot where the first part of the `If` ends, and the "else" starts. I think we should standardize this for the rest of the codebase.
</blockquote></details>

Sure, that should be in the `Contributing` section then. The same happens with `Switch`?

```
Switch[a,
  b, c
,
  d, e
]
```

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNTxtEOAVxJmfYsNV1Z:bdxjjnz) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

Ok, what do you get for this?

```wl
In[] := Rasterize[
 Notebook[{Cell[BoxData[""], "Input"], 
   Cell[BoxData["4"], "Output", CellLabel -> "label"]}]]
```

And if you don't get a label, do you get it if you open the notebook instead of rasterizing?

```wl
In[] := CreateDocument[
 Notebook[{Cell[BoxData[""], "Input"], 
   Cell[BoxData["4"], "Output", CellLabel -> "label"]}]]
```
</blockquote></details>

maxitg (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 16 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNZJhI4EG9p6a3LyK2H:b-em8n1f) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

This looks like a bug in Mathematica itself. Perhaps we can report it upstream?
</blockquote></details>

I reported it but given the amount of existing `Rasterize` weeds, I'm not too optimistic it will be fixed any time soon. Looks like we'll have to draw the label manually.

It's very important to fix this because Linux is the one platform where Mathematica UI does not have a "Copy As Bitmap" button, so this functionality is necessary to write documentation.
___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MNZfvYHAThKXpI76KYg:b-ux545x) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

That looks good. Should we open a separate issue for this change (and in general do we go through the codebase and rewrite these cases when they occur so that everything is consistent)? 
</blockquote></details>

Opened: #541.
___
*[DevUtils/Rasterization.m, line 49 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNQdR4R2MZ075gJlf:-MNZNQdR4R2MZ075gJlg:b-og1wut) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L49)):*
> ```objc
> 
> SetUsage @ Evaluate["
> RasterizeCellsAndExportToMarkdown['path$', cells$] will rasterize a cell or set of cells, write the result to 'path$', \
> ```

or ***a*** set of cells
___
*[DevUtils/Rasterization.m, line 51 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZN_cX2Kbm5jMovwew:-MNZN_cY4K80jnw0X1aw:b9utm1k) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L51)):*
> ```objc
> RasterizeCellsAndExportToMarkdown['path$', cells$] will rasterize a cell or set of cells, write the result to 'path$', \
> and return an HTML  tag that can be pasted directly into a markdown file.
> * Cells can be Cell[$$] expressions or CellObject[$]] expressions (which will be read with NotebookRead).
> ```

`CellObject` has an extra closing bracket and a missing `$`.
___
*[DevUtils/Rasterization.m, line 52 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNqkS75cCXqwNJEcU:-MNZNqkS75cCXqwNJEcV:b-3v5srm) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L52)):*
> ```objc
> and return an HTML  tag that can be pasted directly into a markdown file.
> * Cells can be Cell[$$] expressions or CellObject[$]] expressions (which will be read with NotebookRead).
> * The resulting images WILL include cell labels (In[]=, Out[]=, etc)." <> $usageSuffix
> ```

Why plural images? Doesn't it produce a single image?
___
*[DevUtils/Rasterization.m, line 52 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNtxA0F3uyTsNclp9:-MNZNtxA0F3uyTsNclpA:b-chb1k1) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L52)):*
> ```objc
> and return an HTML  tag that can be pasted directly into a markdown file.
> * Cells can be Cell[$$] expressions or CellObject[$]] expressions (which will be read with NotebookRead).
> * The resulting images WILL include cell labels (In[]=, Out[]=, etc)." <> $usageSuffix
> ```

It's `In[]:=`, with a `RuleDelayed`.

taliesinb (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: 1 of 2 files reviewed, 16 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNZiBD02T3K5GOi4ZXh:bpvj5vl) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

I reported it but given the amount of existing `Rasterize` weeds, I'm not too optimistic it will be fixed any time soon. Looks like we'll have to draw the label manually.

It's very important to fix this because Linux is the one platform where Mathematica UI does not have a "Copy As Bitmap" button, so this functionality is necessary to write documentation.
</blockquote></details>

Ok, can I suggest we do that in a PR?
___
*[DevUtils/Rasterization.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4cE1D0fF2qAMhohq4:-MNZiMjiEU1Ys7HIIFl_:b-5fqpeq) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L108)):*

Opened: #541.
</blockquote></details>

Great! Thanks.
___
*[DevUtils/Rasterization.m, line 49 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNQdR4R2MZ075gJlf:-MNZhyHdDga7IVjv16-4:bar2upv) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L49)):*

or ***a*** set of cells
</blockquote></details>

In English, "a dog or cat" is equivalent to "a dog or a cat". The "a" distributes over the "or", so-to-speak.
___
*[DevUtils/Rasterization.m, line 51 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZN_cX2Kbm5jMovwew:-MNZiAV9ESMzOg8WM8O9:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L51)):*

`CellObject` has an extra closing bracket and a missing `$`.
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 52 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNqkS75cCXqwNJEcU:-MNZiL8HAMwBwMEIuWPZ:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L52)):*

Why plural images? Doesn't it produce a single image?
</blockquote></details>

Done.
___
*[DevUtils/Rasterization.m, line 52 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNtxA0F3uyTsNclp9:-MNZiLSuDmfzKYOsyq8h:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L52)):*

It's `In[]:=`, with a `RuleDelayed`.
</blockquote></details>

Done.

maxitg (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 12 unresolved discussions (waiting on @daneelsan)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNZjC4bBXanuf0eIAXw:bnlr184) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

Ok, can I suggest we do that in a PR?
</blockquote></details>

I don't object to doing it in another PR, but it's really a question for @daneelsan.
___
*[DevUtils/Rasterization.m, line 49 at r4](https://reviewable.io/reviews/maxitg/setreplace/535#-MNZNQdR4R2MZ075gJlf:-MNZixulASCu78QpEXw-:b-ahle5t) ([raw file](https://github.com/maxitg/setreplace/blob/9c82cab5c1885eedf1047effe623b01886ef8c95/DevUtils/Rasterization.m#L49)):*

In English, "a dog or cat" is equivalent to "a dog or a cat". The "a" distributes over the "or", so-to-speak.
</blockquote></details>

Good to know!

daneelsan (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/535)* status: all files reviewed, 12 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[DevUtils/Rasterization.m, line 51 at r1](https://reviewable.io/reviews/maxitg/setreplace/535#-MN4eUEwAFM0WkKoEuVs:-MNZmzH-0T7m8xJ8FTUu:b-nxkptb) ([raw file](https://github.com/maxitg/setreplace/blob/222bf94443518b70499e2c99161dac8a8e3a04a8/DevUtils/Rasterization.m#L51)):*

I don't object to doing it in another PR, but it's really a question for @daneelsan.
</blockquote></details>

I remember that the rasterization worked correctly (Out[] label) in the old PR, so I'm not sure what happened here.
Still, that should not stop this PR, and should be fixed separately.

taliesinb:
@daneelsan that worked because I was manually attaching the cell label myself. But I couldn't do that accurately enough (e.g. the baselines for Out[] was wrong in general). I thought that `Rasterize[Notebook[..]]` would just do all of that stuff for me, but apparently it doesn't work on Linux. I'll try add that code back to support Linux as soon as I can.

PR #543, opened 2020-12-04
Add parallelism files to the Xcode project

maxitg:
## Changes
* `parallelism*` files were missing from the Xcode project, which caused the build to fail.
* They are now added, so both the library build and the tests pass if run from Xcode.

## Comments
* If you don't use Xcode, I don't expect anyone to specifically install it just to try this out, but someone needs to review this PR even if nobody uses Xcode except for me.

## Examples
Screen Shot 2020-12-04 at 09 51 19

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/543)


taliesinb:
I have XCode, but how do I test this?

maxitg:
You can just open `SetReplace.xcodeproj` in the repository root, and try building the `SetReplace` target and running the `SetReplaceTests` target.

taliesinb:
Builds and runs and passes tests under XCode!

taliesinb (approved):

PR #545, opened 2020-12-04
Fix Xcode warnings

maxitg:
## Changes
* Fixes warnings generated when compiling with Xcode:
  * There was some implicit type conversion.
  * The macOS deployment target is updated to 11.0 (Big Sur).

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/545)

### DIFF
--- a/libSetReplace/Match.cpp
+++ b/libSetReplace/Match.cpp
@@ -221,7 +221,7 @@ class Matcher::Implementation {
       const uint64_t& numThreadsToUse = token->numThreads();
 
       auto addMatchesForRuleRange = [=](RuleID start) {
-        for (RuleID i = start; i < static_cast<int>(rules_.size()); i += numThreadsToUse) {
+        for (RuleID i = start; i < static_cast<RuleID>(rules_.size()); i += numThreadsToUse) {
           addMatchesForRule(expressionIDs, i, shouldAbort);
         }
       };
@@ -237,7 +237,7 @@ class Matcher::Implementation {
         }
       } else {
         // Single-threaded path
-        for (RuleID i = 0; i < static_cast<int>(rules_.size()); ++i) {
+        for (RuleID i = 0; i < static_cast<RuleID>(rules_.size()); ++i) {
           addMatchesForRule(expressionIDs, i, shouldAbort);
         }
       }
--- a/libSetReplace/Match.cpp
+++ b/libSetReplace/Match.cpp
@@ -220,8 +220,8 @@ class Matcher::Implementation {
       const auto token = Parallelism::acquire(Parallelism::HardwareType::StdCpu, rules_.size());
       const uint64_t& numThreadsToUse = token->numThreads();
 
-      auto addMatchesForRuleRange = [=](uint64_t start) {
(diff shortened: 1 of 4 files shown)


libSetReplace/Match.cpp
```diff
@@ -220,8 +220,8 @@ class Matcher::Implementation {
       const auto token = Parallelism::acquire(Parallelism::HardwareType::StdCpu, rules_.size());
       const uint64_t& numThreadsToUse = token->numThreads();
 
-      auto addMatchesForRuleRange = [=](uint64_t start) {
-        for (uint64_t i = start; i < static_cast<uint64_t>(rules_.size()); i += numThreadsToUse) {
+      auto addMatchesForRuleRange = [=](RuleID start) {
+        for (RuleID i = start; i < static_cast<int>(rules_.size()); i += numThreadsToUse) {
```
aokellermann, line 224: You should `static_cast` to `RuleID` rather than `int`. Casting to `int` only "works" because they happen to be equivalent.

libSetReplace/Match.cpp
```diff
@@ -237,7 +237,7 @@ class Matcher::Implementation {
         }
       } else {
         // Single-threaded path
-        for (size_t i = 0; i < rules_.size(); ++i) {
+        for (RuleID i = 0; i < static_cast<int>(rules_.size()); ++i) {
```
aokellermann, line 240: Same here.

aokellermann (changes requested):

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/545)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @aokellermann)

___
*[libSetReplace/Match.cpp, line 224 at r1](https://reviewable.io/reviews/maxitg/setreplace/545#-MNiYAEJrWwjBRNV1R5C-r1-224:-MNif3hW0DrzO7qLd_KM:b-qu6n94) ([raw file](https://github.com/maxitg/setreplace/blob/45f160f48d30db7881c416b8eab8c955dc1f0679/libSetReplace/Match.cpp#L224)):*

You should `static_cast` to `RuleID` rather than `int`. Casting to `int` only "works" because they happen to be equivalent.
</blockquote></details>

Good point. Done.
___
*[libSetReplace/Match.cpp, line 240 at r1](https://reviewable.io/reviews/maxitg/setreplace/545#-MNiYAEJrWwjBRNV1R5C-r1-240:-MNif52i3TkTnOvHxlRK:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/45f160f48d30db7881c416b8eab8c955dc1f0679/libSetReplace/Match.cpp#L240)):*

Same here.
</blockquote></details>

Done.

aokellermann (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/545)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #568, opened 2020-12-10
CI: Build multi-platform paclet

### LINKED ISSUE #552: Paclet builds in CI
### The problem
Currently, paclets (e.g., for releases) are build using internal Wolfram infrastructure. That has a number of issues the main one being that these builds are inaccessible to anyone who is not a Wolfram employee. We should move the paclet building to Circle CI instead.

That will bring a number of advantages:
1. Every collaborator will be able to access it and download paclets.
2. We will have full access to the configuration which will be store purely in YAML files. Currently, even I don't have access to configuring the build settings.
3. Circle CI is much faster especially with respect to the time it takes for the builds to start (due to its scale).
4. It will be completely automatic. We can build a multi-platform paclet for every commit in every branch. Currently, one has to start the builds manually. This will even allow us to do things like CD where users will always have the paclet built from the latest master with zero work on our side.

### Possible solution
Circle CI gives free Linux and Windows credits to open source projects. They appear to give Mac credits as well but there is an approval process, and we don't have access yet (I don't yet know what it will take).

Due to #547, Mathematica installation is no longer required to build the library, so we only need a compiler on Windows and Mac, which means we almost certainly can build it there. We can also publish artifacts since #540, and it's likely possible to pick up the artifacts by other steps in the build process.

So, the main issue is to configure builds on Mac and Windows #371 (ideally using CMake) and request Mac credits from Circle CI.

### Alternative solutions
Install Mathematica on Mac and Windows will make the build process more consistent (since library metadata JSONs will be generated by the same script, and there will be no build code duplication). However, even on Linux Docker downloading WL image takes a long time; we will have a limited number of credits, and installing it on Mac and Windows will be even more difficult.

maxitg:
## Changes
* Closes #552.
* Renames CI jobs to explicitly say if they do building or testing.
* Copies the libraries from macos-build and windows-build, and puts them into the paclet created in wolfram-language-paclet-test before testing.
* As a result, the paclet-artifact from wolfram-language-paclet-test now contains libraries for all platforms.
* Also updates contributing guide accordingly to describe the new workflow and remove references to the internal build process.

## Examples
* Here is [the workflow](https://app.circleci.com/pipelines/github/maxitg/SetReplace/1454/workflows/91394b90-7214-439f-95f7-3e4aa617784b).

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/568)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
   win: circleci/windows@2.2.0
 
 jobs:
-  wolfram-language:
+  wolfram-language-paclet-test:
     docker:
       - image: maxitg/set-replace-wl-ci:12.1.1
         auth:
@@ -20,6 +20,13 @@ jobs:
 
       - store_artifacts:
           path: ./LibraryResources/
+
+      - attach_workspace:
+          at: /tmp/workspace
+
+      - run:
+          name: Copy libraries from other platforms
+          command: cp -r /tmp/workspace/* ./LibraryResources/
 
       - run:
           name: Install
@@ -39,7 +46,8 @@ jobs:
       - run:
           name: Performance Test
           command: ./performanceTest.wls master @HEAD 2
-  cpp:
+
+  cpp-test:
     docker:
       - image: alpine:3.12.1
         auth:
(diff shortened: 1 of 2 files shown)


aokellermann (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/568)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @taliesinb)

PR #569, opened 2020-12-10
Remove internal Wolfram build scripts

maxitg:
## Changes
* Removes internal Wolfram RE scripts, since, after the merge of #552, they are no longer necessary.
* Also, the scripts removed here don't work right now.

## Comments
* This removes all RE scripts, the test script, the XML file with the config, and the internal `GitLink` installation code (since it requires VPN).
* Documentation is updated in #552.
* One can tell these scripts are for the internal build because they are not used anywhere.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/569)


daneelsan (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/569)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @taliesinb)

PR #576, opened 2020-12-17
CI: Fix CMake install script to use the new universal archive

maxitg:
## Changes
* Due to the ongoing change of Macs from x86 to ARM architecture, the archives with CMake releases are now named `-universal` instead of `-Darwin-x86_64`. As a result, macOS CI fails because it cannot find the macOS CMake in the most recent release.
* This fixes the name to search for.

## Comments
* Perhaps, we should be using a specific version to prevent this from happening in the future. However, the architecture changes like this are rare, and it does save time to not have to update CMake manually, so perhaps we can leave it like this until/if the next issue arises.

## Examples
* [CI job](https://app.circleci.com/pipelines/github/maxitg/SetReplace/1501/workflows/52549d15-c635-4143-8650-667c088f390b/jobs/2363) now succeeds.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/576)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,10 +134,10 @@ jobs:
               curl -s https://api.github.com/repos/Kitware/CMake/releases/latest \
               | grep 'browser_download_url' \
               | cut -d\" -f4 \
-              | grep Darwin-x86_64.tar.gz)
-            curl -L --output cmake-Darwin-x86_64.tar.gz $cmakeURL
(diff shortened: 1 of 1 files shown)


aokellermann (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/576)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #578, opened 2020-12-17
Parallel WL Tests on CI

### LINKED ISSUE #448: Parallelize Circle CI jobs
### The problem
We currently have the ability to use up to 4 parallel runs per CI job in Circle CI. However, we are not using that at the moment.

### Possible solution
We can split WL and C++ tests into groups that we can run in parallel. Take a look at [Circle CI docs](https://circleci.com/docs/2.0/parallelism-faster-jobs/).

aokellermann:
## Changes
* Adds 4x parallelism to WL tests on CI
* Closes #448 

## Comments
Currently, we do not have WL [test metadata](https://circleci.com/docs/2.0/collect-test-data/) compatible with CircleCI. Therefore, we can't use CircleCI's built-in [split-by timing](https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-by-timing-data) to distribute tests efficiently. This PR simply runs `matching` and `WolframModel` (which take the most time by far) each in separate executors, and splits the rest among the last two. 

In order to optimize this further, we could:
1. Distribute tests from `matching` and `WolframModel` into separate files.
2. Implement JUnit XML or Cucumber JSON test result reports for WL and enable `split-by timing`.

## Examples
Parallel executors:
image

Test time is reduced by ~10 minutes (over ~2x reduction):
image

compared to master:
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/578)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,8 +45,9 @@ jobs:
           command: |
             ./.circleci/test.sh $(circleci tests glob "Tests/*.wlt" \
               | circleci tests split --split-by=timings \
+              | sed "s/\.wlt//" \
               | sed "s/Tests\///" \
-              | sed "s/performance//")
+              | sed "/performance/d")
 
       - run:
           name: Performance Test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,11 @@ jobs:
 
       - run:
           name: Performance Test
-          command: ./performanceTest.wls master @HEAD 2
+          command: |
+            if [ $CIRCLE_NODE_INDEX -eq 0 ]
+            then
+              ./performanceTest.wls master @HEAD 2
+            fi
 
   cpp-test:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,11 +43,22 @@ jobs:
       - run:
           name: Test
           command: |
-            ./.circleci/test.sh $(circleci tests glob "Tests/*.wlt" \
-              | circleci tests split --split-by=filesize \
-              | sed "s/\.wlt//" \
-              | sed "s/Tests\///" \
-              | sed "/performance/d")
+            if [ $CIRCLE_NODE_INDEX -eq 2 ]
+            then
+              testsToRun=matching
+            elif [ $CIRCLE_NODE_INDEX -eq 3 ]
+            then
+              testsToRun=WolframModel
+            else
+              testsToRun=$(circleci tests glob "Tests/*.wlt" \
+                | sed "/Tests\/performance.wlt/d" \
+                | sed "/Tests\/matching.wlt/d" \
+                | sed "/Tests\/WolframModel.wlt/d" \
+                | circleci tests split --total=2 --split-by=filesize \
+                | sed "s/\.wlt//" \
+                | sed "s/Tests\///")
+            fi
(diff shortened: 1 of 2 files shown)


maxitg (changes requested):
This is incredible! Thanks!

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/578#-:-MOmznfC6oIdtQeptyi-:bc9p58b)* status: all files reviewed, 3 unresolved discussions (waiting on @aokellermann)

___
*[.circleci/config.yml, line 9 at r1](https://reviewable.io/reviews/maxitg/setreplace/578#-MOmysUI5lT8ZBRjb0dN:-MOmysUI5lT8ZBRjb0dO:b-hr651w) ([raw file](https://github.com/maxitg/setreplace/blob/aefc018bb376bb435a281ab3546ced31b2504c13/.circleci/config.yml#L9)):*
> ```YAML
>   wolfram-language-paclet-test:
>     docker:
>       - image: maxitg/set-replace-wl-ci:12.1.1
> ```

It seems redundant that we are downloading the image 4 times separately. There is definitely a way around it, but we might have to download WL outside of the Docker image and put it into the `workspace` (not in this PR, obviously).
___
*[.circleci/config.yml, line 23 at r1](https://reviewable.io/reviews/maxitg/setreplace/578#-MOmzKFY0LAZa_k7wmGC:-MOmzKFY0LAZa_k7wmGD:b-ke0jul) ([raw file](https://github.com/maxitg/setreplace/blob/aefc018bb376bb435a281ab3546ced31b2504c13/.circleci/config.yml#L23)):*
> ```YAML
> 
>       - store_artifacts:
>           path: ./LibraryResources/
> ```

Similarly, we don't need to store artifacts from every parallel run. The way to fix that would be to build in a separate job, however, doing that will cost us ~60 secs for another WL download, so it will need to be done after the previous comment.
___
*[.circleci/config.yml, line 65 at r1](https://reviewable.io/reviews/maxitg/setreplace/578#-MOmyhNBED8NoxjGgwLt:-MOmyhNC48V7ILAbCmzf:b-aovodo) ([raw file](https://github.com/maxitg/setreplace/blob/aefc018bb376bb435a281ab3546ced31b2504c13/.circleci/config.yml#L65)):*
> ```YAML
>       - run:
>           name: Performance Test
>           command: ./performanceTest.wls master @HEAD 2
> ```

If we only run this on the zeroth or the first node, we will save another ~30 secs.

aokellermann (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/578)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @maxitg)

___
*[.circleci/config.yml, line 65 at r1](https://reviewable.io/reviews/maxitg/setreplace/578#-MOmyhNBED8NoxjGgwLt:-MOn7nvw2yaSOSO_PH93:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/aefc018bb376bb435a281ab3546ced31b2504c13/.circleci/config.yml#L65)):*

If we only run this on the zeroth or the first node, we will save another ~30 secs.
</blockquote></details>

Done.

maxitg (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/578)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #591, opened 2021-01-11
Replacing makeMessage with General::msg

### LINKED ISSUE #79: Use General:: messages that apply to multiple functions
There are some messages in the package that are used for multiple symbols, such as messages about unsupported cases for C++ implementation can be produced for `WolframModel`, `SetReplace`, and other functions.

Currently, messages for an appropriate function are produced and registered, and then called on the fly. However, there is a language mechanism to deal with those, specifically, `General::` messages, which should be used instead.

daneelsan:
## Changes

* Closes #79.
* Replaces all instances of `makeMessage[caller, "msg", args___]` with `Message[caller::msg, args]`

## Comments

* Renamed the `"nonopt"` message to `"nonPropertyOpt"` as there already exists `General::nonopt`
```wl
In[] := General::nonopt
Out[] = "Options expected (instead of `1`) beyond position `2` in `3`. An option must be a rule or a list of rules."
```

* Renamed the file `makeMessage.m` to `messages.m`. This file still contains all the `General::msg` definitions.

## Examples

```wl
In[] := General::invalidMethod
Out[] = "Method should be one of Automatic, \"LowLevel\" or \"Symbolic\"."
```

```wl
In[]:= General::pargx
Out[]= "Property \"`1`\" requested with `2` argument`3`; `4``5``6``7` argument`8` `9` expected."
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591)


maxitg (changes requested):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591)* status: all files reviewed, 3 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/messages.m, line 17 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm2w5L8EnXqqbomaae:-MQm2w5L8EnXqqbomaaf:b-sbm0jx) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L17)):*
> ```objc
> ];
> 
> (* General Messages *)
> ```

Why do we need this file? Wouldn't it be better to define these where they are used (like `setSubstitutionSystem.m`)? Some (maybe most) of these are only used in one file, and only need to be general because the function where they are used is called by other functions (like `subSubstitutionSystem`).
___
*[Kernel/messages.m, line 67 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm1xKi0Ic64_0aiZ34:-MQm1xKi0Ic64_0aiZ35:b-6zws71) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L67)):*
> ```objc
>   "The limit for the `2` is not supported for multiway systems.";
> 
> General::nonPropertyOpt =
> ```

Shouldn't we just use `General::nonopt` instead?
___
*[Kernel/utilities.m, line 56 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm5nM98ee0287rXxIA:-MQm5nM98ee0287rXxIB:bv12v1c) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/utilities.m#L56)):*
> ```objc
> 
> listToSentence[list_List] :=
>   Replace[ToString[#, InputForm] & /@ list, {{a__, b_} :> {Riffle[{a}, ", "], " or ", b}}];
> ```

Should we do a `StringJoin` to make the output of this function more clear?

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591)* status: all files reviewed, 3 unresolved discussions (waiting on @maxitg)

___
*[Kernel/messages.m, line 17 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm2w5L8EnXqqbomaae:-MQmLlzo-cIlbkWWq_Mi:b5x2kts) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L17)):*

Why do we need this file? Wouldn't it be better to define these where they are used (like `setSubstitutionSystem.m`)? Some (maybe most) of these are only used in one file, and only need to be general because the function where they are used is called by other functions (like `subSubstitutionSystem`).
</blockquote></details>

How about those messages, e.g. `invalidHypergraph`, where it is used for multiple symbols. In which file should I define the message?

___
*[Kernel/messages.m, line 67 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm1xKi0Ic64_0aiZ34:-MQmLdK3D4UPCCErW4OZ:b-uw4u4h) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L67)):*

Shouldn't we just use `General::nonopt` instead?
</blockquote></details>

Sure
```wl
WolframModel[{{1, 2}, {2, 3}} -> {{1, 3}}, Partition[Range[17], 2, 1], 4]["CausalGraph", {}, {EdgeStyle -> Red}]
```

___
*[Kernel/utilities.m, line 56 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm5nM98ee0287rXxIA:-MQmLkXQ1y05IMSF6luo:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/utilities.m#L56)):*

Should we do a `StringJoin` to make the output of this function more clear?
</blockquote></details>

Done.

maxitg (changes requested):
> Reviewed 3 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591)* status: all files reviewed, 1 unresolved discussion (waiting on @daneelsan)

___
*[Kernel/messages.m, line 17 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm2w5L8EnXqqbomaae:-MQmUesaDehjasLp4PZl:bvxhjz8) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L17)):*

How about those messages, e.g. `invalidHypergraph`, where it is used for multiple symbols. In which file should I define the message?


</blockquote></details>

I would argue in the same place as `hypergraphQ` is defined, which would be in `utilities.m`. I agree that it's a bit weird given that it's not actually used there, but it seems even weirder to keep some `General::` messages in `messages.m` but not the others.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @maxitg)

___
*[Kernel/messages.m, line 17 at r1](https://reviewable.io/reviews/maxitg/setreplace/591#-MQm2w5L8EnXqqbomaae:-MQmifPP2CTDLPvMCoAQ:bh7t6wr) ([raw file](https://github.com/maxitg/setreplace/blob/846b84ae0e6b5267ef862db210d2c82a2533395d/Kernel/messages.m#L17)):*

I would argue in the same place as `hypergraphQ` is defined, which would be in `utilities.m`. I agree that it's a bit weird given that it's not actually used there, but it seems even weirder to keep some `General::` messages in `messages.m` but not the others.
</blockquote></details>

That makes sense. Done.

maxitg (approved):
:lgtm:

> Reviewed 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/591#-:-MQrqp1g1Fv2gZ4dEm5-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #599, opened 2021-01-19
Basic Type System

maxitg:
## Changes

* Implements translations and raw properties of the *Yellowstone* type system system.
* Composite properties (properties depending on other properties) are not yet implemented.

## Comments

* This will mostly be used for `Multihistory`, however, the `Multihistory` itself is not yet implemented.
* There are currently no types, so `SetReplaceTypeConvert`, while implemented, cannot be used for anything.

## Examples

* Examples below do not work directly because they use fake types. In order to try them out, do something like
   ```wl
   AppendTo[$ContextPath, "SetReplace`PackageScope`"];
   ```
   and then evaluate the definitions (line 15 to 68) from `typeSystem.wlt`.
* If you do that and then evaluate
   ```wl
  AppendTo[$ContextPath, "SetReplace`A1$typeSystem`PackagePrivate`"];
  Graph[$typeGraph,
         VertexLabels -> {type[name_] :> name, property[name_] :> ToString[name] <> "[...]"},
         VertexStyle -> {type[_] -> Black, property[_] -> White},
         EdgeStyle -> {DirectedEdge[type[_], type[_]] -> Black,
                       DirectedEdge[type[_], property[_]] -> Lighter[Blue, 0.3],
                       DirectedEdge[property[_], property[_]] -> Gray}]
   ```
   you will get a type graph from the test file:

   image

* You will also be able to do things like these. Here, `"EvenInteger"` type is converted to `"HalfInteger"`:
   ```wl
   In[] := SetReplaceTypeConvert["HalfInteger"] @ evenInteger[4]
   Out[] = halfInteger[2]
   ```   

* Here, `multipliedHalf` property is used on a `"HalfInteger"` directly:
   ```wl
   In[] := multipliedHalf[5] @ halfInteger[3]
   Out[] = 15
   ```

* Here, `"HalfInteger"` is automatically converted to an `"EvenInteger"`, and then `description` property is called.
   ```wl
   In[] := description @ halfInteger[3]
   Out[] = "I am an integer 6."
   ```

* And here, `"EvenInteger"` is converted to `"HalfInteger"` first:
   ```wl
   In[] := multipliedHalf[evenInteger[4], 4]
   Out[] = 8
   ```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)


taliesinb (commented):
Looking nice. 

I'm half of the mind that we should just immediately go for the generic names here, rather than using the term `multihistory` everywhere, since this object system really will work with any kind of objects. In which case the word `property` should be replaced with the word `method`, and the word `multihistory` should be replaced with the word `type`, `class` (or perhaps `object`). 

I understand if you don't want to do that though -- we can always do that later!

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599#-:-MT97nzVElTv2WKgunyk:b-3kzyzp)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @daneelsan)

maxitg (commented):
I think you are right. I think we should keep `Multihistory` symbol though (the type system does not care what the head is as long as it answers to the `*type` package-scope function).

I can split this code into two files. One will be about generic types only (and will contain all the declaration functions). The other one will define `Multihistory`, the function that looks up its type, and some other things, such as generic `MakeBoxes`.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599#-:-MTCPZNbB7jBYo1cd9rS:b-x1bjzd)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @daneelsan)

maxitg (commented):
I changed this PR to a generic type system as @taliesinb suggested. `Multihistory` is still needed (for boxes if nothing else), but it will arrive in a different PR. This is now ready for review.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599#-:-MTgY1gb1tBj1A3B7lS5:bte8blc)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @daneelsan and @taliesinb)

maxitg:
I think there is a problem though. If the type system is generic, what do we do with `TypeConvert`? Surely, it is almost guaranteed to cause namespace collisions. And using packages without adding them to `$ContextPath` is horribly broken in WL at the moment.

taliesinb (approved):
Why not make make `TypeConvert` a `PackageScope` symbol? It's a bit low-level anyway for end-users to call directly. If they want to they can fully qualify it via PackageScope. 

> Reviewed 4 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599#-:-MUFRq2G3dnxgcm7963z:bcikspk)* status: all files reviewed, 5 unresolved discussions (waiting on @daneelsan and @maxitg)

*[a discussion](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFUXYj07eZ2Au-ngGc:-MUFUXYj07eZ2Au-ngGd:bod0euv) (no related file):*
I have one or two suggestions, but no showstoppers and I think it looks good as-is! Nice, look forward to trying it out.

---
___
*[Kernel/A1$typeSystem.m, line 45 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFSWf44poYEhB1-C-J:-MUFSWf44poYEhB1-C-K:b-besspm) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L45)):*
> ```objc
> 
> initializeTypeSystemTranslations[] := (
>   $typeGraph = Graph[DirectedEdge @@@ Rest /@ $translations];
> ```

Why not use the three-arg version of `DirectedEdge`, which lets you put an arbitrary tag on the edge which will be remembered, returned by VertexList etc.
___
*[Kernel/A1$typeSystem.m, line 76 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFTQceB2L9nrQm0wEi:-MUFTQceB2L9nrQm0wEj:byyfni6) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L76)):*
> ```objc
>   fromType = objectType[object];
>   If[!VertexQ[$typeGraph, type[#]], throw[Failure["unconvertibleType", <|"type" -> #|>]]] & /@ {fromType, toType};
>   path = FindShortestPath[$typeGraph, type[fromType], type[toType]];
> ```

You could check that there is only *one* shortest path here, by calling `FindPath` with the `{k}` fourth argument with `k = Length[path]` and ensuring it returns a singleton. 
___
*[Kernel/A1$typeSystem.m, line 80 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFTm7M-GnSx1sRm6LS:-MUFTm7M-GnSx1sRm6LT:bgpyavr) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L80)):*
> ```objc
>     throw[Failure["noConversionPath", <|"from" -> fromType, "to" -> toType|>]];
>   ];
>   edges = DirectedEdge @@@ Partition[path, 2, 1];
> ```

Ah, my comment about the tagged version of DirectedEdge wasn't useful, since FindShortestPath returns a list of vertices rather than a list of edges, so you'd have to pull out the information a different way anyway.
___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUFU0Fm-OCDPS58XzWi:b8gqoze) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*
> ```objc
> ];
> 
> (* declareCompositeProperty declares an implementation for a property that takes other properties as arguments. The
> ```

Do you really prefer the word 'property' to the word 'method'? 'method' has the benefit that it more explicitly conjures the thinking about protocols etc. if you do that rename then 'compositeProperty' becomes 'protocolMethod', for example.

maxitg (commented):
I think it's quite useful to users. For example, C++ objects are attached to a single kernel. If one wants to serialize them for `ParallelEvaluate` or just to store them somewhere, they will need to `TypeConvert` either to a multiset multihistory or, at least, to a serialized C++ instance.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599#-:-MUG-so3AyIFxXe4oMcp:bjmux05)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @daneelsan and @taliesinb)

*[a discussion](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFUXYj07eZ2Au-ngGc:-MUGBLCT7mhyDV3ZETX4:bv0cyie) (no related file):*

I have one or two suggestions, but no showstoppers and I think it looks good as-is! Nice, look forward to trying it out.
</blockquote></details>

Keep in mind though that the plan is to also add `Multihistory` symbol which would have a box form that would display data based on available properties, etc. We also need to add some introspection capabilities such as `TypeMethods`, etc.

---
___
*[Kernel/A1$typeSystem.m, line 45 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFSWf44poYEhB1-C-J:-MUG3H4U9F377i-VZvDp:buwz6dm) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L45)):*

Why not use the three-arg version of `DirectedEdge`, which lets you put an arbitrary tag on the edge which will be remembered, returned by VertexList etc.
</blockquote></details>

The problem is that it's not returned by `FindShortestPath`, so it's not clear how to reliably obtain the sequence of functions (doing `EdgeList @* Subgraph` might produce duplicate edges).
___
*[Kernel/A1$typeSystem.m, line 76 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFTQceB2L9nrQm0wEi:-MUG45Ez9vi0BKSzvIJo:b-85hjia) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L76)):*

You could check that there is only *one* shortest path here, by calling `FindPath` with the `{k}` fourth argument with `k = Length[path]` and ensuring it returns a singleton. 
</blockquote></details>

Do we want to do that though? Let's say, we have a situation like this:


I don't think we should disallow that. We might want to verify that different paths produce the same outputs, but we cannot do that for all cases (since, e.g., evolution video files would be different every time, and there are random properties as well).

We might want to introduce edge weights in the future to disambiguate, but since we don't have those yet, I don't think we should disallow multiple paths.
___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUG5Hwh49VEeC8BYGne:b-2zt64i) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Do you really prefer the word 'property' to the word 'method'? 'method' has the benefit that it more explicitly conjures the thinking about protocols etc. if you do that rename then 'compositeProperty' becomes 'protocolMethod', for example.
</blockquote></details>

Yes, I think you are right. The word 'property' assumes there are no other arguments, which is not correct here.

I'm not so sure about `protocolMethod` though because it sounds to me like a method that a protocol requires: https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#ID270

In this case, the method takes a "nameless protocol-type" as an argument.

maxitg:
I renamed `TypeConvert` -> `SetReplaceTypeConvert`. This has no possibility of collisions, and I think is more user-friendly than ``SetReplace`PackageScope`TypeConvert``.

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 2 of 4 files reviewed, 7 unresolved discussions (waiting on @maxitg and @taliesinb)

*[a discussion](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFUXYj07eZ2Au-ngGc:-MUUESys3k2M_hcH1q_t:b-j8hno7) (no related file):*

Keep in mind though that the plan is to also add `Multihistory` symbol which would have a box form that would display data based on available properties, etc. We also need to add some introspection capabilities such as `TypeMethods`, etc.
</blockquote></details>

This is a really nice workflow.

---
___
*[Documentation/TypeSystem.md, line 18 at r3](https://reviewable.io/reviews/maxitg/setreplace/599#-MUU3xb34eYsxLDAjew6:-MUU3xb4ArnATwkwJ-Lx:b-3ph67g) ([raw file](https://github.com/maxitg/setreplace/blob/18dd26cb8676405be1bfa3022902e76967f18ffe/Documentation/TypeSystem.md#L18)):*
> ```Markdown
> Most of the time, it is sufficient to rely on these automatic conversions, however, sometimes it might be useful to
> convert an object to a different type manually for persistence or optimization, in which case one can use the
> **`TypeConvert`** function:
> ```

`SetReplaceTypeConvert`
___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUU9rvj1CtaBJ2QHHe6:b-ia6xpd) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Yes, I think you are right. The word 'property' assumes there are no other arguments, which is not correct here.

I'm not so sure about `protocolMethod` though because it sounds to me like a method that a protocol requires: https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#ID270

In this case, the method takes a "nameless protocol-type" as an argument.
</blockquote></details>

In `DataStructure` this is called an operation. 

The reason, though, might be that WL actively avoids using oop nomenclature.
___
*[Kernel/A1$typeSystem.m, line 108 at r3](https://reviewable.io/reviews/maxitg/setreplace/599#-MUU8u5q3OomKJU6DEF9:-MUU8u5q3OomKJU6DEFA:b4xogb5) ([raw file](https://github.com/maxitg/setreplace/blob/18dd26cb8676405be1bfa3022902e76967f18ffe/Kernel/A1$typeSystem.m#L108)):*
> ```objc
>   $methodEvaluationFunctions = Association[Thread[newEdges -> (First /@ $rawMethods)]];
> 
>   defineDownValuesForMethod /@ Cases[VertexList[$typeGraph], method[name_] :> name, {1}];
> ```

This also works: `First /@ VertexList[$typeGraph, _method]`

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 2 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/TypeSystem.md, line 18 at r3](https://reviewable.io/reviews/maxitg/setreplace/599#-MUU3xb34eYsxLDAjew6:-MUUG6DHBvIV9B6rvzDG:boj1qgt) ([raw file](https://github.com/maxitg/setreplace/blob/18dd26cb8676405be1bfa3022902e76967f18ffe/Documentation/TypeSystem.md#L18)):*

`SetReplaceTypeConvert`
</blockquote></details>

Good catch. Done.
___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUUGQqI6VxoZjyBbmFH:bfpqrgr) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

In `DataStructure` this is called an operation. 

The reason, though, might be that WL actively avoids using oop nomenclature.
</blockquote></details>

Doesn't operation sound like it's going to mutate the object (which `DataStructure` operations do)?
___
*[Kernel/A1$typeSystem.m, line 108 at r3](https://reviewable.io/reviews/maxitg/setreplace/599#-MUU8u5q3OomKJU6DEF9:-MUUGPPx1T7moGN3cjwC:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/18dd26cb8676405be1bfa3022902e76967f18ffe/Kernel/A1$typeSystem.m#L108)):*

This also works: `First /@ VertexList[$typeGraph, _method]`
</blockquote></details>

Done.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUUHjHIChke0d7vIb1W:b8tsqe3) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Doesn't operation sound like it's going to mutate the object (which `DataStructure` operations do)?
</blockquote></details>

Some do, like `"Fold"`, some don't like `"Length"`.
But yes, the objects here are immutable, and operate has the feeling of mutability.
Still, methods also have the same connotation, right? Methods might be understood as just extracting data from the object (aka properties) OR mutating the object...

daneelsan (dismissed):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

*[a discussion](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFUXYj07eZ2Au-ngGc:-MUUJZQo3XT2FfSv68ei:bnfp4nl) (no related file):*

This is a really nice workflow.
</blockquote></details>

:lgtm:

---

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUUJi6E7hCMNuC5W5C0:bqqiapj) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Some do, like `"Fold"`, some don't like `"Length"`.
But yes, the objects here are immutable, and operate has the feeling of mutability.
Still, methods also have the same connotation, right? Methods might be understood as just extracting data from the object (aka properties) OR mutating the object...

</blockquote></details>

Maybe it's just me, but 'operation' sounds like mutation is its primary purpose (e.g., think of surgical operations).

Do you know a word that sounds immutable (like property) but allows for extra arguments?

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUUKnamAr_VPSCR5w9B:b-lhyx8r) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Maybe it's just me, but 'operation' sounds like mutation is its primary purpose (e.g., think of surgical operations).

Do you know a word that sounds immutable (like property) but allows for extra arguments?
</blockquote></details>

I don't think property assumes there are no other arguments. These extra arguments just specify a different way of extracting the data we want.

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUUPr9z9rvXXpJUSNbk:b-tyb6bs) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

I don't think property assumes there are no other arguments. These extra arguments just specify a different way of extracting the data we want.
</blockquote></details>

Hmm, maybe you are right. Any thoughts, @taliesinb? For example, there are properties in Swift, but they have a different meaning (no arguments).

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 108 at r2](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFU0Fm-OCDPS58XzWh:-MUnDix-Bk_OivjMB0zm:b-kerg4t) ([raw file](https://github.com/maxitg/setreplace/blob/0f7a8a04ae916eb25b9cf9761cc05a1f636c24e6/Kernel/A1$typeSystem.m#L108)):*

Hmm, maybe you are right. Any thoughts, @taliesinb? For example, there are properties in Swift, but they have a different meaning (no arguments).
</blockquote></details>

I renamed it back to the property. I'm quite convinced now that it's the right name to call it, given that it's non-mutating. We might also have operations in the future that would be mutating (e.g., for the C++ objects). However, the API for them should be different as automatic type conversions won't make sense then. We can use the word 'method' to describe both.

daneelsan (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/599)* status: 1 of 4 files reviewed, 7 unresolved discussions (waiting on @daneelsan and @taliesinb)

*[a discussion](https://reviewable.io/reviews/maxitg/setreplace/599#-MUFUXYj07eZ2Au-ngGc:-MUnR9x1AKIdlLIBp96N:b-r2wyus) (no related file):*

:lgtm:
</blockquote></details>

lgtm part 2

---

PR #606, opened 2021-02-04
Messages with Association arguments

maxitg:
## Changes

* Usually, messages in Wolfram Language take numbered arguments.
* However, `Failure` objects have an `Association` as a second argument, which fits perfectly with specifying the arguments of the failure with descriptive keys.
* Then, it would be convenient to be able to turn that `Failure` object directly into a message which would take the same keys as arguments.
* The built-in `Message` function, however, does not support passing `Association` arguments to the message strings.
* This PR implements `PackageScope` `declareMessage` and `message` that allow one to do that.
* `message` can also directly take `Failure` objects and create messages from them.

## Comments

* This will be particularly important for the `Multihistory` system where conversion functions and properties will need to report errors "far away" from where the messages will be produced.
* The idea is that these functions will throw `Failure` objects which will then be caught by a function from the `Multihistory` infrastructure which will then generate a message with an appropriate head depending on which symbol triggered the conversion/property computation.

## Examples

* Declare a message:
```wl
declareMessage[General::vertexNotFound, "Could not find vertex `vertex` in hypergraph `hypergraph`."];
```

* Generate a message:
```wl
message[hypergraphNeighborhood::vertexNotFound, <|"hypergraph" -> {{1, 2, 3}, {3, 4, 5}}, "vertex" -> 0|>];
```

image

* You can also create a message directly from a `Failure`:

image

* All this technology allows for a very simple implementation of message-throwing functions:

```wl
expr : HalfInteger[x_] := ModuleScope[
  result = Catch[halfInteger[x], _ ? FailureQ, message[HalfInteger, #, <|"expr" -> HoldForm[expr]|>] &];
  result /; !FailureQ[result]
];

halfInteger[x_ ? EvenQ] := x / 2;

declareMessage[HalfInteger::odd, "The number `number` in `expr` is odd."];
halfInteger[x_ ? OddQ] := throw[Failure["odd", <|"number" -> x|>]];
```

```wl
In[] := HalfInteger[4]
Out[] = 2
```

```wl
In[] := HalfInteger[3]
During evaluation of In[] := HalfInteger::odd: The number 3 in HalfInteger[3] is odd.
Out[] = HalfInteger[3]
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/606)


maxitg:
@taliesinb, @daneelsan, this is ready for review. I don't plan to implement anything else here.

maxitg:
@taliesinb, @daneelsan, could you give an ETA on when can you look at this?

daneelsan:
@maxitg Looking at it now

daneelsan:
It seems there is/was an attempt to make `Message` work with templates:
```wl
In[] := Message[MessageObject[<|"MessageSymbol" -> HoldForm[b], 
    "MessageTemplate" -> "Hello `world`!", 
    "MessageParameters" -> <|"world" -> 3|>|>]]; $Failed
Hello 3!

Out[] = $Failed
```

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/606)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/messages.m, line 15 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpbjd1CC3fiIbBwFRr:-MSpbjd1CC3fiIbBwFRs:b-3ccl80) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L15)):*
> ```objc
>   templateObject = StringTemplate[template];
>   argumentNames = First /@ Cases[templateObject[[1]], _TemplateSlot];
>   namesToIndices = Association[Thread[argumentNames -> Range[Length[argumentNames]]]];
> ```

`AssociationThread[argumentNames -> Range[Length[argumentNames]]]` :P 
___
*[Kernel/messages.m, line 28 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSplT3u5ePAw-gxMamq:-MSplT3u5ePAw-gxMamr:bpog1us) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L28)):*
> ```objc
> message[messageName_, args_ ? AssociationQ] := ModuleScope[
>   argumentsOrder =
>     Lookup[$messageSlotNames, Hold[messageName], $messageSlotNames[ReplacePart[Hold[messageName], {1, 1} -> General]]];
> ```

`$messageSlotNames[ReplacePart[Hold[messageName], {1, 1} -> General]]` 
At first, I didn't get it.

So this looks if there is a `sym::msg` first, then it looks at `General::msg`.
___
*[Kernel/messages.m, line 34 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpgoJBA4W8D0wGIfwe:-MSpgoJC7NVltr8rxCxD:b-uoscnm) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L34)):*
> ```objc
>     missingArgs = Complement[argumentsOrder, Keys[args]];
>     If[missingArgs =!= {},
>       Message[MessageName[message, "missingArgs"], HoldForm[messageName], missingArgs];
> ```

`message::missingArgs` like `message::messageNotFound` above.
___
*[Kernel/messages.m, line 36 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpraDo9d2CEMXgcG9p:-MSpraDo9d2CEMXgcG9q:b7hrya0) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L36)):*
> ```objc
>       Message[MessageName[message, "missingArgs"], HoldForm[messageName], missingArgs];
>     ,
>       Message[messageName, ##] & @@ Replace[argumentsOrder, args, 1];
> ```

TIL Replace works with Associations as the 2nd argument.

daneelsan (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/606)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Tests/message.wlt, line 19 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpy62NBgMGXh-Dtr4y:-MSpy62OC9BxV-eY-Msr:b-ymkduq) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Tests/message.wlt#L19)):*
> ```Mathematica
> 
>       declareMessage[HalfInteger::odd, "The number `number` is odd."];
>       halfInteger[x_ ? OddQ] := throw[Failure["odd", <|"number" -> x|>]];
> ```

Why `Failure["odd", <|"number" -> x|>]` over `Failure["odd", <|"MessageParameters" -> <|"number" -> x|>|>]`
which I think is the convention?

maxitg (commented):
It's formatted differently though and does not show the head which would not work for us.

By the way, I should point out that there are `ThrowFailure` and `CatchFailureAsMessage` functions in *GeneralUtilities*, but I think they address different problems as they don't seem to support associations (and, unless I'm missing something, you won't be able to pass `expr` to them at the last moment).

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/606#-:-MSq-h1kEcKnU4qRlJKB:b-hbo2jh)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[Kernel/messages.m, line 15 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpbjd1CC3fiIbBwFRr:-MSq-uBA9eBdBuC54SH7:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L15)):*

`AssociationThread[argumentNames -> Range[Length[argumentNames]]]` :P 
</blockquote></details>

Done.
___
*[Kernel/messages.m, line 28 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSplT3u5ePAw-gxMamq:-MSq0KEbCdBNmwCupbGf:byq9z58) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L28)):*

`$messageSlotNames[ReplacePart[Hold[messageName], {1, 1} -> General]]` 
At first, I didn't get it.

So this looks if there is a `sym::msg` first, then it looks at `General::msg`.
</blockquote></details>

Added a comment to clarify.
___
*[Kernel/messages.m, line 34 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpgoJBA4W8D0wGIfwe:-MSq0QJdDdUnNYSUmsp4:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Kernel/messages.m#L34)):*

`message::missingArgs` like `message::messageNotFound` above.
</blockquote></details>

Done.
___
*[Tests/message.wlt, line 19 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpy62NBgMGXh-Dtr4y:-MSq0UbX42wQappHJ6Fr:boly9h0) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Tests/message.wlt#L19)):*

Why `Failure["odd", <|"number" -> x|>]` over `Failure["odd", <|"MessageParameters" -> <|"number" -> x|>|>]`
which I think is the convention?
</blockquote></details>

Mainly because it's shorter. I believe the convention would actually be

```wl
Failure["odd", <|"MessageTempalte" -> HalfInteger::odd, "MessageParameters" -> <|"number" -> x|>|>]
```

which is a lot longer and does not contain any additional information because:

* `Failure` does not know the symbol name yet.
* The only thing it would generally contain would be message parameters. If it needs to contain additional information, that's no problem as well, because extra parameters are just ignored.
* We cannot actually use `Failure` to generate the message; we need to do that manually anyway. The only case where using the convention will make a difference is for displaying it in the Front End. However, that should never happen because we are supposed to catch these before they escape.

We could perhaps define a new PackageScope symbol that would generate a conventionally-formatted message, e.g.,

```wl
setReplaceFailure[name_, params_] :=
  Failure[name, <|"MessageTempalte" -> MessageName[General, name], "MessageParameters" -> params|>]
```

but even that would be weird because we are not actually supposed to know the symbol name at this point, and always using `General` (like above) might not be correct. Plus, I don't really see how we would benefit from using the convention.

daneelsan (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/606)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @daneelsan, @maxitg, and @taliesinb)

___
*[Tests/message.wlt, line 19 at r2](https://reviewable.io/reviews/maxitg/setreplace/606#-MSpy62NBgMGXh-Dtr4y:-MSrWN1J70iCI-ls6_aX:b69qhvq) ([raw file](https://github.com/maxitg/setreplace/blob/1a69b1753c7816cd68bc5dc351bcd85e4ded3fd3/Tests/message.wlt#L19)):*

Mainly because it's shorter. I believe the convention would actually be

```wl
Failure["odd", <|"MessageTempalte" -> HalfInteger::odd, "MessageParameters" -> <|"number" -> x|>|>]
```

which is a lot longer and does not contain any additional information because:

* `Failure` does not know the symbol name yet.
* The only thing it would generally contain would be message parameters. If it needs to contain additional information, that's no problem as well, because extra parameters are just ignored.
* We cannot actually use `Failure` to generate the message; we need to do that manually anyway. The only case where using the convention will make a difference is for displaying it in the Front End. However, that should never happen because we are supposed to catch these before they escape.

We could perhaps define a new PackageScope symbol that would generate a conventionally-formatted message, e.g.,

```wl
setReplaceFailure[name_, params_] :=
  Failure[name, <|"MessageTempalte" -> MessageName[General, name], "MessageParameters" -> params|>]
```

but even that would be weird because we are not actually supposed to know the symbol name at this point, and always using `General` (like above) might not be correct. Plus, I don't really see how we would benefit from using the convention.
</blockquote></details>

I see. If we do not plan to return `Failure` objects then using MessageTemplate/MessageParameters is not worth it.

PR #603, opened 2021-01-27
Add pattern names in SyntaxInformation and missing usage, syntax information and syntax autocompletion

maxitg:
## Changes

* In the Mathematica options inspector, there is an option called `"HighlightMissingArgumentsWithTemplate"`.
* If you turn it on, in case of missing arguments, instead of displaying a single red arrow, you will get a helpful hint with argument names which can serve as a super fast lightweight documentation (like you would get in, e.g., Xcode after autocompleting a function):

HighlightMissingArgumentsWithTemplate

* The names of the arguments are taken from pattern names in `SyntaxInformation`.
* This PR adds names to all `SyntaxInformation` except for the built-in `RulePlot` thus enabling this feature for all *SetReplace* symbols.

* This also adds missing usage messages, syntax information, and syntax autocompletion.
* Expands argument names in usage messages to be descriptive so that using autocompleted templates are more informative.

## Comments

* `Null` is due to `OptionsPattern[]`. It can be changed to, say, `opts` if that pattern is named as `opts : OptionsPattern[]`. However, if one does that, some optional arguments will always be displayed in the template, which is, of course, unacceptable. I think we'll have to wait until this weed is fixed upstream:

image

* Autocompletion is generally disabled for option names. The problem is that I'm not aware of a way to enable autocompletion for multiple arguments at a time. In order to solve this, we need to use symbols instead of strings for options (#604), and we need to figure out how to autocomplete the values for enum-like options (#605).

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/603)

PR #610, opened 2021-02-06
Causal Density Dimension

SantiagoGiner:
## Changes

* Create utility function `GraphDimension`, which, given a causal graph, a method for estimation, and a pair of vertices in the graph, will return an estimate for the dimension of the graph.

* Currently, the method supported is `FlatCausalDiamondRelationProbability`, which is based on the Myrheim-Meyer dimension estimation algorithm. This method comes from Causal Set Theory and works by counting the number of causal relations and events in an interval of the graph defined as the intersection of the past of one event with the future of another (making use of `AcyclicGraphTake`). A ratio between said number of events and number of relations is a function solely of the dimension of the manifold into which the causal graph embeds. Thus, GraphDimension solves this expression of dimension (which can be found in *GraphDimension.md*), returning the estimation value.

* Modified `AcyclicGraphTake` so as to generalize the error messages it displays.

## Comments
* In the FlatCausalDiamondRelationProbability method, if the events (vertices) provided are causally disconnected or are the same event, the function will return infinity.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610)


maxitg (changes requested):
Looking forward to using this! I have some comments, mostly about formatting and documentation but nothing major. @taliesinb, I think it would be helpful if you could take a look as well, particularly if you have some ideas about the API.

> Reviewed 4 of 7 files at r1, 2 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610#-:-MSueu1t7_S_zQjH7fjn:b828yr9)* status: all files reviewed, 20 unresolved discussions (waiting on @SantiagoGiner and @taliesinb)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 1 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSsw20v3bIF0IhGL94F:-MSsw20wBBynDLFv6QJr:bnyqo8g) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L1)):*
> ```Markdown
> ###### [Symbols and Functions](/README.md#symbols-and-functions) > Utility Functions >
> ```

Add a link to this file from the table of contents in `README.md`.
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 9 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSswlbM4CXklwvtEh7q:-MSswlbM4CXklwvtEh7r:b-lt5yx8) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L9)):*
> ```Markdown
> 
> ## Dimension estimation methods
> ### `FlatCausalDiamondRelationProbability`:
> ```

I think it'd be better to not format this as code, add the quotes and remove the colon at the end, to be consistent with, e.g., [HypergraphToGraph](https://github.com/maxitg/SetReplace/blob/GraphDimension/Documentation/SymbolsAndFunctions/UtilityFunctions/HypergraphToGraph.md).
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 11 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSt1xZq2PmjbSZTY9p9:-MSt1xZq2PmjbSZTY9pA:b-lr1gdl) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L11)):*
> ```Markdown
> ### `FlatCausalDiamondRelationProbability`:
> Uses the Myrheim-Meyer dimension estimation algorithm. This method is
> based on Causal Set Theory, an approach to quantum gravity which, much like the Wolfram Model, describes spacetime
> ```

It would be good to link "Causal Set Theory" either to its Wikipedia page or to a paper about it (up to you).
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 13 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSsx7s8Bwlea8DoXgFL:-MSsx7s8Bwlea8DoXgFM:b-v6pz18) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L13)):*
> ```Markdown
> based on Causal Set Theory, an approach to quantum gravity which, much like the Wolfram Model, describes spacetime
> not as a continuum but as a large collection of discrete points endowed with a causal partial order of the form 
> x \[Precedes] y, i.e. "x is to the past of y." A causal set then contains a finite set of causal relations following
> ```

`\[Precedes]` would not render in html. You should use the Unicode character ≺ instead.
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 14 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSszKIi1pKVZYugPP-v:-MSszKIi1pKVZYugPP-w:bdlq7do) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L14)):*
> ```Markdown
> not as a continuum but as a large collection of discrete points endowed with a causal partial order of the form 
> x \[Precedes] y, i.e. "x is to the past of y." A causal set then contains a finite set of causal relations following
> this partial order in space. Starting from d-dimensional flat (Minkowski) space, the Myrheim-Meyer dimension 
> ```

space -> spacetime?
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 16 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSszmQQ1m5R6g5n-1NG:-MSszmQQ1m5R6g5n-1NH:b-use1yy) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L16)):*
> ```Markdown
> this partial order in space. Starting from d-dimensional flat (Minkowski) space, the Myrheim-Meyer dimension 
> estimation works essentially as a box-counting method, whereby one can consider two points *p* and *q* in that space 
> that are causally connected.The Alexandrov Interval or causal diamond *A[p, q]* is then defined to be the intersection
> ```

A space character is missing before "The Alexandrov Interval".
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 17 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSt-G99-6e4gTN_XJ_4:-MSt-G99-6e4gTN_XJ_5:b-9luimr) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L17)):*
> ```Markdown
> estimation works essentially as a box-counting method, whereby one can consider two points *p* and *q* in that space 
> that are causally connected.The Alexandrov Interval or causal diamond *A[p, q]* is then defined to be the intersection
> of the future of the event *p* with the past of the event *q*. In this interval, define C<sub>2</sub> to be the 
> ```

For consistency, we need to either make all variables italic including x and y before and C<sub>1</sub> and C<sub>2</sub> after or none of them.
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 20 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSt-VhxEq-JJ9WCo_kA:-MSt-VhxEq-JJ9WCo_kB:b-oi319) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L20)):*
> ```Markdown
> average number of causal relations, i.e. the number of pairs x \[Precedes] y between *p* and *q*, and define 
> C<sub>1</sub> to be the average number of events (i.e. vertices) in the interval. It is found that a suitable ratio 
> between C<sub>2</sub> and C<sub>1</sub> is purely dependent on the dimensionality d into which the causal set embeds:
> ```

I think saying that it's "purely dependent" is a bit misleading because it also depends on curvature. I think it would make sense to reemphasize, "It is found that in a flat spacetime a suitable ratio ... dependent on the dimensionality ...".
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 20 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSt25Fc-ifflyndQqof:-MSt25Fc-ifflyndQqog:b19k2qm) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L20)):*
> ```Markdown
> average number of causal relations, i.e. the number of pairs x \[Precedes] y between *p* and *q*, and define 
> C<sub>1</sub> to be the average number of events (i.e. vertices) in the interval. It is found that a suitable ratio 
> between C<sub>2</sub> and C<sub>1</sub> is purely dependent on the dimensionality d into which the causal set embeds:
> ```

We need either an explicit derivation of this or a reference to a paper that derives it. I think Theorem III.2 from https://dspace.mit.edu/handle/1721.1/14328 is a good one to refer (note, the spacetime there is *1 + d* dimensional so *d* is defined differently).
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 37 at r1](https://reviewable.io/reviews/maxitg/setreplace/610#-MSsfnAqB4wLodNAB_1i:-MSsfnArCGQmf2x45Lfb:b-lnxnln) ([raw file](https://github.com/maxitg/setreplace/blob/57f276de5dc4bc79c926cc3887a6851811736ccc/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L37)):*
> ```Markdown
> ```wl
> In[] := graph = BlockRandom[
>   IndexGraph[ResourceFunction["FlatSpacetimeSprinkling"][2, 50, "CausalGraph"], VertexLabels -> Automatic],
> ```

I would avoid using resource functions if possible. It's not as bad using them in documentation but they can still change without any warning and require Internet connection.

In this case, I suggest to just put the code for sprinkling here since it's very simple (plus, the code below creates a causal diamond already, so there is no need to manually pick vertices):

```wl
sprinkledGraph = BlockRandom[Module[{coordinates, causalSet},
  coordinates = Sort[Join[{{0, 0}, {1, 1}}, RandomReal[1, {48, 2}]] . RotationMatrix[Pi / 4]];
  causalSet = RelationGraph[Abs[#2[[2]] - #1[[2]]] < #2[[1]] - #1[[1]] &, coordinates];
  TransitiveReductionGraph @
    IndexGraph[causalSet, VertexCoordinates -> ({#2, 1 - #1} & @@@ coordinates), VertexLabels -> Automatic]
], RandomSeeding -> 0]
```

It would be even better to add a sprinkling function to *SetReplace*, but not in this PR of course.
___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md, line 35 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuWTnV4rZWmConRhXn:-MSuWTnV4rZWmConRhXo:bb4bvdt) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Documentation/SymbolsAndFunctions/UtilityFunctions/GraphDimension.md#L35)):*
> ```Markdown
> relationship in order to determine the dimension *d*. It is worth noting that when the events *p* and *q* defining
> the causal diamond are either causally disconnected or are the same event, the left-hand side of the above expression
> will diverge, meaning that the dimensionality estimation will return infinity.
> ```

I think the left-hand side turns to zero in this case rather than diverges.
___
*[Kernel/GraphDimension.m, line 19 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuX7Db7ehk0ph4amKW:-MSuX7Db7ehk0ph4amKX:b-pycy) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Kernel/GraphDimension.m#L19)):*
> ```objc
> (* Documentation *)
> SetUsage @ "
> GraphDimension[gr$, mth$, vrts$] gives an estimate of the dimension of a subgraph of the graph gr$ \
> ```

Please use full names here, e.g., `graph$`. These names are used in templates for autocompletion, and full names make these templates more clear (the other functions are changed in #603). Plus, it's better for them to be consistent with the names in `SyntaxInformation`.
___
*[Kernel/GraphDimension.m, line 44 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuYlB1AqYVXQDBMsnL:-MSuYlB1AqYVXQDBMsnM:b-rzqndw) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Kernel/GraphDimension.m#L44)):*
> ```objc
> 
> (* Normal form *)
> graphDimension[causalGraph_, "FlatCausalDiamondRelationProbability", vertices_] := ModuleScope[
> ```

`ModuleScope` here does not actually do anything. `ModuleScope` automatically detects assignments, e.g., `x = ...`, and creates a `Module` with corresponding variables in the first argument. However, in this case, `d` is not assigned to anything in this function, and so `ModuleScope` would not pick it up. You need to use `Module` and specify `d` explicitly in the first argument.
___
*[Kernel/GraphDimension.m, line 45 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuZMhh1NnosmaGJbuA:-MSuZLxM4hgCD-3yV7Qj:b75k55x) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Kernel/GraphDimension.m#L45)):*
> ```objc
> (* Normal form *)
> graphDimension[causalGraph_, "FlatCausalDiamondRelationProbability", vertices_] := ModuleScope[
>     With[{
> ```

Nit: Use 2 spaces for indentation instead of 4.
___
*[Kernel/GraphDimension.m, line 47 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuZWpr1or0tzNgc8rY:-MSuZWpsEDHQN1FpFo2O:b-13doc) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Kernel/GraphDimension.m#L47)):*
> ```objc
>     With[{
>         diamond = TransitiveClosureGraph[acyclicGraphTake[causalGraph, vertices]]},
>         If[EmptyGraphQ[diamond], Infinity,
> ```

Nit: Put `Infinity` on the next line so that there is "symmetry" between `If` branches. Also, to make it clear that there is an *else* instead of just a compound expression, we have a convention for arranging `If` like this:

```wl
If[EmptyGraphQ[diamond],
  Infinity
,
  Replace[d, FindRoot[...]]
]
```

Note that `,` does not have indentation with respect to `If` similar how an `else` would not have in a language such as C++.
___
*[Kernel/GraphDimension.m, line 49 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuaT2o9ezyz0Ar4zOB:-MSuaT2o9ezyz0Ar4zOC:b-qcczht) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Kernel/GraphDimension.m#L49)):*
> ```objc
>         If[EmptyGraphQ[diamond], Infinity,
>           Replace[d, FindRoot[
>             {EdgeCount[diamond]/((VertexCount[diamond])^2) == (Gamma[d + 1]*Gamma[d/2])/(4 Gamma[3 d/2])},
> ```

Nit: Add spaces between operators, i.e.,

```wl
{EdgeCount[diamond] / ((VertexCount[diamond])^2) == (Gamma[d + 1] * Gamma[d / 2]) / (4 Gamma[3 d / 2])}
```

By the way, you can use

```wl
<< CodeFormatter`;
CodeFormat[
  "{EdgeCount[diamond]/((VertexCount[diamond])^2) == (Gamma[d + 1]*Gamma[d/2])/(4 Gamma[3 d/2])}",
  "LineWidth" -> 120]
```

to do that automatically, although it does not work well enough to our standards for large blocks of code.
___
*[Tests/GraphDimension.wlt, line 13 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSubbh44BhmKB6YmvSr:-MSubbh44BhmKB6YmvSs:b-wo8dzv) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Tests/GraphDimension.wlt#L13)):*
> ```Mathematica
>         undirectedGraph = Graph[{1 <-> 2, 2 <-> 3}],
>         cyclicGraph = Graph[{1 -> 2, 2 -> 1}],
>         $flatRelationProb = "FlatCausalDiamondRelationProbability"
> ```

Nit: I think for code clarity purposes, it would be better to use the full name below.

If you really want to shorten it, I would instead define something like this here:

```wl
dimensionFunction = GraphDimension[#1, "FlatCausalDiamondRelationProbability", #2] &
```

This way, you'll still be using full names, and it's even shorter than before.
___
*[Tests/GraphDimension.wlt, line 16 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSucKPL0dqCY72vTlqp:-MSucKPL0dqCY72vTlqq:bf4citz) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Tests/GraphDimension.wlt#L16)):*
> ```Mathematica
>       },
>       {
>         (* Verification tests *)
> ```

Nit: Add an empty line after "Verification tests" so that it does not look like it's only related to the first one.
___
*[Tests/GraphDimension.wlt, line 27 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSucbM501oqh5ilA68E:-MSucbM501oqh5ilA68F:b-kzgses) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Tests/GraphDimension.wlt#L27)):*
> ```Mathematica
>             VerificationTest[
>               GraphDimension[DirectedGraph[PathGraph[Range[100]], "Acyclic"], $flatRelationProb, {1, 90}],
>               Replace[d, FindRoot[(Gamma[d + 1]*Gamma[d/2])/(4 Gamma[3 d/2]) == Binomial[90, 2]/(90^2), {d, 1, 5}]],
> ```

Nit: Same here with respect to spacing.
___
*[Tests/GraphDimension.wlt, line 34 at r3](https://reviewable.io/reviews/maxitg/setreplace/610#-MSuciWuDyn62CD7qOSM:-MSuciWuDyn62CD7qOSN:b4twv7w) ([raw file](https://github.com/maxitg/setreplace/blob/b2084e48a96354d6c7dbb027720878de8c9c1a63/Tests/GraphDimension.wlt#L34)):*
> ```Mathematica
>               GraphDimension[IndexGraph[Graph[Join["i" -> # & /@ Range[1000], # -> "o" & /@ Range[1000]]]],
>                 "FlatCausalDiamondRelationProbability", {1, 1002}], Replace[d, FindRoot[
>                   (Gamma[d + 1]*Gamma[d/2])/(4 Gamma[3 d/2]) == (2*(1002 - 2) + 1)/(1002^2), {d,1, 10}]],
> ```

Nit: Same with respect to spacing, also add a space after `d,` in `{d,1, 10}`.

maxitg (approved):
> Reviewed 2 of 5 files at r4, 2 of 2 files at r5, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @taliesinb)

SantiagoGiner:
Finally passed lint tests!

taliesinb (commented):
Looks very nice functionality to have! 

I'm wondering about the name, for a couple reasons:
1. the name `GraphDimension` suggests it is the dimension of the whole graph, rather than a 'local dimension' wrt two vertices.
2. other dimension estimators, like ball volume growth rate, depend on a *single* vertex, and have parameters, like what ball radius to use. so I don't see how that is directly compatible with the argument structure of this particular function.
3. if the name Myrheim-Meyer dimension is the standard name, why not use that (as "MyrheimMeyer") instead of "FlatCausalDiamondRelationProbability", which is less googleable. Other parts of Mathematica usually use the literature names for methods, rather than CamcelCased, truncated descriptions of how the methods work. 

I think one way to address all these concerns is not not pretend that we have enough data to create a general API yet, and so expose this as its own function, e.g. `MyrheimMeyerGraphDimension[graph, vertex1, vertex2]`. We can always factor later when we have more examples of graph dimension implemented, and a general design becomes clear (if there is one). A possible analogy here are also the various centrality graph functions in Mathematica, which are *not* all shoe-horned into one function.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610#-:-MT94ubl9rVmNhhcnK1J:b-pfobsx)* status: 6 of 8 files reviewed, all discussions resolved (waiting on @maxitg)

maxitg (commented):
1. I think the plan in the future is to allow it to take entire graphs as arguments as long as they are causal diamonds already (it's undefined otherwise). But I see your point in that most dimension estimators operate on subgraphs (sometimes multiple subgraphs at a time).
2. So, is the idea that we should not have arguments, structure of which depends on other arguments? E.g.,
   ```wl
   GraphDimension[graph, "FlatCausalDiamondRelationProbability", {v1, v2}]
   GraphDimension[hypergraph, "BallGrowth", {v0, radius}]
   ```
   I think the third argument here has the same meaning (it specifies a region in the graph), however, not the same structure. I didn't think of this as a problem before, but maybe we should use it as an additional design constraint.
3. Because it's a name that has no information about what is being done. In particular someone might use it without realizing that it assumes flat spacetime. In addition, there is actually a family of these estimators. E.g., both "flat" and "causal diamond" can be changed to something yielding an estimator that is similar to "MyrheimMeyer" but can no longer be called that. (Also, "Myrheim-Meyer" and a link to its derivation is mentioned in the documentation of course.)

With that said, I'm not against having a separate function for Myrheim-Meyer-like dimension estimators, but I think we need a more descriptive name for it than `MyrheimMeyerGraphDimension`.

> Reviewed 2 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610#-:-MTBZ7AFAHiHEfJQjmak:b-iiwvxc)* status: :shipit: complete! all files reviewed, all discussions resolved

SantiagoGiner:
Should we then still use the name `GraphDimension` and change the usage to, e.g.:

```wl
SetUsage @ "
GraphDimension[graph$, method$, subregion$] gives an estimate of the dimension of a subregion of the graph graph$ \
specified by subregion$, using the method method$.
";
```

And then, depending on the method, modify the error messages such that, for instance, in the case of the "Ball Radius" method, the message would read something like:
```wl
The argument at postion 3 of GraphDimension[graph, "BallRadius", {}] should be a list of a vertex and a radius.
```

maxitg (commented):
@santiagoginer, if we are going to keep calling it `GraphDimension`, then I think yes, we should do that.

@taliesinb, I think our disagreement here boils down to the following design principles:

1. ***Descriptive names***. I think we can come up with a better name than `"MyrheimMeyer"` (not necessary `"FlatCausalDiamondRelationProbability"`, there may well be better options here).
2. ***Avoiding hidden non-trivial defaults***. `"MyrheimMeyer"` contains within itself an assumption that the spacetime is flat, which needs to be explicit in the code. (Perhaps we can use `"FlatCausalRelationProbability"` because the function would not let one to use it for non-causal-diamonds anyway.)
3. ***Avoiding defactorization***. This could happen if we have, e.g., `{"Flat", "DeSitter"}` and `{"CausalRelationProbability", "MidpointScaling"}`.

Perhaps, we can do something like this?

```wl
CausalGraphDimension[graph, "FlatSpacetimeCausalRelationDensity", {v1, v2}]
```

Then, we'll have a different symbol for computing dimensions of states (hypergraphs). This will make the third arguments more consistent as, say, midpoint-scaling dimension requires a causal diamond as well.

We can even call it `CausalDiamondDimension` to address Tali's first point.

@santiagoginer, @taliesinb, what do you think?

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610#-:-MTCQwYND-jxz0yCsXAu:b-fc4jvt)* status: :shipit: complete! all files reviewed, all discussions resolved

taliesinb:
I think I'm happy with the name CausalGraphDimension. What do we think about using the word Estimate here? EstimateCausalGraphDimension? Aren't these things generally called "dimension estimators"?

As for the design principle: yes, the enum value of one argument shouldn't change the *meaning* of another positional argument. That's definitely a no-no. 

The way around it is to combine the two arguments into a single data-carrying argument (like how eg NetEncoder works), so that the "spooky action at a distance" is now contained within one argument:

```
CausalGraphDimension[graph, {"FlatCausalDiamondWhatever", v1, v2}]

CausalGraphDimension[graph, {"BallVolume", v, r}]
```

and then you just list the possible enum tuples and their argument structure in a table in the doc page (and in SetUsage too if you want to use the markdown-style table feature there)

SantiagoGiner:
I quite like the single data-carrying argument because it conveys that the different specifications of the region in the graph, i.e. either the two vertices or the vertex and radius, are related to the method itself. And I also agree with the name `CausalGraphDimension`.

So if @maxitg also agrees, I can make this change.

maxitg:
@taliesinb, @SantiagoGiner, do we expect other argument structures for `CausalGraphDimension`? `"BallVolume"` is for spatial hypergraphs, which would be a different function in this case.

Also, some estimators might have other arguments such as causal chain length, which is not necessarily a problem, but it might be awkward to put them in the same list with vertices.

One very simple thing we can do to avoid this problem entirely until we have more information is that we can require causal diamond as an input to this function, and thus require users to use `AcyclicGraphTake`, e.g.,

```wl
CausalGraphDimension[AcyclicGraphTake[graph, {v1, v2}], "FlatSpacetimeCausalRelationDensity"]
```

Also, I'm not against `Estimate`, although it seems a bit redundant since that's the only thing we can do.

maxitg:
The more I think about it, the more I realize that this is really quite complicated which makes me agree more with @taliesinb's original suggestion of trying to keep this as simple as possible.

Maybe it should just be something like this?

```wl
CausalChainDensityDimension @ AcyclicGraphTake[graph, v1, v2]
```

or even

```wl
CausalDensityDimension @ AcyclicGraphTake[graph, v1, v2]
```

I realize it does not say "flat" anywhere, but maybe it's not such a bad default. We can then extend it with additional arguments as needed:

```wl
CausalDensityDimension[curvature, chainLength] @ AcyclicGraphTake[graph, v1, v2]
```

And if we manage to make it work for non-causal-diamonds, it can naturally be extended to take any graphs as arguments:

```wl
CausalDensityDimension[0., 2] @ graph
```

`CausalDensity` (`CausalChainDensity`) refers to Myrheim-Meyer, so, e.g., Midpoint-scaling will get a different function. And we pretty much know already that the only parameters to Myrheim-Meyer are the chain length and the manifold which causal chain density is compared against (the sane cases of which are probably limited to a single number, i.e., the curvature, corresponding to flat, de Sitter and anti-de Sitter spaces). If we start comparing to more complicated manifold, we'll probably need a whole new setup anyway.

SantiagoGiner:
I don't know if I really like the fact that the user would have to explicitly use `AcyclicGraphTake`. I agree with the points that the format we have now doesn't really generalize well for other cases. Thus, why don't we more simply create one specific standalone function that simply uses the Myerheim-Meyer method for flat spacetimes, that is, something like:
```wl
FlatCausalDiamondDimension[graph, {v1, v2}]
```
That way we can have something specific and which shares the format of `AcyclicGraphTake`.

maxitg:
@SantiagoGiner, but then how will you call the function that computes, say, a midpoint scaling dimension instead?

Plus, with this naming convention, we might end up getting a whole grid of functions (API defactorization) such as `FlatCausalDiamondDimenion`, `DeSitterCausalDiamondDimension`, `FlatLightConeDimension`, `DeSitterLightConeDimension`, etc.

But I think it would make sense to do something like this:

```wl
CausalDensityDimension[graph, {v1, v2}]
```

which can then be extended as needed into

```wl
CausalDensityDimension[0., 2][graph, {v1, v2}]
```

and

```wl
CausalDensityDimension @ graph
```

And, by the way, I think it would be nice to support the last syntax already in case `graph` is already a causal diamond (in a separate PR).

SantiagoGiner:
@maxitg I very much like that convention. It is specific yet also generalizable, as you say. If we are settled on it I can then change the name of the function.

And yes, I definitely think that it would be good to support the latter syntax—I can submit a separate PR later.

maxitg:
@SantiagoGiner, I think so. It would be nice to get @taliesinb's opinion, but he might not be available until Sunday, and I think we essentially settled on his original design just with a different name.

SantiagoGiner:
@maxitg Yes, I agree. Also, should `CausalDensityDimension` also accept methods as `GraphDimension` does? I.e. how would we differentiate between each method? And should we continue to use `FlatCausalDiamondRelationProbability` for the flat Myrheim-Meyer estimation?

maxitg:
“CausalDensity” is the method name. The other one would be, e.g., `MidpointScalingDimension`.

SantiagoGiner:
Just completed the appropriate changes!

maxitg (approved):
Very nice!

> Reviewed 5 of 6 files at r11.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610#-:-MTcZB3sDBRxoyngWE52:b-j7qk9t)* status: all files reviewed, 1 unresolved discussion (waiting on @SantiagoGiner)

___
*[Documentation/SymbolsAndFunctions/UtilityFunctions/CausalDensityDimension.md, line 8 at r11](https://reviewable.io/reviews/maxitg/setreplace/610#-MTcZJkk3Gf9IJkko3YR:-MTcZJkk3Gf9IJkko3YS:bc8e5rg) ([raw file](https://github.com/maxitg/setreplace/blob/705a21bd7cc29de3a3efd423c5431249b26c2b21/Documentation/SymbolsAndFunctions/UtilityFunctions/CausalDensityDimension.md#L8)):*
> ```Markdown
> dimension estimation algorithm.
> 
> ## Myrheim Meyer Dimension Estimation
> ```

Nit: should there be a hyphen between Myrheim and Meyer?

SantiagoGiner:
Yes, that's true. Good catch!

maxitg (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/610)* status: all files reviewed, 1 unresolved discussion (waiting on @SantiagoGiner)

PR #612, opened 2021-02-10
Load-order-fixing file names

maxitg:
## Changes

* Adds `A0$`, `A1$` and `A2$` prefixes to files that should be loaded before others.
* Without this, things like `declareMessage` and `declareMultihistoryTranslation` won't work because they need to loaded before other files.

## Comments

* I know this is ugly but until we have a custom package-loading system (which @taliesinb has been suggesting), I don't think there is a better option. Certainly, this is better than fiddling with file names.
* `Multihistory` will be `A2$` once implemented.
* I will also move loading of C++ functions to a separate file which will be in `A0$`, as I'm going to implement functions that will be used outside of `setSubstitutionSystem$cpp.m`.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/612)


taliesinb (approved):
I'm wondering if `_` would be a better character than `$`, since it doesn't have any shell-expansion meaning (for example). But will pre-approve to speed things up -- your call. 

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/612#-:-MT92C7R5c8q5Oae7I4K:b-6lx8nl)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @taliesinb)

maxitg (commented):
I would be, but we cannot use it for the same reason we cannot start filenames with a number. The filename becomes part of the context, ``SetReplace`filename`PackagePrivate` ``, and context names cannot contain `_`.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/612#-:-MTBO2U46v2UwZsE0iAB:b-5lyvot)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @taliesinb)

PR #613, opened 2021-02-10
Separate file for libSetReplace loading

maxitg:
## Changes

* Moves the code that loads *libSetReplace* to a separate file.
* This will make it more natural to use *libSetReplace* outside of `setSubstitutionSystem$cpp`, for example, for path searching on directed hypergraphs which is needed for the `Multihistory` system.

## Comments

* Also renames `$cpp$function` to `cpp$function` to be more consistent with our conventions.
* Specific functions are still loaded where they are used with `importLibSetReplaceFunction` which is defined in `A01$libSetReplace.m`.
* The rough conventions for load-order naming:
  * `A0*` -> Loading code that is not specific to *SetReplace*.
  * `A1*` -> Utilities that are somewhat specific to *SetReplace* but could be moved to another package.
  * `A2*` -> Utilities that are specific to *SetReplace*, e.g., `style.m` and `Multihistory.m`.

## Examples

* Add explicit inputs and outputs (as screenshots in case of graphics) showcasing new functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/613)


taliesinb (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/613)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #619, opened 2021-03-03
Adding "MaxDestroyerEvents" as a step spec to WolframModel (C++)

### LINKED ISSUE #611: MaxDestroyerEvents
### The problem

In project Yellowstone, the format for `Multihistory` generators will be

```wl
Generator[System[rules_], generatorArgs] @ init
```

Some possible choices for `Generator` would be `GenerateSingleHistory` and `GenerateFullMultihistory` (or `GenerateAllHistories`). These allow one to select between generating a single or a multihistory.

However, we need a generator that can interpolate between them such as

```wl
GenerateMultihistory[system_, selectionCondition_, ordering_, stoppingCondition_]
```

In order to switch between single and multihistories, we will need to specify `"MaxDestroyerEvents"`. If `"MaxDestroyerEvents" -> 1`, it's a single history, and if `"MaxDestroyerEvents" -> Infinity`, it's a multihistory.

So, to implement `GenerateMultihistory` completely, we need `"MaxDestroyerEvents"` stopping condition which we don't currently have.

### Possible solution

Before `GenerateMultihistory` is implemented, we can add `"MaxDestroyerEvents"` as a `WolframModel` stopping condition, e.g.,

```wl
WolframModel[{{1, 2}, {2, 3}} -> {{2, 3}, {2, 4}, {3, 4}, {2, 1}},
             {{1, 1}, {1, 1}},
             <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 2|>,
             "EventSelectionFunction" -> "MultiwaySpacelike"]
```

To implement this, one will need to roughly do the following:

In C++,
1. Get rid of `SystemType` and pass `maxDestroyerEvents` instead to the `Set` initialization.
2. Add an instance variable to `CausalGraph` to track the number of destroyer events per expression.
3. In `Set::replaceOnce`, modify the following code to perform `matcher_.removeMatchesInvolvingExpressions` and related operations once the limit from step (2) is reached:
https://github.com/maxitg/SetReplace/blob/9d3d7fb3f1d405a25c208c7a1b03a69223b95ae2/libSetReplace/Set.cpp#L102-L111

In Wolfram Language:
1. Add another step spec to `setSubstitutionSystem.m` (and, as a result, to `WolframModel`).
2. Pass it as an extra parameter to the second argument of `$cpp$setReplace` in `setSubstitutionSystem$cpp.m`.

### Additional context

Implementing `GenerateMultihistory` itself is a big refactoring and is separate from this feature (although requires it). There will be another issue for that.

daneelsan:
## Changes

* Closes #611.
* Adds the `"MaxDestroyerEvents"` (per expression) to `WolframModel` as a step specification:
```wl
WolframModel[{{1, 2}, {2, 3}} -> {{2, 3}, {2, 4}, {3, 4}, {2, 1}},
             {{1, 1}, {1, 1}},
             <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 2|>,
             "EventSelectionFunction" -> "MultiwaySpacelike"]
```

## Comments
* This works as follows:
The causal graph keeps track of how many destroyer events each expression has (`expressionsIDsToDestroyerEventsCount_`).
During evolution, we check if the destroyer event count of an expression (`causalGraph_.destroyerEventsCount(expressionID)`) exceeds the maximum allowed (`maxDestroyerEvents_`), and if it does, we eliminate that expression and matches related to that expression from the matcher. 

* Missing tests, both for C++ and WL.
* Missing documentation.

## Examples

* Notice how `"MaxDestroyerEvents"` is not a stopping condition, but behaves similar to an event selection function:
```wl
In[] := Table[
  WolframModel[
    {{1, 2}, {2, 3}} -> {{2, 3}, {2, 4}, {3, 4}, {2, 1}},
    {{1, 1}, {1, 1}},
    <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> maxDestroyerEvents|>,
    "EventSelectionFunction" -> "MultiwaySpacelike"
    ]["ExpressionsEventsGraph", VertexLabels -> "Index"],
  {maxDestroyerEvents, {1, 2, Infinity}}] // GraphicsRow
```

MaxDestroyerEventsComparison

* This is also supported for `"EventSelectionFunction" -> None`:
 ```wl
In[] := Table[
  WolframModel[
    {{1, 2}, {2, 3}} -> {{2, 3}, {2, 4}, {3, 4}, {2, 1}},
    {{1, 1}, {1, 1}},
    <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> maxDestroyerEvents|>,
    "EventSelectionFunction" -> None
    ]["ExpressionsEventsGraph", VertexLabels -> "Index"],
  {maxDestroyerEvents, {1, 2, 3, Infinity}}] // GraphicsRow
```
MaxDestroyerEventsComparison2

* `"MaxDestroyerEvent" -> 1`, for any `"EventSelectionFunction"`, behaves effectively as a singleway system:
```wl
In[] := Table[
  WolframModel[
    {{1, 2}, {2, 3}} -> {{2, 3}, {2, 4}, {3, 4}, {2, 1}},
    {{1, 1}, {1, 1}},
    <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 1|>,
    "EventSelectionFunction" -> esf
    ]["ExpressionsEventsGraph", VertexLabels -> "Index"],
  {esf, {"GlobalSpacelike", "MultiwaySpacelike", 
    None}}] // GraphicsRow
```
MaxDestroyerEventsComparison3

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)


maxitg (changes requested):
Fantastic stuff!

We need documentation and tests in this PR, though, to maintain consistency. I think most of the tests should be on the C++ side so that we don't have to change them once we have the new API. (We need a few WL tests to test for error catching and that the WL API actually works.) And documentation can just add a few lines [here](https://github.com/maxitg/SetReplace/blob/master/Documentation/SymbolsAndFunctions/WolframModelAndWolframModelEvolutionObject/WolframModelAndWolframModelEvolutionObject.md#step-limiters).

> Reviewed 7 of 9 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619#-:-MUtuDRsBipzO_boWZ_2:b58noig)* status: all files reviewed, 9 unresolved discussions (waiting on @aokellermann, @daneelsan, and @taliesinb)

___
*[Kernel/setSubstitutionSystem.m, line 45 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtxI_PAhD6fKV1ZVjd:-MUtxI_PAhD6fKV1ZVje:bt2wzwh) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/Kernel/setSubstitutionSystem.m#L45)):*
> ```objc
> $maxEvents = "MaxEvents";
> $maxGenerationsLocal = "MaxGenerationsLocal";
> $maxDestroyerEvents = "MaxDestroyerEvents";
> ```

Do we actually need a value here, given that it's not a returnable termination reason?
___
*[libSetReplace/Event.cpp, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtzZHhChxSQc5HEDVK:-MUtzZHhChxSQc5HEDVL:bw4s16i) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Event.cpp#L38)):*
> ```cpp
>                                      const std::vector<ExpressionID>& inputExpressions,
>                                      const int outputExpressionsCount) {
>     updateDestroyerEventsCount(inputExpressions);
> ```

I think it would be more clear to call this function `incrementDestroyerEventsCount`. `update...` sounds to me like it might be an idempotent function.
___
*[libSetReplace/Set.hpp, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu0WuS6yZhsCnmSZw3:-MUu0WuS6yZhsCnmSZw4:bwcgs80) ([raw file](https://github.com/maxitg/setreplace/blob/3fd403fe49c97b009c06652409e12ecd7f0e0ffd/libSetReplace/Set.hpp#L26)):*
> ```cpp
>     NonPositiveAtoms,
>     AtomCountOverflow,
>     InvalidSystemType,
> ```

What will happen if someone passes a negative `maxDestroyerEvents` to the C++ API? Maybe, it makes sense to make it `uint64_t` to avoid internal argument checking.
___
*[libSetReplace/Set.cpp, line 109 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtuHTUBY7MJqM4ZKwW:-MUtuHTUBY7MJqM4ZKwX:b-gbsqm0) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Set.cpp#L109)):*
> ```cpp
>     addExpressions(outputExpressionIDs, namedRuleOutputs);
> 
>     if (maxDestroyerEvents_ == 1) {  // @TODO: maxDestroyerEvents_ == 0
> ```

Don't forget about the TODO.
___
*[libSetReplace/Set.cpp, line 118 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu2xRH8P-e7piKRTIY:-MUu2xRH8P-e7piKRTIZ:b26d4ru) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Set.cpp#L118)):*
> ```cpp
>       matcher_.deleteMatch(match);
>     } else {
>       // Only remove expressions whose destroyer events count exceed the maximum.
> ```

exceed -> reached
___
*[libSetReplace/Set.cpp, line 168 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu5vuu-dueDEa7r2F8:-MUu5vuu-dueDEa7r2F9:bswo7xe) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Set.cpp#L168)):*
> ```cpp
> 
>  private:
>   Implementation(std::vector<Rule> rules,
> ```

We should add `const std::vector<Rule>& rules` here as well. The constructor below used to say `rules_(std::move(rules))`, but for some reason, I removed it a few months ago.
___
*[libSetReplace/Set.cpp, line 197 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu3nNt8wLQZ4CKVJit:-MUu3nNuERUcMS7N2Ki6:b-lva5hl) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Set.cpp#L197)):*
> ```cpp
>   }
> 
>   std::vector<Rule> optimizeRules(std::vector<Rule> rules, int64_t maxDestroyerEvents) {
> ```

`const std::vector<Rule>& rules`
___
*[libSetReplace/Set.cpp, line 200 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu70h17ZmMg_br_KLY:-MUu70h17ZmMg_br_KLZ:byapqgc) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/Set.cpp#L200)):*
> ```cpp
>     if (maxDestroyerEvents == 1) {
>       /* EventSelectionFunction is set to All when maxDestroyerEvents is 1 because all concurrently matched
>         expressions are always spacelike, and All is much faster to evaluate. */
> ```

I think it would be useful to mention that in this case, `CausalGraph::SeparationTrackingMethod` is set to `None` in `separationTrackingMethod`. The real optimization is there, not here, and we need this `optimizeRules` to prevent it from breaking.
___
*[libSetReplace/test/Set_test.cpp, line 32 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu80Cz7QdIdEBT803M:-MUu80Cz7QdIdEBT803N:b-yqu059) ([raw file](https://github.com/maxitg/setreplace/blob/de49f5f09ca899b4b680ef599de60921f8ea0d09/libSetReplace/test/Set_test.cpp#L32)):*
> ```cpp
> constexpr auto doNotAbort = []() { return false; };
> 
> const int64_t max64int = std::numeric_limits<int64_t>::max();
> ```

`constexpr`

daneelsan (commented):
Yes, I'll work on the tests :)

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619#-:-MV9NB97-VIFERthsg6C:bo4qw6n)* status: 2 of 9 files reviewed, 9 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem.m, line 45 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtxI_PAhD6fKV1ZVjd:-MV9NMoD7vAUCMx2hUti:b-hn1ppj) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/Kernel/setSubstitutionSystem.m#L45)):*

Do we actually need a value here, given that it's not a returnable termination reason?
</blockquote></details>

Right, removed.
___
*[libSetReplace/Event.cpp, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtzZHhChxSQc5HEDVK:-MV9NXMG3Qk8B3GiCUAM:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Event.cpp#L38)):*

I think it would be more clear to call this function `incrementDestroyerEventsCount`. `update...` sounds to me like it might be an idempotent function.
</blockquote></details>

Done.
___
*[libSetReplace/Set.hpp, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu0WuS6yZhsCnmSZw3:-MV9Nd_D4JDB4_-xIVeG:b5zmyxy) ([raw file](https://github.com/maxitg/setreplace/blob/e569c5d2996f2d702643e42d7197818a778b08a7/libSetReplace/Set.hpp#L26)):*

What will happen if someone passes a negative `maxDestroyerEvents` to the C++ API? Maybe, it makes sense to make it `uint64_t` to avoid internal argument checking.
</blockquote></details>

Done. Should I also change the "max" number also? I.e. `$maxInt64` and `std::numeric_limits<int64_t>::max()` (both the value I pass it from WL, and the value used on the C++ side).
___
*[libSetReplace/Set.cpp, line 109 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtuHTUBY7MJqM4ZKwW:-MV9OI_56CnJ6AjIt4SK:b9eoohk) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L109)):*

Don't forget about the TODO.
</blockquote></details>

Should I do it in `replace(...)`? -> do a check `if (maxDestroyerEvents == 0) return 0;` before entering the `while(true)` loop.
___
*[libSetReplace/Set.cpp, line 118 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu2xRH8P-e7piKRTIY:-MV9OpEt0F9HgvYMq5Zo:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L118)):*

exceed -> reached
</blockquote></details>

Done.
___
*[libSetReplace/Set.cpp, line 168 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu5vuu-dueDEa7r2F8:-MV9OrOF7EvJIytnwTDn:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L168)):*

We should add `const std::vector<Rule>& rules` here as well. The constructor below used to say `rules_(std::move(rules))`, but for some reason, I removed it a few months ago.
</blockquote></details>

Done.
___
*[libSetReplace/Set.cpp, line 197 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu3nNt8wLQZ4CKVJit:-MV9OsB48YjjMaNj5lxW:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L197)):*

`const std::vector<Rule>& rules`
</blockquote></details>

Done.
___
*[libSetReplace/Set.cpp, line 200 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu70h17ZmMg_br_KLY:-MV9Osqu7FTEujylIet0:benxrrj) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L200)):*

I think it would be useful to mention that in this case, `CausalGraph::SeparationTrackingMethod` is set to `None` in `separationTrackingMethod`. The real optimization is there, not here, and we need this `optimizeRules` to prevent it from breaking.
</blockquote></details>

Let me know if the comments are clear. I (separation) tracked why this prevent breaking, but maybe I'm mistaken.
___
*[libSetReplace/test/Set_test.cpp, line 32 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu80Cz7QdIdEBT803M:-MV9PAEt59vwHtzLyPB3:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/test/Set_test.cpp#L32)):*

`constexpr`
</blockquote></details>

Done.

maxitg (changes requested):
> Reviewed 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: all files reviewed, 1 unresolved discussion (waiting on @aokellermann, @daneelsan, and @taliesinb)

___
*[libSetReplace/Set.hpp, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu0WuS6yZhsCnmSZw3:-MV9SuxK188b6gBshy8B:b-yiozbp) ([raw file](https://github.com/maxitg/setreplace/blob/e569c5d2996f2d702643e42d7197818a778b08a7/libSetReplace/Set.hpp#L26)):*

Done. Should I also change the "max" number also? I.e. `$maxInt64` and `std::numeric_limits<int64_t>::max()` (both the value I pass it from WL, and the value used on the C++ side).
</blockquote></details>

I think `mint` is signed, so I'm not sure it's possible.
___
*[libSetReplace/Set.cpp, line 109 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtuHTUBY7MJqM4ZKwW:-MV9TOHZ7pAdDBznPnii:btqek1) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L109)):*

Should I do it in `replace(...)`? -> do a check `if (maxDestroyerEvents == 0) return 0;` before entering the `while(true)` loop.
</blockquote></details>

Yeah, that'll work.
___
*[libSetReplace/Set.cpp, line 200 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu70h17ZmMg_br_KLY:-MV9U8x54eBUVXBYBhkz:b-h6tzrq) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L200)):*

Let me know if the comments are clear. I (separation) tracked why this prevent breaking, but maybe I'm mistaken.
</blockquote></details>

Yes, that's correct. Causal graph returns `SeparationType::Unknown` in this case.

daneelsan (commented):
Added documentation and some tests on the WL side. C++ tests coming soon...

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619#-:-MVCzCdXBuzZUlLnlUy8:b-a2j5n9)* status: all files reviewed, 1 unresolved discussion (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[libSetReplace/Set.hpp, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu0WuS6yZhsCnmSZw3:-MVCz4vNBz2xLF_AScuu:b1v1hi4) ([raw file](https://github.com/maxitg/setreplace/blob/e569c5d2996f2d702643e42d7197818a778b08a7/libSetReplace/Set.hpp#L26)):*

I think `mint` is signed, so I'm not sure it's possible.
</blockquote></details>

You are right.
By the way, the compilation fails on Windows because of 
`maxDestroyerEvents_ == std::numeric_limits<int64_t>::max()`.
I switched it to
`maxDestroyerEvents_ == static_cast<uint64_t>(std::numeric_limits<int64_t>::max())`
 and it still complaints that theres a type mismatch.
How would you fix this?
___
*[libSetReplace/Set.cpp, line 109 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUtuHTUBY7MJqM4ZKwW:-MVCz7Zz15_07EpMDXzA:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L109)):*

Yeah, that'll work.
</blockquote></details>

Done.
___
*[libSetReplace/Set.cpp, line 200 at r2](https://reviewable.io/reviews/maxitg/setreplace/619#-MUu70h17ZmMg_br_KLY:-MVCz8cC3a8w5l9MwFLp:bmi8ppe) ([raw file](https://github.com/maxitg/setreplace/blob/2e56807a37611a957849c7c7fbba42b1d1e6e3af/libSetReplace/Set.cpp#L200)):*

Yes, that's correct. Causal graph returns `SeparationType::Unknown` in this case.
</blockquote></details>

:100:

daneelsan:
> 
> You are right.
> By the way, the compilation fails on Windows because of
> `maxDestroyerEvents_ == std::numeric_limits<int64_t>::max()`.
> I switched it to
> `maxDestroyerEvents_ == static_cast<uint64_t>(std::numeric_limits<int64_t>::max())`
> and it still complaints that theres a type mismatch.
> How would you fix this?
> 
Ignore this. Apparently I didn't push my changes.

maxitg (changes requested):
I didn't have time to look at all of it yet, sorry. (I have to spend Sun-Mon working for Stephen.) I'll try to finish the review tonight, or, if not, definitely tomorrow morning.

> Reviewed 2 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619#-:-MVHLDnh0yM61htMo0dZ:bm8d5k7)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @aokellermann, @daneelsan, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVHKzeHBIrbhyMrATHf:b6yen0f) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*
> ```objc
> (* GlobalSpacelike is syntactic sugar for "EventSelectionFunction" -> "MultiwaySpacelike", "MaxDestroyerEvents" -> 1 *)
> 
> maxDestroyerEvents[Automatic | _ ? MissingQ, $globalSpacelike] = 1;
> ```

Shouldn't `"GlobalSpacelike"` force `"MaxDestroyerEvents" -> 1`, because otherwise, it's not really globally spacelike?

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVHNe9n6eMmbo4-nzct:b5slymh) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*

Shouldn't `"GlobalSpacelike"` force `"MaxDestroyerEvents" -> 1`, because otherwise, it's not really globally spacelike?
</blockquote></details>

It is a discussion about who has the priority: `"MaxDestroyerEvents"` or `"GlobalSpacelike"`.

I didn't like that:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 2|>]`
was transforming into:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 1|>]`
just because the default is `"EventSelectionFunction" -> "GlobalSpacelike"`.

So with this change:
`WolframModel[rule, init, <|"MaxEvents" -> 5|>]`
is "GlobalSpacelike", but when you specify "MaxDestroyerEvents", it esentially becomes:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> n|>, "ESF" -> "MultiwaySpacelike"]`.

maxitg (changes requested):
> Reviewed 3 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: all files reviewed, 3 unresolved discussions (waiting on @aokellermann, @daneelsan, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVJIu9h254cwnbhd_f-:b-1g6a7a) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*

It is a discussion about who has the priority: `"MaxDestroyerEvents"` or `"GlobalSpacelike"`.

I didn't like that:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 2|>]`
was transforming into:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 1|>]`
just because the default is `"EventSelectionFunction" -> "GlobalSpacelike"`.

So with this change:
`WolframModel[rule, init, <|"MaxEvents" -> 5|>]`
is "GlobalSpacelike", but when you specify "MaxDestroyerEvents", it esentially becomes:
`WolframModel[rule, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> n|>, "ESF" -> "MultiwaySpacelike"]`.


</blockquote></details>

The problem is that `"GlobalSpacelike"` by definition implies `"MaxDestroyerEvents" -> 1` as multistates in a multihistory with destroyer event counts larger than one wouldn't be globally spacelike anymore. So, using `"MaxDestroyerEvents"` larger than 1 is not actually correct in that case.

On the other hand, `1 <= 2`, so only doing a single destroyer event if `"MaxDestroyerEvents" -> 2` for `"GlobalSpacelike"` is not technically wrong, just slightly weird.

I understand that it's annoying to put `"EventSelectionFunction" -> "MultiwaySpacelike"` every time, but that's a limitation of the old design. We won't have this problem in *Yellowstone*.
___
*[libSetReplace/test/Set_test.cpp, line 74 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJNcKDBB2B33Q8LjLP:-MVJNcKDBB2B33Q8LjLQ:btif9j9) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L74)):*
> ```cpp
>   Matcher::OrderingSpec orderingSpec = {
>       {Matcher::OrderingFunction::SortedExpressionIDs, Matcher::OrderingDirection::Normal},
>       {Matcher::OrderingFunction::ReverseSortedExpressionIDs, Matcher::OrderingDirection::Normal},
> ```

Doesn't this do nothing, given that it's a reverse of the previous one?
___
*[libSetReplace/test/Set_test.cpp, line 82 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJQMghCoaA3FxNvCWg:-MVJQMghCoaA3FxNvCWh:b-mw9f52) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L82)):*
> ```cpp
> }
> 
> TEST(Set, maxDestroyerEvents1) {
> ```

It's a bit strange that your WL tests are more complete than C++ tests. Given that the evolution logic is entirely in C++, it seems like it would make more sense to put most tests in C++ (plus, they won't be WL API-dependent in that case). But, if it's easier to implement them in WL, it's not a big deal. They test the same thing anyway.
___
*[Tests/WolframModel.wlt, line 385 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJPrfy0Gnx3c44IEkb:-MVJPrfy0Gnx3c44IEkc:b9jqkan) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Tests/WolframModel.wlt#L385)):*
> ```Mathematica
>                                               "EventSelectionFunction" -> selectionFunction],
>                                 {selectionFunction, {"GlobalSpacelike", "MultiwaySpacelike", None}}]},
>         VerificationTest[Evaluate[SameQ @@ evolutions]]
> ```

`VerificationTest` should not take evaluated results as an input because it produces incomprehensible error messages. It should compute the result within its first argument.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: all files reviewed, 3 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVOv76m4cURnd2uOWzg:b51xn5t) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*

The problem is that `"GlobalSpacelike"` by definition implies `"MaxDestroyerEvents" -> 1` as multistates in a multihistory with destroyer event counts larger than one wouldn't be globally spacelike anymore. So, using `"MaxDestroyerEvents"` larger than 1 is not actually correct in that case.

On the other hand, `1 <= 2`, so only doing a single destroyer event if `"MaxDestroyerEvents" -> 2` for `"GlobalSpacelike"` is not technically wrong, just slightly weird.

I understand that it's annoying to put `"EventSelectionFunction" -> "MultiwaySpacelike"` every time, but that's a limitation of the old design. We won't have this problem in *Yellowstone*.
</blockquote></details>

Right, I understand.
But my logic is this:
`WolframModel[rules, init, <|"MaxEvents" -> 5|>]` -> `{1, "GlobalSpacelike"}`
`WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>]` -> `{3, "MultiwaySpacelike"}`

Now we have make choice
`WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>, "EventSelectionFunction" -> "GlobalSpacelike"]`
be `{1, "GlobalSpacelike"}` or `{3, "MultiwaySpacelike"}`?

I prefer the second one. We could say that the leftmost argument has a higher priority.

Ultimately, as you say it doesn't really matter...
___
*[libSetReplace/test/Set_test.cpp, line 74 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJNcKDBB2B33Q8LjLP:-MVOwDK816K4l2AZJaYw:b-vqgexj) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L74)):*

Doesn't this do nothing, given that it's a reverse of the previous one?
</blockquote></details>

I just copied and pasted `orderingSpec` from `testSet` above.
I'm not quite sure how ordering works, tbh.
Should I just remove the line?
___
*[libSetReplace/test/Set_test.cpp, line 82 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJQMghCoaA3FxNvCWg:-MVOwUGR5VSf5VAt0nfL:bo9a26s) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L82)):*

It's a bit strange that your WL tests are more complete than C++ tests. Given that the evolution logic is entirely in C++, it seems like it would make more sense to put most tests in C++ (plus, they won't be WL API-dependent in that case). But, if it's easier to implement them in WL, it's not a big deal. They test the same thing anyway.
</blockquote></details>

I'm just struggling to portray some of the WL tests in C++ form.
Do you have some recommended tests I should add?
___
*[Tests/WolframModel.wlt, line 385 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJPrfy0Gnx3c44IEkb:-MVOx20n2qEWMcyAKCNq:b-4hvi0m) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Tests/WolframModel.wlt#L385)):*

`VerificationTest` should not take evaluated results as an input because it produces incomprehensible error messages. It should compute the result within its first argument.
</blockquote></details>

Got it.

maxitg (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: all files reviewed, 3 unresolved discussions (waiting on @aokellermann, @daneelsan, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVOxVfl-C_nqIUCJD7h:byistsl) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*

Right, I understand.
But my logic is this:
`WolframModel[rules, init, <|"MaxEvents" -> 5|>]` -> `{1, "GlobalSpacelike"}`
`WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>]` -> `{3, "MultiwaySpacelike"}`

Now we have make choice
`WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>, "EventSelectionFunction" -> "GlobalSpacelike"]`
be `{1, "GlobalSpacelike"}` or `{3, "MultiwaySpacelike"}`?

I prefer the second one. We could say that the leftmost argument has a higher priority.

Ultimately, as you say it doesn't really matter...
</blockquote></details>

It's

```wl
WolframModel[rules,
             init,
             <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>,
             "EventSelectionFunction" -> "GlobalSpacelike"] -> {3, "GlobalSpacelike"}
```

which happens to be equivalent to

```wl
{1, "GlobalSpacelike"}
```

because of how `"GlobalSpacelike"` works.

It's similar to specifying

```wl
<|"MaxEvents" -> 1, "MaxGenerations" -> 10|>
```

and terminating after one event.

And, unfortunately,

```wl
WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>]
```

is equivalent to the case above because `"GlobalSpacelike"` is the default for the `"EventSelectionFunction"`.
___
*[libSetReplace/test/Set_test.cpp, line 74 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJNcKDBB2B33Q8LjLP:-MVOyeAE7W0T81EPI6pN:b-oy7ryh) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L74)):*

I just copied and pasted `orderingSpec` from `testSet` above.
I'm not quite sure how ordering works, tbh.
Should I just remove the line?
</blockquote></details>

Yes, I think so. You can remove them from both places.
___
*[libSetReplace/test/Set_test.cpp, line 82 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJQMghCoaA3FxNvCWg:-MVOyroEEF46jnh-kBun:b-10p5cq) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L82)):*

I'm just struggling to portray some of the WL tests in C++ form.
Do you have some recommended tests I should add?
</blockquote></details>

Well, the current WL tests would map into C++ quite well. They will require more code to make comparisons, and the `Table` will become a nested for-loop. But, as I said, there is no good reason to spend time on it now, especially since it will be more code in C++.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: 8 of 11 files reviewed, 3 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 121 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVHKzeHBIrbhyMrATHe:-MVPLQeY8lLYCXyrt_7P:b8tumw1) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Kernel/setSubstitutionSystem$cpp.m#L121)):*

It's

```wl
WolframModel[rules,
             init,
             <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>,
             "EventSelectionFunction" -> "GlobalSpacelike"] -> {3, "GlobalSpacelike"}
```

which happens to be equivalent to

```wl
{1, "GlobalSpacelike"}
```

because of how `"GlobalSpacelike"` works.

It's similar to specifying

```wl
<|"MaxEvents" -> 1, "MaxGenerations" -> 10|>
```

and terminating after one event.

And, unfortunately,

```wl
WolframModel[rules, init, <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>]
```

is equivalent to the case above because `"GlobalSpacelike"` is the default for the `"EventSelectionFunction"`.
</blockquote></details>

Reverted the change I made. This is now the behavior:
```wl
WolframModel[rules,
             init,
             <|"MaxEvents" -> 5, "MaxDestroyerEvents" -> 3|>,
             "EventSelectionFunction" -> "GlobalSpacelike"] -> {1, "GlobalSpacelike"}
```
___
*[libSetReplace/test/Set_test.cpp, line 74 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJNcKDBB2B33Q8LjLP:-MVPLk7yEQWP9A-fp6Vg:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L74)):*

Yes, I think so. You can remove them from both places.
</blockquote></details>

Done.
___
*[libSetReplace/test/Set_test.cpp, line 82 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJQMghCoaA3FxNvCWg:-MVPLkq3Ap7CexcgNSdf:b-vu70u0) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/libSetReplace/test/Set_test.cpp#L82)):*

Well, the current WL tests would map into C++ quite well. They will require more code to make comparisons, and the `Table` will become a nested for-loop. But, as I said, there is no good reason to spend time on it now, especially since it will be more code in C++.
</blockquote></details>

I moved some tests to the C++ side.
___
*[Tests/WolframModel.wlt, line 385 at r4](https://reviewable.io/reviews/maxitg/setreplace/619#-MVJPrfy0Gnx3c44IEkb:-MVPLqEm7j3G9GVGY6As:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/a62916b71232b82f40fc2d9d42eea6ed5a97ebe9/Tests/WolframModel.wlt#L385)):*

Got it.
</blockquote></details>

Done.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: 9 of 11 files reviewed, 4 unresolved discussions (waiting on @aokellermann, @maxitg, and @taliesinb)

___
*[libSetReplace/test/Set_test.cpp, line 88 at r5](https://reviewable.io/reviews/maxitg/setreplace/619#-MVPOMQD-YvP8AyxLrUP:-MVPcIUzATWz64M7IosV:bgoo58g) ([raw file](https://github.com/maxitg/setreplace/blob/c75cf25dd55f8b603953c1b7bb53e2f85d6db91b/libSetReplace/test/Set_test.cpp#L88)):*

It would be nice to refactor this into a separate function and also test for `.expressions()` equality in it for completeness.
</blockquote></details>

So a function that checks "equality" of 2 causal graphs?

```c
bool eqCausalGraph(... causalGraph1, ... causalGraph2) {
  if (causalGraph1.size() != causalGraph2.size()) return false;
  for (int i = 0; i <= causalGraph1.size(); ++i) {
     <compare input expressions>;
     <compare output expressions>; // ?
     <compare rule>; // ?
  }
}
```
___
*[libSetReplace/test/Set_test.cpp, line 95 at r5](https://reviewable.io/reviews/maxitg/setreplace/619#-MVPOF6e5DAszRqTjUQY:-MVPdURzB6v7Lgg8PWx9:bc6guyh) ([raw file](https://github.com/maxitg/setreplace/blob/c75cf25dd55f8b603953c1b7bb53e2f85d6db91b/libSetReplace/test/Set_test.cpp#L95)):*

Did you mean to do 50 events? With only 5 events, there doesn't seem to be much point in going with max destroyer events all the way to 20.
</blockquote></details>

Yes, midnight is not a good time to add tests.
___
*[libSetReplace/test/Set_test.cpp, line 99 at r5](https://reviewable.io/reviews/maxitg/setreplace/619#-MVPPD0Z0km1U5-mM6Jv:-MVPdZFUD7lj1ZaD54SQ:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/c75cf25dd55f8b603953c1b7bb53e2f85d6db91b/libSetReplace/test/Set_test.cpp#L99)):*

Why would they be equal? Isn't the idea here to check the number of destroyer events only?
</blockquote></details>

Done.
___
*[libSetReplace/test/Set_test.cpp, line 103 at r5](https://reviewable.io/reviews/maxitg/setreplace/619#-MVPQVfBBU7Vb8Olushr:-MVPda9xDdP0zUzVx76S:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/c75cf25dd55f8b603953c1b7bb53e2f85d6db91b/libSetReplace/test/Set_test.cpp#L103)):*

You should check `events.size()` instead. Some evolutions (in particular, `n == 0`) will terminate early. As a result, you are trying to access `events()[i]` out of bound, hence, the segfault in the tests.
</blockquote></details>

Done.

maxitg (approved):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/619)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aokellermann and @taliesinb)

___
*[libSetReplace/test/Set_test.cpp, line 88 at r5](https://reviewable.io/reviews/maxitg/setreplace/619#-MVPOMQD-YvP8AyxLrUP:-MVSHsIKCcsqAsoFUTKc:b7fjdo4) ([raw file](https://github.com/maxitg/setreplace/blob/c75cf25dd55f8b603953c1b7bb53e2f85d6db91b/libSetReplace/test/Set_test.cpp#L88)):*

So a function that checks "equality" of 2 causal graphs?

```c
bool eqCausalGraph(... causalGraph1, ... causalGraph2) {
  if (causalGraph1.size() != causalGraph2.size()) return false;
  for (int i = 0; i <= causalGraph1.size(); ++i) {
     <compare input expressions>;
     <compare output expressions>; // ?
     <compare rule>; // ?
  }
}
```

</blockquote></details>

Two *sets*, but never mind, since you are not doing this in the tests below anymore, it won't actually make the code any smaller.

PR #635, opened 2021-04-01
New names for MultisetSubstitutionSystem ordering functions

maxitg:
## Changes

* `"SortedInputExpressions"` and `"UnsortedInputExpressions"` weren't good names for ordering functions because, using the word "expression" instead of "token", they are specific to the multiset system, and it's better if they are not since they can then have shared documentation.
* This PR renames them to `"SortedInputTokenIndices"` and `"InputTokenIndices"`.

## Comments

* `"SortedInputTokenIndices"` might be harder to understand at first than `"OldestToken"`, but I think it's better because it explains exactly what it is whereas `"OldestToken"` might be interpreted in different ways. `"OldestToken"` is actually closer to `"MinInputTokenIndex"`, which we don't have yet, but we might in the future.
* Not sure if `"Unsorted"` is necessary in `"InputTokenIndices"`, so I removed it.

## Examples

* New names:

```wl
In[] := EventOrderingFunctions[MultisetSubstitutionSystem]
Out[] = {"InputCount", "SortedInputTokenIndices", "InputTokenIndices", "RuleIndex"}
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/635)

### DIFF
--- a/Kernel/A1$GenerateMultihistory.m
+++ b/Kernel/A1$GenerateMultihistory.m
@@ -52,7 +52,7 @@ $constraintsSpecPattern =
      MultisetSubstitutionSystem,
      <|"MaxGeneration" -> {Infinity, "NonNegativeIntegerOrInfinity"},
        "MinEventInputs" -> {0, "NonNegativeIntegerOrInfinity"}|>,
-     {"InputCount", "SortedInputExpressions", "UnsortedInputExpressions", "RuleIndex"},
+     {"InputCount", "SortedInputTokenIndices", "InputTokenIndices", "RuleIndex"},
      <|"MaxEvents" -> {Infinity, "NonNegativeIntegerOrInfinity"}|>] *)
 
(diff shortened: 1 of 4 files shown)


daneelsan (approved):
I agree that "Unsorted" is unnecessary. Also, +1 for "Token*", just like in "TokenEventGraph".

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/635#-:-MXFFnEP7P4TOxMiHiTz:b-vnk4kd)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @taliesinb)

taliesinb (approved):
:lgtm:

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/635#-:-MXbh8kj82aWrtktukW4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @maxitg)

PR #647, opened 2021-05-19
Move TimeConstrained from setSubstitutionSystem$cpp to libSetReplace

daneelsan:
## Changes

* Moves `TimeConstrained[...]` from `setSubstitutionSystem$cpp[...]` to inside `HypergraphSubstitutionSystem.cpp` in `libSetReplace`.
* Adds a new `TerminationReasonCode`: `8 -> "TimeConstrained"`.

## Comments

* `AbsoluteTiming` doesn't really match the time constraint:

```wl
In[] := WolframModel[
             {{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, {{1, 1}},
             Infinity,
             TimeConstraint -> 1] // AbsoluteTiming // First
Out[] = 2.61478
```

## Examples

```wl
In[] := WolframModel[{{1, 2}} -> {{1, 3}, {1, 3}, {3, 2}}, {{1, 1}}, Infinity, TimeConstraint -> 1]
```

image

```wl
In[] := %["TerminationReason"]
Out[] = "TimeConstraint"
```

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)

### DIFF
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -193,8 +193,11 @@ setSubstitutionSystem$cpp[
   ,
     If[!returnOnAbortQ, Abort[], Missing["Unknown", $Aborted]]
   ];
-  terminationReason = Replace[$terminationReasonCodes[cpp$terminationReason[setID]], {
-    $notTerminated -> $timeConstraint}];
+
+  terminationReason = $terminationReasonCodes[cpp$terminationReason[setID]];
+  If[(terminationReason === $timeConstraint) && !returnOnAbortQ, Abort[]];
+  terminationReason = Replace[terminationReason, $notTerminated -> $timeConstraint];
+
   resultAtoms = Union[Catenate[numericAtomLists]];
   inversePartialGlobalMap = Association[Reverse /@ Normal @ globalIndex];
   inverseGlobalMap = Association @ Thread[resultAtoms
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -186,6 +186,11 @@ setSubstitutionSystem$cpp[
   ,
     If[!returnOnAbortQ, Abort[]]
   ];
+
+  terminationReason = $terminationReasonCodes[cpp$terminationReason[setID]];
+  If[(terminationReason === $timeConstraint) && !returnOnAbortQ, Return @ $Aborted];
+  terminationReason = Replace[terminationReason, $notTerminated -> $timeConstraint];
+
   numericAtomLists = decodeAtomLists[cpp$setExpressions[setID]];
   events = decodeEvents[cpp$setEvents[setID]];
   maxCompleteGeneration = CheckAbort[
@@ -193,10 +198,6 @@ setSubstitutionSystem$cpp[
   ,
     If[!returnOnAbortQ, Abort[], Missing["Unknown", $Aborted]]
   ];
-
-  terminationReason = $terminationReasonCodes[cpp$terminationReason[setID]];
-  If[(terminationReason === $timeConstraint) && !returnOnAbortQ, Abort[]];
-  terminationReason = Replace[terminationReason, $notTerminated -> $timeConstraint];
 
   resultAtoms = Union[Catenate[numericAtomLists]];
   inversePartialGlobalMap = Association[Reverse /@ Normal @ globalIndex];
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -105,7 +105,6 @@ ruleAtomsToIndices[left_ :> right_, globalIndex_, localIndex_] := ModuleScope[
 
 $maxInt64 = 2^63 - 1;
 $maxUInt32 = 2^32 - 1;
-$maxDouble = 1.7969*^308;
 
 $terminationReasonCodes = <|
   0 -> $notTerminated,
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -13,7 +13,8 @@ importLibSetReplaceFunction[
    Integer,                  (* event selection function *)
    {Integer, 1, "Constant"}, (* ordering function index, forward / reverse, function, forward / reverse, ... *)
    Integer,                  (* event deduplication *)
-   Integer},                 (* random seed *)
+   (* random seed, passed as two numbers because LibraryLink does not support unsigned ints *)
+   {Integer, 1, "Constant"}},
   "Void"];
 
 importLibSetReplaceFunction[
@@ -103,7 +104,7 @@ ruleAtomsToIndices[left_ :> right_, globalIndex_, localIndex_] := ModuleScope[
   newLeft -> newRight
 ];
 
-$maxInt64 = 2^63 - 1;
+$unset = -1;
 $maxUInt32 = 2^32 - 1;
 
 $terminationReasonCodes = <|
@@ -121,7 +122,7 @@ $terminationReasonCodes = <|
 (* GlobalSpacelike is syntactic sugar for "EventSelectionFunction" -> "MultiwaySpacelike", "MaxDestroyerEvents" -> 1 *)
 
 maxDestroyerEvents[_, $globalSpacelike] = 1;
-maxDestroyerEvents[Automatic | _ ? MissingQ | Infinity, _] = $maxInt64;
+maxDestroyerEvents[Automatic | _ ? MissingQ | Infinity, _] = $unset;
 maxDestroyerEvents[n_, _] := n;
 
 (* 0 -> All
@@ -173,15 +174,16 @@ setSubstitutionSystem$cpp[
     maxDestroyerEvents[stepSpec[$maxDestroyerEvents], eventSelectionFunction],
     Catenate[Replace[eventOrderingFunction, $orderingFunctionCodes, {2}]],
     Replace[eventDeduplication, $eventDeduplicationCodes],
-    RandomInteger[{0, $maxUInt32}]
+    IntegerDigits[RandomInteger[{0, $maxUInt32}], 2^16, 2]
   ];
+  
   CheckAbort[
     cpp$setReplace[
       setID,
       stepSpec /@ {
           $maxEvents, $maxGenerationsLocal, $maxFinalVertices, $maxFinalVertexDegree, $maxFinalExpressions} /.
-        {Infinity | (_ ? MissingQ) -> $maxInt64},
-      timeConstraint /. Infinity -> -1]
+        {Infinity | (_ ? MissingQ) -> $unset},
+      timeConstraint /. Infinity -> $unset]
   ,
     If[!returnOnAbortQ, Abort[]]
   ];
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -176,7 +176,7 @@ setSubstitutionSystem$cpp[
     Replace[eventDeduplication, $eventDeduplicationCodes],
     IntegerDigits[RandomInteger[{0, $maxUInt32}], 2^16, 2]
   ];
-  
+
   CheckAbort[
     cpp$setReplace[
       setID,
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -19,7 +19,8 @@ importLibSetReplaceFunction[
 importLibSetReplaceFunction[
   "hypergraphSubstitutionSystemReplace" -> cpp$setReplace,
   {Integer,                   (* set ID *)
-   {Integer, 1, "Constant"}}, (* {events, generations, atoms, max expressions per atom, expressions} *)
+   {Integer, 1, "Constant"},  (* {events, generations, atoms, max expressions per atom, expressions} *)
+   Real},                     (* time constraint *)
   "Void"];
 
 importLibSetReplaceFunction[
@@ -104,6 +105,7 @@ ruleAtomsToIndices[left_ :> right_, globalIndex_, localIndex_] := ModuleScope[
 
 $maxInt64 = 2^63 - 1;
 $maxUInt32 = 2^32 - 1;
+$maxDouble = 1.7969*^308;
 
 $terminationReasonCodes = <|
   0 -> $notTerminated,
@@ -113,7 +115,8 @@ $terminationReasonCodes = <|
   4 -> $maxFinalVertexDegree,
   5 -> $maxFinalExpressions,
   6 -> $fixedPoint,
-  7 -> $Aborted
+  7 -> $Aborted,
+  8 -> $timeConstraint
 |>;
 
 (* GlobalSpacelike is syntactic sugar for "EventSelectionFunction" -> "MultiwaySpacelike", "MaxDestroyerEvents" -> 1 *)
@@ -173,16 +176,14 @@ setSubstitutionSystem$cpp[
     Replace[eventDeduplication, $eventDeduplicationCodes],
     RandomInteger[{0, $maxUInt32}]
   ];
-  TimeConstrained[
-    CheckAbort[
-      cpp$setReplace[
-        setID,
-        stepSpec /@ {
-            $maxEvents, $maxGenerationsLocal, $maxFinalVertices, $maxFinalVertexDegree, $maxFinalExpressions} /.
-          {Infinity | (_ ? MissingQ) -> $maxInt64}],
-      If[!returnOnAbortQ, Abort[], terminationReason = $Aborted]],
-    timeConstraint,
-    If[!returnOnAbortQ, Return[$Aborted], terminationReason = $timeConstraint]];
+  CheckAbort[
+    cpp$setReplace[
+      setID,
+      stepSpec /@ {
+          $maxEvents, $maxGenerationsLocal, $maxFinalVertices, $maxFinalVertexDegree, $maxFinalExpressions} /.
+        {Infinity | (_ ? MissingQ) -> $maxInt64},
+      timeConstraint /. Infinity -> $maxDouble],
+    If[!returnOnAbortQ, Abort[], terminationReason = $Aborted]];
   numericAtomLists = decodeAtomLists[cpp$setExpressions[setID]];
   events = decodeEvents[cpp$setEvents[setID]];
   maxCompleteGeneration = CheckAbort[
--- a/Kernel/setSubstitutionSystem$cpp.m
+++ b/Kernel/setSubstitutionSystem$cpp.m
@@ -182,15 +182,18 @@ setSubstitutionSystem$cpp[
       stepSpec /@ {
           $maxEvents, $maxGenerationsLocal, $maxFinalVertices, $maxFinalVertexDegree, $maxFinalExpressions} /.
         {Infinity | (_ ? MissingQ) -> $maxInt64},
-      timeConstraint /. Infinity -> $maxDouble],
-    If[!returnOnAbortQ, Abort[], terminationReason = $Aborted]];
+      timeConstraint /. Infinity -> -1]
+  ,
+    If[!returnOnAbortQ, Abort[]]
+  ];
   numericAtomLists = decodeAtomLists[cpp$setExpressions[setID]];
   events = decodeEvents[cpp$setEvents[setID]];
   maxCompleteGeneration = CheckAbort[
-    Replace[cpp$maxCompleteGeneration[setID], LibraryFunctionError[___] -> Missing["Unknown", $Aborted]],
-    If[!returnOnAbortQ, Abort[], terminationReason = $Aborted; Missing["Unknown", $Aborted]]];
+    Replace[cpp$maxCompleteGeneration[setID], LibraryFunctionError[___] -> Missing["Unknown", $Aborted]]
+  ,
+    If[!returnOnAbortQ, Abort[], Missing["Unknown", $Aborted]]
+  ];
   terminationReason = Replace[$terminationReasonCodes[cpp$terminationReason[setID]], {
-    $Aborted -> terminationReason,
     $notTerminated -> $timeConstraint}];
   resultAtoms = Union[Catenate[numericAtomLists]];
   inversePartialGlobalMap = Association[Reverse /@ Normal @ globalIndex];
--- a/libSetReplace/HypergraphSubstitutionSystem.cpp
+++ b/libSetReplace/HypergraphSubstitutionSystem.cpp
@@ -1,7 +1,5 @@
 #include "HypergraphSubstitutionSystem.hpp"
 
-#include <time.h>
-
 #include <algorithm>
 #include <limits>
 #include <memory>
@@ -58,7 +56,9 @@ class HypergraphSubstitutionSystem::Implementation {
               return causalGraph_.tokenSeparation(first, second);
             }) {}
 
-  int64_t replaceOnce(const std::function<bool()> shouldAbort, bool resetStepSpec = false) {
+  int64_t replaceOnce(const std::function<bool()> shouldAbort,
+                      const std::function<bool()> shouldTimeOut,
+                      bool resetStepSpec = false) {
     if (resetStepSpec) {
       updateStepSpec(StepSpecification{});
     }
@@ -69,10 +69,14 @@ class HypergraphSubstitutionSystem::Implementation {
       return 0;
     }
 
-    indexNewTokens([this, &shouldAbort]() {
+    indexNewTokens([this, &shouldAbort, &shouldTimeOut]() {
       const bool isAborted = shouldAbort();
       if (isAborted) terminationReason_ = TerminationReason::Aborted;
-      return isAborted;
+
+      const bool isTimeConstrained = shouldTimeOut();
+      if (isTimeConstrained) terminationReason_ = TerminationReason::TimeConstrained;
+
+      return isAborted || isTimeConstrained;
     });
     if (matcher_.empty()) {
       if (causalGraph_.largestGeneration() == stepSpec_.maxGenerationsLocal) {
@@ -134,24 +138,28 @@ class HypergraphSubstitutionSystem::Implementation {
 
   int64_t replace(const StepSpecification stepSpec,
                   const std::function<bool()>& shouldAbort,
-                  double const timeConstraint) {
+                  double const timeConstraintSec) {
     updateStepSpec(stepSpec);
     int64_t count = 0;
     if (maxDestroyerEvents_ == 0) {
       return count;
     }
-    time_t startTime = time(0);
+
+    auto timeConstraint = HypergraphSubstitutionSystem::maxTimeConstraint;
+    if (timeConstraintSec > 0) {
+      // https://stackoverflow.com/questions/45884081/terse-conversion-from-double-seconds-to-stdchronosteady-clockduration
+      timeConstraint = std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+          std::chrono::duration<double>(timeConstraintSec));
+    }
+    auto startTime = std::chrono::steady_clock::now();
+    const std::function<bool()> shouldTimeOut = [startTime, timeConstraint]() {
+      return (std::chrono::steady_clock::now() - startTime) > timeConstraint;
+    };
+
     while (true) {
-      if (replaceOnce(shouldAbort)) {
+      if (replaceOnce(shouldAbort, shouldTimeOut)) {
         ++count;
       } else {
-        return count;
-      }
-
-      // Custom TimeConstraint function
-      double secondsElapsed = difftime(time(0), startTime);
-      if (secondsElapsed > timeConstraint) {
-        terminationReason_ = TerminationReason::TimeConstrained;
         return count;
       }
     }
@@ -409,14 +417,15 @@ HypergraphSubstitutionSystem::HypergraphSubstitutionSystem(
     : implementation_(std::make_shared<Implementation>(
           rules, initialTokens, maxDestroyerEvents, orderingSpec, eventDeduplication, randomSeed)) {}
 
-int64_t HypergraphSubstitutionSystem::replaceOnce(const std::function<bool()>& shouldAbort) {
-  return implementation_->replaceOnce(shouldAbort, true);
+int64_t HypergraphSubstitutionSystem::replaceOnce(const std::function<bool()>& shouldAbort,
+                                                  const std::function<bool()>& shouldTimeOut) {
+  return implementation_->replaceOnce(shouldAbort, shouldTimeOut, true);
 }
 
 int64_t HypergraphSubstitutionSystem::replace(const StepSpecification& stepSpec,
                                               const std::function<bool()>& shouldAbort,
-                                              double const timeConstraint) {
-  return implementation_->replace(stepSpec, shouldAbort, timeConstraint);
+                                              double const timeConstraintSec) {
+  return implementation_->replace(stepSpec, shouldAbort, timeConstraintSec);
 }
 
 std::vector<AtomsVector> HypergraphSubstitutionSystem::tokens() const { return implementation_->tokens(); }
--- a/libSetReplace/HypergraphSubstitutionSystem.cpp
+++ b/libSetReplace/HypergraphSubstitutionSystem.cpp
@@ -56,9 +56,7 @@ class HypergraphSubstitutionSystem::Implementation {
               return causalGraph_.tokenSeparation(first, second);
             }) {}
 
-  int64_t replaceOnce(const std::function<bool()> shouldAbort,
-                      const std::function<bool()> shouldTimeOut,
-                      bool resetStepSpec = false) {
+  int64_t replaceOnce(const std::function<bool()> shouldAbortOrTimeOut, bool resetStepSpec = false) {
     if (resetStepSpec) {
       updateStepSpec(StepSpecification{});
     }
@@ -69,15 +67,7 @@ class HypergraphSubstitutionSystem::Implementation {
       return 0;
     }
 
-    indexNewTokens([this, &shouldAbort, &shouldTimeOut]() {
-      const bool isAborted = shouldAbort();
-      if (isAborted) terminationReason_ = TerminationReason::Aborted;
-
-      const bool isTimeConstrained = shouldTimeOut();
-      if (isTimeConstrained) terminationReason_ = TerminationReason::TimeConstrained;
-
-      return isAborted || isTimeConstrained;
-    });
+    indexNewTokens(shouldAbortOrTimeOut);
     if (matcher_.empty()) {
       if (causalGraph_.largestGeneration() == stepSpec_.maxGenerationsLocal) {
         terminationReason_ = TerminationReason::MaxGenerationsLocal;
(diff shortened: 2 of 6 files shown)


daneelsan:
@maxitg Right now I covert `"TimeConstraint" -> Infinity` to `maxdouble` before passing in to `cpp$setReplace`. 
Inside `replace` I'm doing this:
```cpp
    time_t startTime = time(0);
    while (true) {
      if (replaceOnce(shouldAbort)) {
        ++count;
      } else {
        return count;
      }

      // Custom TimeConstraint function
      double secondsElapsed = difftime(time(0), startTime);
      if (secondsElapsed > timeConstraint) {
        terminationReason_ = TerminationReason::TimeConstrained;
        return count;
      }
    }
```

So even if I pass `Infinity`, it still computes `time(0)` and `difftime`. Do you think it would be better to do something like:
```cpp
    time_t startTime = time(0);
    auto timeConstrainted = []() { return false; };
    if (timeConstraint > 0) {
       timeConstrained = []() { return difftime(time(0), startTime) > timeConstraint; }
    }
   ...
      if (timeConstrained()) {
        terminationReason_ = TerminationReason::TimeConstrained;
        return count;
      }
```
(don't know how lambdas work but that's the idea)

maxitg (changes requested):
It does not match the time constraint on master as well. That is because it takes time to post-process the output from C++ on the WL side. This problem will go away once we switch to keeping objects in C++. However, there is an additional problem in your code, which is that `time(...)` has the resolution of seconds, and so the time constraint could go as high as 2 seconds in your example depending on when it was started.

As for disabling checking, we need to make sure performance does not get worse because of this. And if it doesn't, it does not matter if we disable the check or not. CI seems to be failing due to performance, though...

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647#-:-Ma4Sf8_3yryRy42VPuO:b-p0cny4)* status: all files reviewed, 8 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4Z8OZCby_6Sjr1fuz:-Ma4Z8OZCby_6Sjr1fv-:budv3xr) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L108)):*
> ```objc
> $maxInt64 = 2^63 - 1;
> $maxUInt32 = 2^32 - 1;
> $maxDouble = 1.7969*^308;
> ```

Is this equivalent to `$MaxMachineNumber`?
___
*[Kernel/setSubstitutionSystem$cpp.m, line 186 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4_K5A7jD_7AWiv-G0:-Ma4_K5A7jD_7AWiv-G1:b-5qv38p) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L186)):*
> ```objc
>         {Infinity | (_ ? MissingQ) -> $maxInt64},
>       timeConstraint /. Infinity -> $maxDouble],
>     If[!returnOnAbortQ, Abort[], terminationReason = $Aborted]];
> ```

There is no point assigning `terminationReason` to anything here since the disambiguation below is no longer needed.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 191 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aP295MAMxtMnplib:-Ma4aP295MAMxtMnplic:b-taqph0) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L191)):*
> ```objc
>   maxCompleteGeneration = CheckAbort[
>     Replace[cpp$maxCompleteGeneration[setID], LibraryFunctionError[___] -> Missing["Unknown", $Aborted]],
>     If[!returnOnAbortQ, Abort[], terminationReason = $Aborted; Missing["Unknown", $Aborted]]];
> ```

Setting `terminationReason` is no longer necessary since `HypergraphMatcher` will set it automatically.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 193 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aZGt8F4Git08OqvB:-Ma4aZGt8F4Git08OqvC:b-8jqkag) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L193)):*
> ```objc
>     If[!returnOnAbortQ, Abort[], terminationReason = $Aborted; Missing["Unknown", $Aborted]]];
>   terminationReason = Replace[$terminationReasonCodes[cpp$terminationReason[setID]], {
>     $Aborted -> terminationReason,
> ```

This disambiguation is no longer needed.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aeC67rgyVyjDhmX0:-Ma4aeC67rgyVyjDhmX1:br8v01n) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*
> ```cpp
>   int64_t replace(const StepSpecification& stepSpec,
>                   const std::function<bool()>& shouldAbort,
>                   const double timeConstraint);
> ```

`timeConstraint` should either have `sec` in the name (e.g., `timeConstraintSec`), or, better yet, take [`std::chrono::duration`](https://en.cppreference.com/w/cpp/chrono/duration), which would be a C++ equivalent of using `Quantity`.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-Ma4ctWZCwId2lIeRX5t:b-qi2goh) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*
> ```cpp
>   int64_t replace(const StepSpecification& stepSpec,
>                   const std::function<bool()>& shouldAbort,
>                   const double timeConstraint);
> ```

I think it would be better to make `timeConstraint` an optional argument (infinity by default).
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 143 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4b1yY-XzpMDuz8lX7:-Ma4b1yY-XzpMDuz8lX8:bn8beuv) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L143)):*
> ```cpp
>       return count;
>     }
>     time_t startTime = time(0);
> ```

There are two problems with `time`:
1. Its resolution is 1 second, which is not always good enough (`time_t` is an integer, not a real).
2. It's a wall time rather than a steady time, meaning that it may change if the system clock is adjusted, for example, which will break the time constraint.

It would be better to use `std::chrono::steady_clock`, which does not have these problems.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 152 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4bxDK6fzDUp49FV9H:-Ma4bxDK6fzDUp49FV9I:b-33f0zi) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L152)):*
> ```cpp
> 
>       // Custom TimeConstraint function
>       double secondsElapsed = difftime(time(0), startTime);
> ```

The problem with this is that even a single event may take a long time (because it might suddenly encounter an exponential number of possible matches). This is why `shouldAbort` is passed further. The time constraint checking should be passed further as well (probably to the same place).

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: all files reviewed, 8 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4Z8OZCby_6Sjr1fuz:-Ma6EiDdBIIdgMNSOeMZ:b-iydjcd) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L108)):*

Is this equivalent to `$MaxMachineNumber`?
</blockquote></details>

It is, I googled max double (though it could be wrong).
___
*[Kernel/setSubstitutionSystem$cpp.m, line 186 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4_K5A7jD_7AWiv-G0:-Ma6FF8X7v6czhBX07Sp:b-80bnvp) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L186)):*

There is no point assigning `terminationReason` to anything here since the disambiguation below is no longer needed.
</blockquote></details>

True.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 191 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aP295MAMxtMnplib:-Ma6FLM48ZKNy2WCk8PF:b-80bnvp) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L191)):*

Setting `terminationReason` is no longer necessary since `HypergraphMatcher` will set it automatically.
</blockquote></details>

True.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 193 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aZGt8F4Git08OqvB:-Ma6FNHh6_OuZ_ZRWiqq:b-80bnvp) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L193)):*

This disambiguation is no longer needed.
</blockquote></details>

True.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aeC67rgyVyjDhmX0:-Ma6FPMsBb9HOq6XGr7P:bhy9ot6) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

`timeConstraint` should either have `sec` in the name (e.g., `timeConstraintSec`), or, better yet, take [`std::chrono::duration`](https://en.cppreference.com/w/cpp/chrono/duration), which would be a C++ equivalent of using `Quantity`.
</blockquote></details>

Will do.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-Ma6FRfD2gHgsQBwOzPM:b3d9r9u) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

I think it would be better to make `timeConstraint` an optional argument (infinity by default).
</blockquote></details>

Will do. By infinity do you mean: `std::chrono::duration<double>::max();` ?
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 143 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4b1yY-XzpMDuz8lX7:-Ma6FvoW07yWltYav97M:b-4hvi0m) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L143)):*

There are two problems with `time`:
1. Its resolution is 1 second, which is not always good enough (`time_t` is an integer, not a real).
2. It's a wall time rather than a steady time, meaning that it may change if the system clock is adjusted, for example, which will break the time constraint.

It would be better to use `std::chrono::steady_clock`, which does not have these problems.
</blockquote></details>

Got it.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: all files reviewed, 8 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-Ma6Mzse3PcbYWDju0LW:b-ixywvg) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Will do. By infinity do you mean: `std::chrono::duration<double>::max();` ?
</blockquote></details>

Better yet `std::chrono::seconds::max();`

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: all files reviewed, 8 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 152 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4bxDK6fzDUp49FV9H:-Ma6G-s1E2EoBhLkhhGi:b5w0z3p) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L152)):*

The problem with this is that even a single event may take a long time (because it might suddenly encounter an exponential number of possible matches). This is why `shouldAbort` is passed further. The time constraint checking should be passed further as well (probably to the same place).
</blockquote></details>

Im following the track of `shouldAbort` across all functions and man it goes deep.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4Z8OZCby_6Sjr1fuz:-Ma6vRN28rQn_N_77JhA:bdp440u) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L108)):*

It is, I googled max double (though it could be wrong).
</blockquote></details>

Decided to pass -1.0, instead. Internally, any negative double would be converted to the max duration possible.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 186 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4_K5A7jD_7AWiv-G0:-Ma6vbyUAhq7DISAs14L:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L186)):*

True.
</blockquote></details>

Done.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 191 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aP295MAMxtMnplib:-Ma6vkWk7O42-wTufCvB:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L191)):*

True.
</blockquote></details>

Done.
___
*[Kernel/setSubstitutionSystem$cpp.m, line 193 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aZGt8F4Git08OqvB:-Ma6vlUm8knoiswLBYxM:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L193)):*

True.
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aeC67rgyVyjDhmX0:-Ma6vnzeEvbv1n-xvc5l:b-epaspu) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Will do.
</blockquote></details>

I still pass a double to `replace`. Internally it is converted to `std::chrono::steady_clock::duration`. Let me know what you think.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-Ma6w51d6WwFZwY_peCJ:b-dvrooj) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Better yet `std::chrono::seconds::max();`
</blockquote></details>

Bestest yet: `std::chrono::steady_clock::duration::max()`.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 143 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4b1yY-XzpMDuz8lX7:-Ma6wFXa3tLoGTnUlfx0:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L143)):*

Got it.
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 152 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4bxDK6fzDUp49FV9H:-Ma6wGbd5t-zB7Aiy_HT:blj0229) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.cpp#L152)):*

Im following the track of `shouldAbort` across all functions and man it goes deep.
</blockquote></details>

I copied what you did with `shouldAbort` and passed it down to `replaceOnce`.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 75 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9XzDp6hVeJkQpR_Bo:-Ma9XzDp6hVeJkQpR_Bp:b-t4p8q2) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L75)):*
> ```cpp
>       const bool isAborted = shouldAbort();
>       if (isAborted) terminationReason_ = TerminationReason::Aborted;
> ```

@maxitg Or:
```cpp
      if (shouldAbort()) {
        terminationReason_ = TerminationReason::Aborted;
        return true;
      }

      if (shouldTimeOut()) {
         terminationReason_ = TerminationReason::TimeConstrained;
         return true;
      }

      return false;

```
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 149 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9YPhR92Li3mVy_Zrk:-Ma9YPhR92Li3mVy_Zrl:bmnc29z) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L149)):*
> ```cpp
> 
>     auto timeConstraint = HypergraphSubstitutionSystem::maxTimeConstraint;
>     if (timeConstraintSec > 0) {
> ```

Here is the place where I could set `shouldTimeOut` to be `{ return false; }` if `timeConstraintSec` is negative.

maxitg (changes requested):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: all files reviewed, 10 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4Z8OZCby_6Sjr1fuz:-Ma9q0t15D1s5Xd2J-2x:b-vjfxkn) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L108)):*

Decided to pass -1.0, instead. Internally, any negative double would be converted to the max duration possible.
</blockquote></details>

So, `$maxDouble` is no longer needed then?
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aeC67rgyVyjDhmX0:-Ma9swfW95jnq5ZguvyA:b-bnum3n) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

I still pass a double to `replace`. Internally it is converted to `std::chrono::steady_clock::duration`. Let me know what you think.
</blockquote></details>

Why not just use a duration from `std::chrono`? Then, users will get the benefit of type safety, etc. as well.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-Ma9tYXuDhbDFQnoa5lQ:b-dk82sa) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Bestest yet: `std::chrono::steady_clock::duration::max()`.
</blockquote></details>

Also, we don't need to do tricks with `-1` here anymore since we don't have a problem with 64-bit support in C++, and `::duration::max()` seems a lot more natural here (not to mention it works automatically).
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 33 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9vNm6-LKFLEp82YDK:-Ma9vNm6-LKFLEp82YDL:b-eh2iuh) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L33)):*
> ```cpp
>   static constexpr int64_t stepLimitDisabled = std::numeric_limits<int64_t>::max();
> 
>   static constexpr std::chrono::steady_clock::duration maxTimeConstraint = std::chrono::steady_clock::duration::max();
> ```

Since this is public, surely we should make it possible to pass it into `replace`?
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 33 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-MaAb0tw2fMD7dP3OCq1:-MaAb0tw2fMD7dP3OCq2:b-vel0ms) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L33)):*
> ```cpp
>   static constexpr int64_t stepLimitDisabled = std::numeric_limits<int64_t>::max();
> 
>   static constexpr std::chrono::steady_clock::duration maxTimeConstraint = std::chrono::steady_clock::duration::max();
> ```

Also, I think a clearer name would be `timeConstraintDisabled`, similar to the values in `StepSpecification`.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 89 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9sDRZ8mOLU_G82x6U:-Ma9sDRZ8mOLU_G82x6V:b-7nupbo) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L89)):*
> ```cpp
>    * @return 1 if substitution was made, 0 if no matches were found.
>    */
>   int64_t replaceOnce(const std::function<bool()>& shouldAbort, const std::function<bool()>& shouldTimeOut);
> ```

Surely, we don't want the user to pass this function?
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 60 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9umsu2-3-ApmVKBEX:-Ma9umsu2-3-ApmVKBEY:bbrrnwi) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L60)):*
> ```cpp
> 
>   int64_t replaceOnce(const std::function<bool()> shouldAbort,
>                       const std::function<bool()> shouldTimeOut,
> ```

I'm not sure why would we need to pass another lambda here. Why not just pass a different `shouldAbort` that calls the old one and also incorporates the time constraint (like you do below)?
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 75 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9XzDp6hVeJkQpR_Bo:-Ma9vAOv0_AkTV1-xDxQ:bevwzp7) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L75)):*

@maxitg Or:
```cpp
      if (shouldAbort()) {
        terminationReason_ = TerminationReason::Aborted;
        return true;
      }

      if (shouldTimeOut()) {
         terminationReason_ = TerminationReason::TimeConstrained;
         return true;
      }

      return false;

```
</blockquote></details>

Yes, this version looks cleaner to me.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 149 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9YPhR92Li3mVy_Zrk:-MaAbiZZ94K6Xu3z1P_v:bq0gycs) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L149)):*

Here is the place where I could set `shouldTimeOut` to be `{ return false; }` if `timeConstraintSec` is negative.
</blockquote></details>

I don't see any measurable performance degradation, so I don't think it's necessary:

```console
$ ./performanceTest.wls master HEAD 20
Testing 016287622e8537086e922700d731665cf89831e2

Testing 0fb1ff5df0d50e99a63a3eef8956b49402f5f519

Single-input rule                       3.4 ± 0.9 %
Medium rule                             4.0 ± 0.4 %
Sequential rule                         3.3 ± 1.3 %
Large rule                              -7.2 ± 3.1 %
Exponential-match-count rule            0.7 ± 0.4 %
CA emulator                             -3. ± 4. %
Multiset addition                       7.8 ± 3.0 %
```
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 150 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-MaAbrex64ZB8Y3c87Vg:-MaAbrex64ZB8Y3c87Vh:bwehrhp) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L150)):*
> ```cpp
>     auto timeConstraint = HypergraphSubstitutionSystem::maxTimeConstraint;
>     if (timeConstraintSec > 0) {
>       // https://stackoverflow.com/questions/45884081/terse-conversion-from-double-seconds-to-stdchronosteady-clockduration
> ```

I don't think we need a reference for what is essentially standard (even if complicated) C++.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: 1 of 6 files reviewed, 8 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/setSubstitutionSystem$cpp.m, line 108 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4Z8OZCby_6Sjr1fuz:-MaAz0I6C6E4i1pU00vL:bu3f617) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/Kernel/setSubstitutionSystem$cpp.m#L108)):*

So, `$maxDouble` is no longer needed then?
</blockquote></details>

Oops, forgot to remove it.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4aeC67rgyVyjDhmX0:-MaB-4G5CFGaZyBbgOS2:b-47129s) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Why not just use a duration from `std::chrono`? Then, users will get the benefit of type safety, etc. as well.
</blockquote></details>

hmmm true. I thought that it would be easier for users to just pass number of seconds (as a double).
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 95 at r1](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma4ctWZCwId2lIeRX5s:-MaB-D_R-uoNgxYcJw-K:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/189b96a22bf4e5ce05422bec96b1cec5fe097f64/libSetReplace/HypergraphSubstitutionSystem.hpp#L95)):*

Also, we don't need to do tricks with `-1` here anymore since we don't have a problem with 64-bit support in C++, and `::duration::max()` seems a lot more natural here (not to mention it works automatically).
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 33 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9vNm6-LKFLEp82YDK:-MaAzg351yg7ccelvb33:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L33)):*

Since this is public, surely we should make it possible to pass it into `replace`?
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 33 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-MaAb0tw2fMD7dP3OCq1:-MaAzeqi3vPsV__CXqUF:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L33)):*

Also, I think a clearer name would be `timeConstraintDisabled`, similar to the values in `StepSpecification`.
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.hpp, line 89 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9sDRZ8mOLU_G82x6U:-MaAzrJa3kxKN-T1-7R_:bujzrq0) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.hpp#L89)):*

Surely, we don't want the user to pass this function?
</blockquote></details>

Merged shouldAbort and shouldTimeOut into one a you suggested.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 60 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9umsu2-3-ApmVKBEX:-MaB-OYi8zGjMReYD6wj:b-7zm128) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L60)):*

I'm not sure why would we need to pass another lambda here. Why not just pass a different `shouldAbort` that calls the old one and also incorporates the time constraint (like you do below)?
</blockquote></details>

Merged the lambdas.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 75 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-Ma9XzDp6hVeJkQpR_Bo:-MaCUAs716tDxa2m2xHP:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L75)):*

Yes, this version looks cleaner to me.
</blockquote></details>

Done.
___
*[libSetReplace/HypergraphSubstitutionSystem.cpp, line 150 at r2](https://reviewable.io/reviews/maxitg/setreplace/647#-MaAbrex64ZB8Y3c87Vg:-MaCUD8SAS2ZOB_PK_wF:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/0fb1ff5df0d50e99a63a3eef8956b49402f5f519/libSetReplace/HypergraphSubstitutionSystem.cpp#L150)):*

I don't think we need a reference for what is essentially standard (even if complicated) C++.
</blockquote></details>

Done.

maxitg (changes requested):
> Reviewed 3 of 5 files at r3, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: all files reviewed, 3 unresolved discussions (waiting on @daneelsan)

___
*[libSetReplace/test/HypergraphSubstitutionSystem_test.cpp, line 5 at r5](https://reviewable.io/reviews/maxitg/setreplace/647#-MaE8mXl0J5zUzShCHY3:-MaE8mXl0J5zUzShCHY4:btgvap4) ([raw file](https://github.com/maxitg/setreplace/blob/57b825aaa59339a29ee01a1efc989228ca395371/libSetReplace/test/HypergraphSubstitutionSystem_test.cpp#L5)):*
> ```cpp
> #include <gtest/gtest.h>
> 
> #include <chrono>
> ```

You don't need to include <chrono> here since it's not used here explicitly.
___
*[libSetReplace/test/profile_tests.cpp, line 12 at r5](https://reviewable.io/reviews/maxitg/setreplace/647#-MaE8QKm4QNtnuqHGvba:-MaE8QKm4QNtnuqHGvbb:bvp45uw) ([raw file](https://github.com/maxitg/setreplace/blob/57b825aaa59339a29ee01a1efc989228ca395371/libSetReplace/test/profile_tests.cpp#L12)):*
> ```cpp
> constexpr auto doNotAbort = []() { return false; };
> 
> constexpr auto doNotTimeOut = []() { return false; };
> ```

This is no longer needed.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[libSetReplace/test/HypergraphSubstitutionSystem_test.cpp, line 5 at r5](https://reviewable.io/reviews/maxitg/setreplace/647#-MaE8mXl0J5zUzShCHY3:-MaEyci5C7LGOL4lcITr:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57b825aaa59339a29ee01a1efc989228ca395371/libSetReplace/test/HypergraphSubstitutionSystem_test.cpp#L5)):*

You don't need to include <chrono> here since it's not used here explicitly.
</blockquote></details>

Done.
___
*[libSetReplace/test/profile_tests.cpp, line 12 at r5](https://reviewable.io/reviews/maxitg/setreplace/647#-MaE8QKm4QNtnuqHGvba:-MaEyc9S57yvHMqdxpID:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/57b825aaa59339a29ee01a1efc989228ca395371/libSetReplace/test/profile_tests.cpp#L12)):*

This is no longer needed.
</blockquote></details>

Done.

maxitg (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/647#-:-MaFItqS2zqI9gQ3Ac8-:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @daneelsan)

PR #643, opened 2021-05-12
HypergraphSubstitutionSystem

daneelsan:
## Changes

* Reference resolved issues.
* Add a few comments about what this pull request changes.

## Comments

* Add comments for the reviewer.

## Examples

* Add explicit inputs and outputs (as screenshots in case of graphics) showcasing new functionality.

PR #657, opened 2021-06-16
Simplified generator system

maxitg:
## Changes

* This is a complete rewrite of the generator system.
* Old `GenerateMultihistory` syntax is removed.
* Instead, all generators have the format

  ```wl
  Generator[System[rules], param1 -> value1, ...] @ init
  ```

  Note that `param1`, etc. are symbols rather than strings. They have usage messages, etc. Also, one can use lists or associations instead, e.g.,

  ```wl
  Generator[System[rules], {param1 -> value1}, <|param2 -> value2|>, param3 -> value3] @ init
  ```

* Parameters are now declared separately from systems and generators.
* Systems need to declare a logical expression specifying which parameters can be specified. For example,

  ```wl
  Implies[MaxEvents || MaxDestroyerEvents, EventOrder]
  ```

  means that `EventOrder` needs to be specified if either `MaxEvents` or `MaxDestroyerEvents` are specified.

* Generators have predefined values for some parameters. E.g., `GenerateSingleHistory` sets `MaxDestroyerEvents -> 1`, which can no longer be changed.

## Comments

* Apologies for a huge PR. There is a lot of refactoring here as all instances of `GenerateMultihistory` had to be changed.
* Ordering functions page is deleted for now but should return as a page for the `EventOrder` parameter once it is used somewhere.
* @daneelsan, unfortunately, it will break #643, but on the flip side, it should make it a lot easier to define parameters (as one does not need to think where to put them anymore).

## Examples

* `GenerateSingleHistory`:

```wl
In[] := #["StatesList"] & @
 SetReplaceTypeConvert[{WolframModelEvolutionObject, 2}] @
  GenerateSingleHistory[MultisetSubstitutionSystem[{a_, b_} :> {a + b}], MaxEvents -> 4] @ {1, 2, 3}
Out[] = {{1, 2, 3}, {3, 3}, {6}}
```

* `GenerateAllHistories`:

```wl
In[] := #["ExpressionsEventsGraph", VertexLabels -> Placed[Automatic, After]] & @
 SetReplaceTypeConvert[{WolframModelEvolutionObject, 2}] @
  GenerateMultihistory[
    MultisetSubstitutionSystem[{a_, b_} /; a < b :> {a + b}], MaxGeneration -> 2] @ Range[3]
```
image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)


daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 7 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Documentation/Generators/MaxEventInputs.md, line 3 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXedCc-F7rtPHcWWLs:-MeXedCc-F7rtPHcWWLt:bests8s) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxEventInputs.md#L3)):*
> ```
> the max and min
> ```
I prefer `the maximum and minimum`
___
*[Documentation/Generators/MaxEventInputs.md, line 9 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXex3f9IgCQMV08Bk1:-MeXex3f9IgCQMV08Bk2:b-i9ib3s) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxEventInputs.md#L9)):*
> ```
> However, if the range of match sizes is known ahead of time, one can set it explicitly.
> ```
`... one can set it explicitly, to reduce/minimize the number of matchings generated` 
___
*[Documentation/Generators/MaxGeneration.md, line 7 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXg2II3n7cCRjuTO_1:-MeXg2II3n7cCRjuTO_2:b-hfgqtj) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L7)):*
> ```
> The generation of an event is defined as the maximum of the generations of its inputs plus one.
> ```
This means that matchings instantiations are sorted by the generations of the inputs? That is, I can't instantiate a match that has an input of generation 3, if there still remains a match with an input of generation 2.
___
*[Documentation/Generators/MaxGeneration.md, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXgqOEDMpGGfknlsCw:-MeXgqOEDMpGGfknlsCx:b-uepl4g) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L29)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
> In[] := #["ExpressionsEventsGraph",
>           VertexLabels -> Placed["Name", After, Replace[{{"Expression", n_} :> #["ExpressionGenerations"][[n]],
>                                                          {"Event", n_} :> #["EventGenerations"][[n]]}]]] & @
>   SetReplaceTypeConvert[{WolframModelEvolutionObject, 2}] @
>     GenerateMultihistory[MultisetSubstitutionSystem[{a__} /; Total[{a}] == 5 :> {Total[{a}] - 1, Total[{a}] + 1}],
>                          {1, 2, 3},
>                          MaxEvents -> 3]
> ```
</details>

We need a `VertexLabels -> "Generation"`
___
*[Documentation/Generators/MaxGeneration.md, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaFn0YEUBSCOx7S51T:-MeaFn0_03ggmaduAfFJ:bm9hrw5) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L38)):*
> ```
> Since `MaxGeneration` is a selection parameter rather than a stopping condition, it will continue evaluation even after
> encountering matches exceeding the generations constraint, which might also result in a different
> event order (i.e., some events being skipped) than if using, e.g., [`MaxEvents`](MaxEvents.md). For this reason,
> ```
Something like:
"MaxGeneration is an event (match?) selection parameter, not a stopping condition. That is, the evolution of the system won't stop 
if a match with a generation greater than the constraint is encountered. Rather, it will ignore those matches and instantiate only those
that satisfy the generation constraint. This selection of matching (events?) according to their generation will (in most cases) affect the 
ordering of events, as opposed to using stopping conditions, e.g. MaxEvents".

PS: I'm unsure how to call the "generation of a match". "Generation index" of a match?
___
*[Documentation/Generators/README.md, line 13 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKB9CAw-dDgvNtZmu:-MeaKB9CAw-dDgvNtZmv:b-lqzwyl) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L13)):*
> ```
> which is essential when the system does not terminate
> ```
Why is it essential?
___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-MeaKV9d-KMZvSrlNMve:bf5e253) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*
> ```Markdown
> 
> ```wl
> Generator[System[rules], init, parameters...]
> ```

Why did you decide to support this syntax?

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 10 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/A1$generatorSystem.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaVsy505TMQr1bt1Gm:-MeaVsy68FpE7JaF5pAA:bcw48pk) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L26)):*
> ```
> Parameters in the fourth argument should be declared with declareSystemParameter.
> ```
So, call declareSystemParameter _before_ declareSystem
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-MeadJhs8eDtebPXAyo0:b-5xb1a) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*
> ```objc
>                    _ ? FailureQ,
>                    message[generator, #, <|"expr" -> HoldForm[expr]|>] &];
>     result /; !FailureQ[result]
> ```

Recently, Leonid gave a couple of talks about error handling in WL. He wen't over almost all conventions, and concluded that messages and unevaluated expressions are really not the way to go. Leaving the input unevaluated was fine when we dealt with symbolic manipulation, which is not the case here. Messages should probably only be used to issue warnings, not errors. Don't you think a workflow that only deals with Failure objects would be easier to implement and design?
___
*[Kernel/systemParameters.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaejSaC2pF1QN37vv7:-MeaejSaC2pF1QN37vv8:b8ydp52) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/systemParameters.m#L26)):*
> ```objc
> (* Stopping-condition parameters *)
> 
> PackageExport["MaxEvents"]
> ```

So now `Package*` are no longer required to be at the top of the file? :eyes:

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 10 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/Generators/MaxEventInputs.md, line 3 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXedCc-F7rtPHcWWLs:-MeeDXz3AliBu2RQPvNt:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxEventInputs.md#L3)):*

> ```
> the max and min
> ```
I prefer `the maximum and minimum`
</blockquote></details>

Done.
___
*[Documentation/Generators/MaxEventInputs.md, line 9 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXex3f9IgCQMV08Bk1:-MeeE1Gk-vwfKeIPi36g:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxEventInputs.md#L9)):*

> ```
> However, if the range of match sizes is known ahead of time, one can set it explicitly.
> ```
`... one can set it explicitly, to reduce/minimize the number of matchings generated` 
</blockquote></details>

Done.
___
*[Documentation/Generators/MaxGeneration.md, line 7 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXg2II3n7cCRjuTO_1:-MeeEFPh8F1TKkho15_A:b8ih236) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L7)):*

> ```
> The generation of an event is defined as the maximum of the generations of its inputs plus one.
> ```
This means that matchings instantiations are sorted by the generations of the inputs? That is, I can't instantiate a match that has an input of generation 3, if there still remains a match with an input of generation 2.

</blockquote></details>

Not if events are independent:

```wl
In[] := Normal @ GenerateSingleHistory[
  MultisetSubstitutionSystem[{{1} -> {2}, {2} -> {5}, {3, 4} -> {6}}], {1, 2, 3, 4}][[2, "EventGenerations"]]
Out[] = {0, 1, 1, 2, 1}
```

Or, by explicitly setting the event order:

```wl
In[] := WolframModel[{{1, 2}, {2, 3}} -> {{1, 3}},
                     {{a, b}, {b, c}, {1, 2}, {2, 3}, {3, 4}},
                     Infinity,
                     "EventOrderingFunction" -> {"NewestEdge"}]["EventGenerations"]
Out[] = {1, 2, 1}
```
`"EventOrderingFunction" -> Automatic` has this property though by design.
___
*[Documentation/Generators/MaxGeneration.md, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXgqOEDMpGGfknlsCw:-MeeH8uI8YItcZoW-Srf:b-38l9e6) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L29)):*

> ```
> In[] := #["ExpressionsEventsGraph",
>           VertexLabels -> Placed["Name", After, Replace[{{"Expression", n_} :> #["ExpressionGenerations"][[n]],
>                                                          {"Event", n_} :> #["EventGenerations"][[n]]}]]] & @
>   SetReplaceTypeConvert[{WolframModelEvolutionObject, 2}] @
>     GenerateMultihistory[MultisetSubstitutionSystem[{a__} /; Total[{a}] == 5 :> {Total[{a}] - 1, Total[{a}] + 1}],
>                          {1, 2, 3},
>                          MaxEvents -> 3]
> ```
We need a `VertexLabels -> "Generation"`
</blockquote></details>

I was just working on it yesterday:

```wl
In[] := TokenEventGraph[VertexLabels -> (#Generation &)] @
  GenerateMultihistory[MultisetSubstitutionSystem[{a__} /; Total[{a}] == 5 :> {Total[{a}] - 1, Total[{a}] + 1}],
                       {1, 2, 3},
                       MaxEvents -> 3]
```



Still trying to figure out what the API should be.
___
*[Documentation/Generators/MaxGeneration.md, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaFn0YEUBSCOx7S51T:-MeeKhgr8k0UnXvuUfql:b-7wrlch) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L38)):*

> ```
> Since `MaxGeneration` is a selection parameter rather than a stopping condition, it will continue evaluation even after
> encountering matches exceeding the generations constraint, which might also result in a different
> event order (i.e., some events being skipped) than if using, e.g., [`MaxEvents`](MaxEvents.md). For this reason,
> ```
Something like:
"MaxGeneration is an event (match?) selection parameter, not a stopping condition. That is, the evolution of the system won't stop 
if a match with a generation greater than the constraint is encountered. Rather, it will ignore those matches and instantiate only those
that satisfy the generation constraint. This selection of matching (events?) according to their generation will (in most cases) affect the 
ordering of events, as opposed to using stopping conditions, e.g. MaxEvents".

PS: I'm unsure how to call the "generation of a match". "Generation index" of a match?
</blockquote></details>

Yes, that is much more clear. What's wrong with "generation of a match"? A match is just a "tentative" event, it has all the same properties except for output tokens.
___
*[Documentation/Generators/README.md, line 13 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKB9CAw-dDgvNtZmu:-MeeLL1u5hMCYGT5x-Yo:b-yemntl) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L13)):*

> ```
> which is essential when the system does not terminate
> ```
Why is it essential?
</blockquote></details>

Because if the system does not terminate and no parameters are specified, it's going to run forever. I clarified the explanation a bit.
___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-MeeMS3e2bv05dBAgMhj:bp8loyv) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*

Why did you decide to support this syntax?
</blockquote></details>

It just seemed awkward to use operator form in cases where the init is very small. Especially in cases like these:

```wl
GenerateMultihistory[MultisetSubstitutionSystem[{{} -> {1, 2}, {a_, b_} :> {a + b}}]] @ {}
```

However, it does cause problems as the parameters might get interpreted as an init if the pattern for that init is too general (like in the case of `AtomicStateSystem`). So what do you think, should I remove it?
___
*[Kernel/A1$generatorSystem.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaVsy505TMQr1bt1Gm:-MeeOQma9zOnduvH4ZcG:b554tad) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L26)):*

> ```
> Parameters in the fourth argument should be declared with declareSystemParameter.
> ```
So, call declareSystemParameter _before_ declareSystem
</blockquote></details>

They are called in separate files, but I reordered the definitions in this file if that's what you meant.
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-Mef1ZeaAy4aUplAIyJN:b-dupf1p) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*

Recently, Leonid gave a couple of talks about error handling in WL. He wen't over almost all conventions, and concluded that messages and unevaluated expressions are really not the way to go. Leaving the input unevaluated was fine when we dealt with symbolic manipulation, which is not the case here. Messages should probably only be used to issue warnings, not errors. Don't you think a workflow that only deals with Failure objects would be easier to implement and design?
</blockquote></details>

Yes, that does seem like a good idea. All other languages seem to be doing this as well. In fact, that's what we are doing internally precisely because it's more convenient. It might also be helpful outside of the package. Plus, it will indeed simplify things a lot because we won't need to keep two separate versions for every function.

Although I think instead of returning Failure objects, we should throw exceptions.

It's very unconventional in WL, but I don't see what the downsides would be. @taliesinb, what do you think?

PS. Not in this PR, though, as we need to make it consistent across all functions.
___
*[Kernel/systemParameters.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaejSaC2pF1QN37vv7:-Mef2fDb1kbLyvEAzn6g:bnj57p4) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/systemParameters.m#L26)):*

So now `Package*` are no longer required to be at the top of the file? :eyes:
</blockquote></details>

Good catch! I moved it to the top.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 6 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Documentation/Generators/MaxGeneration.md, line 7 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXg2II3n7cCRjuTO_1:-Meh71-CBlq7RSS6el6A:bbet5mp) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L7)):*
> Normal @ GenerateSingleHistory[
>   MultisetSubstitutionSystem[{{1} -> {2}, {2} -> {5}, {3, 4} -> {6}}], {1, 2, 3, 4}][[2, "EventGenerations"]]
So it instantiates any match that has a less-or-equal than the max generation constraint, not necessarily following a low-to-high generation order. That makes sense.
___
*[Documentation/Generators/MaxGeneration.md, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXgqOEDMpGGfknlsCw:-Meh6Xox9oK616MIDnHD:b-6p0e4g) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L29)):*

I was just working on it yesterday:

```wl
In[] := TokenEventGraph[VertexLabels -> (#Generation &)] @
  GenerateMultihistory[MultisetSubstitutionSystem[{a__} /; Total[{a}] == 5 :> {Total[{a}] - 1, Total[{a}] + 1}],
                       {1, 2, 3},
                       MaxEvents -> 3]
```



Still trying to figure out what the API should be.
</blockquote></details>

Interesting... _The user is filled with power_
___
*[Documentation/Generators/MaxGeneration.md, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaFn0YEUBSCOx7S51T:-Meh7u022sI5fAAF2A_c:bvz24os) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L38)):*

Yes, that is much more clear. What's wrong with "generation of a match"? A match is just a "tentative" event, it has all the same properties except for output tokens.
</blockquote></details>

I'm just unsure whether we've explained what the difference between a match and an event is. I like "tentative" event. TODO event, maybe?
___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-Meh5Bm-2G2mdjwsaPzU:b-3f6zij) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*

It just seemed awkward to use operator form in cases where the init is very small. Especially in cases like these:

```wl
GenerateMultihistory[MultisetSubstitutionSystem[{{} -> {1, 2}, {a_, b_} :> {a + b}}]] @ {}
```

However, it does cause problems as the parameters might get interpreted as an init if the pattern for that init is too general (like in the case of `AtomicStateSystem`). So what do you think, should I remove it?
</blockquote></details>

I think we should go for the easier thing to implement. I like the idea of there only being one way to generate a multihistory: `init // Generator[...]`. It clearly separates the rule and the system parameters from a specific initial condition. Ideally `Generator[...]` should parse and process its contents without the need of specifying an initial condition, so `{init1, init2} // Map[Generator[...]]` only process Generator once, but I'm unsure how to go about it.
___
*[Kernel/A1$generatorSystem.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaVsy505TMQr1bt1Gm:-Meh51Vl0OYmN8RkA5Hi:bhf5j4k) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L26)):*

They are called in separate files, but I reordered the definitions in this file if that's what you meant.
</blockquote></details>

I actually meant that to define a system you'd have to first call defineSystemParameter, but now I get it.
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-Meh4UzRCKEbmb12mffL:b-7gjkg9) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*

Yes, that does seem like a good idea. All other languages seem to be doing this as well. In fact, that's what we are doing internally precisely because it's more convenient. It might also be helpful outside of the package. Plus, it will indeed simplify things a lot because we won't need to keep two separate versions for every function.

Although I think instead of returning Failure objects, we should throw exceptions.

It's very unconventional in WL, but I don't see what the downsides would be. @taliesinb, what do you think?

PS. Not in this PR, though, as we need to make it consistent across all functions.
</blockquote></details>

If by throwing you mean throwing and not catching, I'm unsure :thinking: 
 
I agree this is not the PR to do that (it would be much bigger than it already is!)

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 6 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/Generators/MaxGeneration.md, line 7 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXg2II3n7cCRjuTO_1:-MejIHh1820VLw1lomMb:btih660) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L7)):*

> Normal @ GenerateSingleHistory[
>   MultisetSubstitutionSystem[{{1} -> {2}, {2} -> {5}, {3, 4} -> {6}}], {1, 2, 3, 4}][[2, "EventGenerations"]]
So it instantiates any match that has a less-or-equal than the max generation constraint, not necessarily following a low-to-high generation order. That makes sense.
</blockquote></details>

Yes, exactly.
___
*[Documentation/Generators/MaxGeneration.md, line 29 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeXgqOEDMpGGfknlsCw:-Mek2DfS1anw40Mh9DgC:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L29)):*

Interesting... _The user is filled with power_
</blockquote></details>

Done.
___
*[Documentation/Generators/MaxGeneration.md, line 38 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaFn0YEUBSCOx7S51T:-MejJVA8AdLJpYD6HR0P:b5gcbrz) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/MaxGeneration.md#L38)):*

I'm just unsure whether we've explained what the difference between a match and an event is. I like "tentative" event. TODO event, maybe?
</blockquote></details>

Ok, added "tentative event" as clarification.
___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-Mejn11lEbvOaY6EJUuR:b-o4cems) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*

I think we should go for the easier thing to implement. I like the idea of there only being one way to generate a multihistory: `init // Generator[...]`. It clearly separates the rule and the system parameters from a specific initial condition. Ideally `Generator[...]` should parse and process its contents without the need of specifying an initial condition, so `{init1, init2} // Map[Generator[...]]` only process Generator once, but I'm unsure how to go about it.
</blockquote></details>

Ok, I removed it for now, mainly because parameters can be unintentionally matched to inits. And it seems that it is more natural to think about parameters before the init, so the order of arguments in the operator form seems to make more sense.

It still seems more convenient to use the non-operator form in tests, but it will be the reverse once we allow existing multihistories to be continued.

Also, what kind of preprocessing do you think it should do?
___
*[Kernel/A1$generatorSystem.m, line 26 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaVsy505TMQr1bt1Gm:-MejtMSECaPX7OOeVqCO:b5tn9sg) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L26)):*

I actually meant that to define a system you'd have to first call defineSystemParameter, but now I get it.
</blockquote></details>

No, the order does not matter because all non-trivial stuff happens in `initializeSystemGenerators` and in calls to functions.
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-MejtXnyAhcoODnEg8W8:bt6lrl2) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*

If by throwing you mean throwing and not catching, I'm unsure :thinking: 
 
I agree this is not the PR to do that (it would be much bigger than it already is!)
</blockquote></details>

Here is why it might be a good idea to throw and not catch. Suppose someone wants to implement a function that takes multiset rules as an argument. Of course, they would not want to implement a check themselves, so in the current API, they might do something like this:

```wl
runOnRange::invalidMultisetRules = "Rules `1` are invalid.";

runOnRange[rules_] := ModuleScope[
  result = Quiet[
    Check[GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10],
          $Failed,
          GenerateSingleHistory::invalidMultisetRules],
    GenerateSingleHistory::invalidMultisetRules];
  If[result === $Failed, Message[runOnRange::invalidMultisetRules, rules]];
  Normal @ (#[[2, "Expressions"]] &) @ result /; result =!= $Failed
];
```

Pretty complicated. If we return a `Failure`, it becomes a bit better:

```wl
runOnRange[rules_] := ModuleScope[
  result = GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10];
  If[FailureQ[result], result, Normal @ (#[[2, "Expressions"]] &) @ result]
];
```

However, if we throw, we don't need to do anything at all:

```wl
runOnRange[rules_] :=
  Normal @ (#[[2, "Expressions"]] &) @ GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10];
```

since the evaluation is going to stop immediately once the throw happens. One case I can think of where this might be unwanted is if one does a large enumeration of, say, rules and does not care if a few of them fail because the goal is to find something, and getting all results is not important. However, I would argue that one should be deliberate in these cases, and it is very easy to fix it by doing a `Catch @` in front of an exception-throwing function.

daneelsan (approved):
Looks good to me. I now understand how much `Generator[System[rules], param1 -> value1, ...] @ init` reduces the time we spend typing. 

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657#-:-Mepw2XF1vDlZX8Cmx31:b-k2os2f)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-MepgXsT7taazC2q6fIa:b-ah6zlw) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*

Ok, I removed it for now, mainly because parameters can be unintentionally matched to inits. And it seems that it is more natural to think about parameters before the init, so the order of arguments in the operator form seems to make more sense.

It still seems more convenient to use the non-operator form in tests, but it will be the reverse once we allow existing multihistories to be continued.

Also, what kind of preprocessing do you think it should do?
</blockquote></details>

The idea I had in mind was to use no entry to avoid parsing the arguments every time:
```wl
In[]:= ClearAll[f]
f[x_] ? System`Private`HoldEntryQ := (
   Print["Processing x..."];
   If[x > 0,
    System`Private`ConstructNoEntry[f, # + x &],
    $Failed
    ]
   );

In[]:= func = f[3]
During evaluation of In[141]:= Processing x...
Out[]= f[#1 + 3 &]

In[]:= func
Out[]= f[#1 + 3 &]
```
But now that I think about it, if the mapped function is not a pure function then it will be processed only once:
```wl
In[]:= g[3] /@ {1, 2, 3}
During evaluation of In[148]:= Processing x...
Out[]= {g[#1 + 3 &][1], g[#1 + 3 &][2], g[#1 + 3 &][3]}
```

So it doesn't matter. I think NoEntry will be useful when the head is the same as the constructor, e.g. Graph, SparseArray, etc.
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-MeptJrO-mlCnT1__pgf:breuk4n) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*

Here is why it might be a good idea to throw and not catch. Suppose someone wants to implement a function that takes multiset rules as an argument. Of course, they would not want to implement a check themselves, so in the current API, they might do something like this:

```wl
runOnRange::invalidMultisetRules = "Rules `1` are invalid.";

runOnRange[rules_] := ModuleScope[
  result = Quiet[
    Check[GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10],
          $Failed,
          GenerateSingleHistory::invalidMultisetRules],
    GenerateSingleHistory::invalidMultisetRules];
  If[result === $Failed, Message[runOnRange::invalidMultisetRules, rules]];
  Normal @ (#[[2, "Expressions"]] &) @ result /; result =!= $Failed
];
```

Pretty complicated. If we return a `Failure`, it becomes a bit better:

```wl
runOnRange[rules_] := ModuleScope[
  result = GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10];
  If[FailureQ[result], result, Normal @ (#[[2, "Expressions"]] &) @ result]
];
```

However, if we throw, we don't need to do anything at all:

```wl
runOnRange[rules_] :=
  Normal @ (#[[2, "Expressions"]] &) @ GenerateSingleHistory[MultisetSubstitutionSystem[rules]] @ Range[10];
```

since the evaluation is going to stop immediately once the throw happens. One case I can think of where this might be unwanted is if one does a large enumeration of, say, rules and does not care if a few of them fail because the goal is to find something, and getting all results is not important. However, I would argue that one should be deliberate in these cases, and it is very easy to fix it by doing a `Catch @` in front of an exception-throwing function.
</blockquote></details>

Definitely, the first check sucks.

I like the second option, it's explicit. Although this will appear everywhere: `result = ...; If[FailureQ[result], result, func[result]]`

I've been playing with `Confirm`/`Enclose` quite a bit. Using those, it would look like:
```wl
runOnRange[rules_] :=
  Enclose[Normal @ (#[[2, "Expressions"]] &) @ Confirm[GenerateSingleHistory[MultisetSubstitutionSystem[rules]]] @ Range[10], "Expression"]
```

There are definitely some issues with the confirm/enclose approach. For instance, there is not a explicit way of throwing something, i.e. `Throw[failure]`.
You have to do `Confirm[$Failed, failure]`. Also, I remember an email thread comparing the performance of confirm/enclose vs catch/throw and return, and confirm/enclose was slower. I think it's to be expected as confirm/enclose is a wrapper over catch/throw...

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/657)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/Generators/README.md, line 48 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeaKV9d-KMZvSrlNMvd:-MeyxU94DpTy_qbSwMc8:b-ba81gw) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Documentation/Generators/README.md#L48)):*

The idea I had in mind was to use no entry to avoid parsing the arguments every time:
```wl
In[]:= ClearAll[f]
f[x_] ? System`Private`HoldEntryQ := (
   Print["Processing x..."];
   If[x > 0,
    System`Private`ConstructNoEntry[f, # + x &],
    $Failed
    ]
   );

In[]:= func = f[3]
During evaluation of In[141]:= Processing x...
Out[]= f[#1 + 3 &]

In[]:= func
Out[]= f[#1 + 3 &]
```
But now that I think about it, if the mapped function is not a pure function then it will be processed only once:
```wl
In[]:= g[3] /@ {1, 2, 3}
During evaluation of In[148]:= Processing x...
Out[]= {g[#1 + 3 &][1], g[#1 + 3 &][2], g[#1 + 3 &][3]}
```

So it doesn't matter. I think NoEntry will be useful when the head is the same as the constructor, e.g. Graph, SparseArray, etc.
</blockquote></details>

Yes, in this case, it's not processed at all until the operator has an argument. And yes, NoEntry should be useful in cases like `Hypergraph`.
___
*[Kernel/A1$generatorSystem.m, line 183 at r2](https://reviewable.io/reviews/maxitg/setreplace/657#-MeadJhL4tJ_BuUwbYqW:-MeyxuaF4KoWO7ZoMLN7:bzhkyia) ([raw file](https://github.com/maxitg/setreplace/blob/6fc5f48fb18aaa28529d5d7c5b2ce7aa23b3be8d/Kernel/A1$generatorSystem.m#L183)):*

Definitely, the first check sucks.

I like the second option, it's explicit. Although this will appear everywhere: `result = ...; If[FailureQ[result], result, func[result]]`

I've been playing with `Confirm`/`Enclose` quite a bit. Using those, it would look like:
```wl
runOnRange[rules_] :=
  Enclose[Normal @ (#[[2, "Expressions"]] &) @ Confirm[GenerateSingleHistory[MultisetSubstitutionSystem[rules]]] @ Range[10], "Expression"]
```

There are definitely some issues with the confirm/enclose approach. For instance, there is not a explicit way of throwing something, i.e. `Throw[failure]`.
You have to do `Confirm[$Failed, failure]`. Also, I remember an email thread comparing the performance of confirm/enclose vs catch/throw and return, and confirm/enclose was slower. I think it's to be expected as confirm/enclose is a wrapper over catch/throw...
</blockquote></details>

What are the downsides of throwing?

PR #662, opened 2021-07-13
Type versions

maxitg:
## Changes

* Expands `SetReplaceType` to now accept up to two arguments: `SetReplaceType[name, version]`.
* It is now used not only in type graphs but everywhere: `SetReplaceObjectType` returns it, `Multihistory` types use it, etc.
* `objectType` can return types wrapped in `SetReplaceType` or not. However, types will be wrapped automatically once they reach any user-facing functions or other declarations such as `declareTypeTranslation` (in other words, `SetReplaceType[name]` can be used interchangeably with `name`).
* There are custom boxes for `SetReplaceType`: image
* `SetReplaceTypeConvert` can accept types without versions, e.g., `SetReplaceTypeConvert[WolframModelEvolutionObject] @ object`. In this case, the latest known type is used.

## Comments

* It is a bit awkward (although somewhat unrelated to this PR) that multihistory types use the same symbols as corresponding systems. One would think it would make more sense to use types like `MultisetSubstitutionMultihistory`, etc.

  However, if we do that, `Multihistory` will look weird because the word 'multihistory' will be repeated twice: in the `Head` and in the top raw of the object details:
  image

  We can alternatively move that name to the 'Head' part
  image
  but I think we want to keep the head in boxes consistent with the real head.

  If we then do in fact use different heads for different multihistories, it will make them harder to implement, since one will have to implement `objectType`, usage, etc. for each one.

  So, given all that, I'm keeping the type names the same.

* [Done] Should we add any extra styling to the boxes so that it does not look exactly like the original type name, especially if no version is specified?

  image

## Examples

* Types now have the new formatting everywhere:

  image

* Even in `$SetReplaceTypeGraph`:

  ```wl
  In[] := $SetReplaceTypeGraph
  ```

  image

* And conversions can be done without specifying a version:

  image

* Although specifying the version explicitly is still allowed:

  image

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662)


maxitg:
@daneelsan, @taliesinb, this is rebased and ready for review.

daneelsan (changes requested):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 30 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKANOB859siohIZ1FN:-MfKANOC5utAiWETF6CY:b-fmbgs1) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L30)):*
> ```objc
> SyntaxInformation[SetReplaceType] = {"ArgumentsPattern" -> {name_, version_.}};
> 
> SetReplaceType /: MakeBoxes[expr : SetReplaceType[name_], form : StandardForm | TraditionalForm] := With[{
> ```

I wonder if `SetReplaceType` should parse its arguments, e.g. verify it is an already defined `SetReplace` type, or
check if the version exists:

For example:
```wl
In[]:= SetReplaceType["Not a type"] // ToString
Out[]= "SetReplaceType[Not a type]"
```
Return `Failure` instead?

```
In[]:= SetReplaceType[WolframModelEvolutionObject] // ToString
Out[]= "SetReplaceType[WolframModelEvolutionObject]"
```
Parse to `SetReplaceType[WolframModelEvolutionObject, 2]`?

Basically, `SetReplaceType` would be an object with a "constructor".
___
*[Kernel/A1$typeSystem.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKBNVF4G-m3JJhKB9d:-MfKBNVG2Xq_oZcqAlgQ:b-1418qd) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L39)):*
> ```objc
>     nameBoxes = MakeBoxes[name, form],
>     versionString = "v" <> ToString @ version},
>   With[{boxes = RowBox[{nameBoxes, "\[ThickSpace]", StyleBox[versionString, FontColor -> Gray]}]},
> ```

I don't particularly like this format, though I understand why it is that way: You want for these types to look nice and compact when they are inside the `Multihistory` object.

Perhaps a tiny box, like `Iconize`?

daneelsan (commented):
I don't think `SetReplaceType[system]` where system is the same as `GenerateMultihistory[system[...], ...]` is awkward. 

You say that because the result from calling `SetReplaceObjectType[multihistory]` should say the object is a multihistory in some way, not just the name of the system that generated that multihistory?

I guess it would be confusing if there were a `singlehistory` object, then `SetReplaceObjectType[singlehistory]`would return the same exact type as a multihistory object...

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662#-:-MfKCfqe6BPOYW4CFe1P:b-ad04r9)* status: 0 of 34 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

maxitg (commented):
The awkward part is that if we ever need `SetReplaceObjectType[MultisetSubstitutionSystem[...]]`, it would need to return `MultisetSubstitutionSystem` instead.

Also, I don't think calling a single history multihistory is a problem. It is a multihistory in the same way as a set is a multiset or an integer is a real.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662#-:-MfOpfBsBqg03_T0V9fM:bklk6ty)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Kernel/A1$typeSystem.m, line 30 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKANOB859siohIZ1FN:-MfOqPXJ-UIx1JWlwaGi:bzz85x) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L30)):*

I wonder if `SetReplaceType` should parse its arguments, e.g. verify it is an already defined `SetReplace` type, or
check if the version exists:

For example:
```wl
In[]:= SetReplaceType["Not a type"] // ToString
Out[]= "SetReplaceType[Not a type]"
```
Return `Failure` instead?

```
In[]:= SetReplaceType[WolframModelEvolutionObject] // ToString
Out[]= "SetReplaceType[WolframModelEvolutionObject]"
```
Parse to `SetReplaceType[WolframModelEvolutionObject, 2]`?

Basically, `SetReplaceType` would be an object with a "constructor".
</blockquote></details>

Where do you think it will be useful? The difficulty here is that types are not really defined. Instead, they are obtained from conversions and property declarations. So, it is possible in principle for `objectType` to return a type that is not otherwise recognized. It is, of course, possible to fix that by declaring types, but it seems like it will complicate things, and I'm not sure what the benefit would be since the types will only need to be manually specified for `SetReplaceTypeConvert`, where they will be parsed anyway.
___
*[Kernel/A1$typeSystem.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKBNVF4G-m3JJhKB9d:-MfPJNGdC-WEGDQuZk_8:b-4ut2gp) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L39)):*

I don't particularly like this format, though I understand why it is that way: You want for these types to look nice and compact when they are inside the `Multihistory` object.

Perhaps a tiny box, like `Iconize`? 


</blockquote></details>

Ok, how about this? It's consistent with the vertex style in the type graph (and inspired by `Entity`):

```wl
In[] := SetReplaceType[AtomicStateSystem, 0]
```



I kept the previous one as `TraditionalForm` (there does not seem to be any other forms available nor is it possible to define new forms) because the new one looks ridiculous in the multihistory box and in the type graph.

daneelsan (commented):
I agree, I take back what I said.
`SetReplaceObjectType[MultisetSubstitutionSystem[...]]`, don't know where that would come.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662#-:-MfYUvpo5EkLxsFzCNNA:bxmk0g5)* status: 0 of 35 files reviewed, 3 unresolved discussions (waiting on @maxitg and @taliesinb)

___
*[Documentation/Types/README.md, line 8 at r3](https://reviewable.io/reviews/maxitg/setreplace/662#-MfYUDu9EG5NSpnJfu-S:-MfYUDu9EG5NSpnJfu-T:b-5w0lwk) ([raw file](https://github.com/maxitg/setreplace/blob/7e7703ce86ab4777c828a9693ffa502483324468/Documentation/Types/README.md#L8)):*
> ```Markdown
> [properties](/Documentation/Properties/README.md) from that data.
> 
> Multihistory types are named after systems that produce them, e.g., `SetReplaceType[MultisetSubstitutionSystem, 1]`.
> ```

Was the `1` here intentional?
___
*[Kernel/A1$typeSystem.m, line 30 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKANOB859siohIZ1FN:-MfYUNcW2uPrwppnbmsA:bcwsnl) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L30)):*

Where do you think it will be useful? The difficulty here is that types are not really defined. Instead, they are obtained from conversions and property declarations. So, it is possible in principle for `objectType` to return a type that is not otherwise recognized. It is, of course, possible to fix that by declaring types, but it seems like it will complicate things, and I'm not sure what the benefit would be since the types will only need to be manually specified for `SetReplaceTypeConvert`, where they will be parsed anyway.
</blockquote></details>

Yes, if it's only used for `SetReplaceTypeConvert` then it doesn't seem necessary...
... I guess it could matter if we want to allow users to construct `Multihistory` objects manually (which I don't think it's a good idea).
___
*[Kernel/A1$typeSystem.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKBNVF4G-m3JJhKB9d:-MfYSaCk3Lr2jSBDq0kH:b-x1rtws) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L39)):*

Ok, how about this? It's consistent with the vertex style in the type graph (and inspired by `Entity`):

```wl
In[] := SetReplaceType[AtomicStateSystem, 0]
```



I kept the previous one as `TraditionalForm` (there does not seem to be any other forms available nor is it possible to define new forms) because the new one looks ridiculous in the multihistory box and in the type graph.
</blockquote></details>

I really like it :)

daneelsan (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662#-:-MfYZmY7EBUBmqp-ZLXL:bg7igkv)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @maxitg and @taliesinb)

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

___
*[Documentation/Types/README.md, line 8 at r3](https://reviewable.io/reviews/maxitg/setreplace/662#-MfYUDu9EG5NSpnJfu-S:-MfZ2zyuEsqMnZbSEic8:buj73yn) ([raw file](https://github.com/maxitg/setreplace/blob/7e7703ce86ab4777c828a9693ffa502483324468/Documentation/Types/README.md#L8)):*

Was the `1` here intentional?
</blockquote></details>

Kind of (it's going to be version 1 by release), but I changed it back to 0 for now to avoid confusion.
___
*[Kernel/A1$typeSystem.m, line 30 at r1](https://reviewable.io/reviews/maxitg/setreplace/662#-MfKANOB859siohIZ1FN:-MfZ36Os7Ri4A8z1KJmr:b-rkiv8o) ([raw file](https://github.com/maxitg/setreplace/blob/384cbc35ebb34448c4f475de803682d5c30a4432/Kernel/A1$typeSystem.m#L30)):*

Yes, if it's only used for `SetReplaceTypeConvert` then it doesn't seem necessary...
... I guess it could matter if we want to allow users to construct `Multihistory` objects manually (which I don't think it's a good idea).
</blockquote></details>

Even if we allow users to construct `Multihistory`, they would need to do it with a generator, which will automatically add the type (or, at least, parse it).

daneelsan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/662#-:-MfZARSaD_oPFACEygAF:bnfp4nl)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @daneelsan and @taliesinb)

Issue #494, opened 2020-10-26
Hypergraph head


maxitg:
### The problem
Is your feature request related to a problem? Please describe. Ex. I'm always frustrated when [...]

maxitg:
The box should be some kind of an interpretation box (ideally with a purple frame like `Graph` if possible). Otherwise, if you copy it and use it as an input, it does not work.

And yes, the format should be `Hyperedge[{vertices...}]` because, as you say, we need the second argument for other functionality.

daneelsan:
Yes, that was really a quick implementation. As you said, the idea would be to `TimeConstrain` it and show the Graph-like box if it takes too much time.

maxitg:
We can probably estimate how long it would take and constrain based on the number of vertices/hyperedges.

taliesinb:
I'd like to suggest the following syntax for when it comes time to implement symmetries on hyperedges:

* DirectedHyperedge[{1, 2, 3}, {4, 5, 6}]
* UndirectedHyperedge[{1, 2, 3}, {4, 5, 6}] (which is the same as UndirectedHyperedge[{4, 5, 6}, {1, 2, 3}])
* ColoredHyperedge9[{1 → 1, 2 → 1, 3 → 2, 4 → 2}] (which is ColoredHyperedge[{1, 2, 3, 4} → {1, 2, 2}])

These all evaluate to Hyperedge[vertexlist, symmetrygroup].
The symmetry for colors is fully symmetric *within each color*. 
The symmetry for directed and undirected hyperedges is fully symmetric within the source, and the destination separately.

PR #664, opened 2021-07-23
Adding Hypergraph object

daneelsan:
## Changes

* Trying to solve #494.
* Adding a basic `Hypergraph` object, accessors and upvalues.

## Comments

* Various designs questions:

  - How to represent the `Hypergraph` object in a notebook? Interpretation boxes (e.g. SparseArrays) or Graph-like plots?
  - The symbol `Hyperedge` is private for now. Should this be exposed?
  - Should we consider hypergraphs with mixed types of `Hyperedges`? See https://github.com/maxitg/SetReplace/issues/494#issuecomment-720580460 

* "ordered" was added (needs to be discussed) because many functions in WFR made the distinction. See:
   - https://resources.wolframcloud.com/FunctionRepository/resources/IsomorphicOrderedHypergraphQ/
   - https://resources.wolframcloud.com/FunctionRepository/resources/FindOrderedHypergraphIsomorphism/

## ToDo

- [ ] Tests
- [ ] Documentation
- [ ] More upvalues and/or redesign other functions to return `Hypergraph` (e.g. `RandomHypergraph`)

## Examples

* Create a random `Hypergraph`:
```wl
In[] := {8, 4} // RandomHypergraph // Hypergraph
```
HypergraphExample1

* Create an "ordered" `Hypergraph`:
```wl
In[] := Hypergraph[{{1, 1, 1}}, True]
```
HypergraphExample2

### DIFF
--- a/Kernel/Hypergraph.m
+++ b/Kernel/Hypergraph.m
@@ -1,0 +1,126 @@
+Package["SetReplace`"]
+
+PackageImport["GeneralUtilities`"]
+
+PackageExport["Hypergraph"]
+PackageExport["HypergraphQ"]
+PackageExport["HypergraphOrderedQ"]
+
+SetRelatedSymbolGroup[Hypergraph, HypergraphQ, HypergraphOrderedQ, EdgeList, VertexList];
+
+(* HypergraphQ *)
+
+SetUsage @ "HypergraphQ[hg$] yields True if hg$ is a valid Hypergraph object and False otherwise.";
+
+SyntaxInformation[HypergraphQ] = {"ArgumentsPattern" -> {expr_}};
+
+HypergraphQ[hypergraph_Hypergraph] := System`Private`HoldNoEntryQ[hypergraph];
+HypergraphQ[_] = False;
+
+(* HypergraphOrderedQ *)
+
+SetUsage @ "HypergraphOrderedQ[hg$] yields True if hg$ is a ordered Hypergraph object and False otherwise.";
+
+SyntaxInformation[HypergraphOrderedQ] = {"ArgumentsPattern" -> {hypergraph_}};
+
+HypergraphOrderedQ[HoldPattern[Hypergraph[_, orderedQ_] ? HypergraphQ]] := orderedQ;
+
+HypergraphOrderedQ[_] = False;
+
+(* Hypergraph *)
+
+SetUsage @ "
+Hypergraph[{he$1, he$2, $$}] yields a hypergraph with hyperedges he$j.
+Hypergraph[$$, ord$] returns an ordered hypergraph if ord$ is True, and an unordered hypergraph if False.
+";
+
+SyntaxInformation[Hypergraph] = {"ArgumentsPattern" -> {hyperedges_, orderedQ_.}};
+
+Hypergraph[hyperedges_, orderedQ : (True | False) : False] ? System`Private`HoldEntryQ :=
+  If[hypergraphQ[hyperedges],
+    System`Private`ConstructNoEntry[Hypergraph, Hyperedge @@@ hyperedges, orderedQ]
+  ,
+    $Failed
+  ];
+
+hypergraphQ = MatchQ[{(_List | _Hyperedge) ...}];
+
+(* EdgeList, EdgeCount, VertexList, VertexCount *)
+
+Hypergraph /: EdgeList[HoldPattern[Hypergraph[hyperedgeList_, _] ? HypergraphQ]] :=
+  hyperedgeList;
+
+Hypergraph /: EdgeCount[hypergraph_Hypergraph ? HypergraphQ] :=
+  Length[EdgeList[hypergraph]];
+
+Hypergraph /: VertexList[hypergraph_Hypergraph ? HypergraphQ] :=
+  DeleteDuplicates[Catenate[Cases[EdgeList[hypergraph], Hyperedge[x___] :> {x}, {1}]]];
+
+Hypergraph /: VertexCount[hypergraph_Hypergraph ? HypergraphQ] :=
+  Length[VertexList[hypergraph]];
+
+(* Normal *)
+
+Hypergraph /: Normal[hypergraph_Hypergraph ? HypergraphQ] := List @@@ EdgeList[hypergraph];
+
+(* SameQ *)
+
+Hypergraph /: SameQ[hg1_Hypergraph, hg2_Hypergraph] := Normal[hg1] === Normal[hg2];
+
+(* HypergraphPlot *)
+
+Hypergraph /: HypergraphPlot[hypergraph_Hypergraph ? HypergraphQ, opts___] := HypergraphPlot[Normal[hypergraph], opts];
+
+(* Boxes *)
+
+Hypergraph /: MakeBoxes[hypergraph_Hypergraph, fmt_] /; HypergraphQ[hypergraph] :=
+  Module[{collapsed, expanded},
+    collapsed = BoxForm`SummaryItem /@ {
+      {"VertexCount: ", VertexCount[hypergraph]},
+      {"EdgeCount: ", EdgeCount[hypergraph]}
+    };
+    expanded = BoxForm`SummaryItem /@ {
+      {"OrderedQ: ", HypergraphOrderedQ[hypergraph]}
+    };
+    BoxForm`ArrangeSummaryBox[
+      Hypergraph,
+      hypergraph,
+      HypergraphPlot[hypergraph, ImageSize -> {29, 29}],
+      collapsed,
+      expanded,
+      fmt,
+      "Interpretable" -> True
+    ]
+  ];
+
+(* WolframModel *)
+
+Hypergraph /: WolframModel[rule_, hypergraph_Hypergraph, args___] :=
+  WolframModel[rule, Normal[hypergraph], args];
+
+(* HypergraphToGraph *)
+
+Hypergraph /: HypergraphToGraph[hypergraph_Hypergraph, args___] :=
+  HypergraphToGraph[Normal[hypergraph], args];
+
+(*
+(* CanonicalHypergraph *)
+
+(* ::Text:: *)
+(*Does CanonicalHypergraph take into consideration if the hypergraph is ordered?*)
+
+Hypergraph /: (func : ResourceFunction["CanonicalHypergraph"])[hypergraph_Hypergraph ? HypergraphQ] :=
+  Hypergraph[func[Normal[hypergraph]]]
+
+(*AdjacencyTensor*)
+
+AdjacencyTensor = ResourceFunction["AdjacencyTensor"][Normal[#], "OrderedHyperedges" -> HypergraphOrderedQ[#]] &;
+
+Hypergraph /: (func : ResourceFunction["AdjacencyTensor"])[hypergraph_Hypergraph ? HypergraphQ, opts : OptionsPattern[]] :=
+  func[Normal[hypergraph], opts, "OrderedHyperedges" -> HypergraphOrderedQ[hypergraph]]
+
+(*KirchhoffTensor*)
+
+Hypergraph /: (func : ResourceFunction["KirchhoffTensor"])[hypergraph_Hypergraph ? HypergraphQ, opts : OptionsPattern[]] :=
+  func[Normal[hypergraph], opts, "OrderedHyperedges" -> HypergraphOrderedQ[hypergraph]]
+*)


maxitg (commented):
* Interpretation boxes are necessary either way because we need them for large hypergraphs that cannot be plotted. Also, if we do return full-size plots, we'll have to be able to disable them (for performance), which means we need the `GraphLayout` option as well. So, I think we should keep it only as interpretation boxes for now.
* Since `Hyperedge`'s are returned by `EdgeList`, yes, of course, we need to export them. But I'm not sure we need `Hyperedge` at all.
* I don't know where we might want to use mixed types, but I think it would be best to design an API that would allow us to have that in the future. If we keep hyperedges as lists for now, we'll be able to do that later with custom hyperedge. However, we cannot use `Hyperedge[v1, v2, ...]` as there wouldn't be a place to put the symmetry.
* The default in SetReplace has always been `"Ordered"` for hypergraphs. As far as I remember, `HypergraphPlot` is the only function that even supports other types. I don't think we need to specify the orderness type at all in the first iteration, but if we do, it should certainly default to `"Ordered"`.

> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664#-:-MfJPJzB2BgONB-7KmNA:b-lq0400)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfJHI8RC-qqrpfK7rnO:b-kbstzx) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*
> ```objc
> PackageExport["Hypergraph"]
> PackageExport["HypergraphQ"]
> PackageExport["HypergraphOrderedQ"]
> ```

I suggest changing ordered/unordered to an enum (could be strings or symbols for now), as we might want to expand to more than two varieties in the future. For example, `"Ordered"`, `"Unordered"`, `"Cyclic"`, `"Directed"`.
___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJSD64EGMPNQtlKFbV:-MfJJ3qKBG6wkQCiyT1Q:bz3amfq) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*
> ```objc
> PackageExport["Hypergraph"]
> PackageExport["HypergraphQ"]
> PackageExport["HypergraphOrderedQ"]
> ```

If `Hyperedge` is returned by `EdgeList`, it should be exported.
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJkVd8iZdd4j0OTsU:-MfJJkVd8iZdd4j0OTsV:b-9py6ha) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*
> ```objc
> SyntaxInformation[Hypergraph] = {"ArgumentsPattern" -> {hyperedges_, orderedQ_.}};
> 
> Hypergraph[hyperedges_, orderedQ : (True | False) : False] ? System`Private`HoldEntryQ :=
> ```

Why would it default to `False`? We almost always deal either with ordered or directed hypergraphs.
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNE-qBiwMsRf2U-jA:-MfJNE-r6PQ8CIy9-13n:b-1tot9h) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*
> ```objc
> SyntaxInformation[Hypergraph] = {"ArgumentsPattern" -> {hyperedges_, orderedQ_.}};
> 
> Hypergraph[hyperedges_, orderedQ : (True | False) : False] ? System`Private`HoldEntryQ :=
> ```

It needs to generate messages if `hyperedges` is not valid. (We need to keep this consistent with the rest of SetReplace until we figure out what to do with the new approach.)
___
*[Kernel/Hypergraph.m, line 46 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJK9x8CqXUhppHgTE3:-MfJK9x8CqXUhppHgTE4:b78jx29) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L46)):*
> ```objc
>   ];
> 
> hypergraphQ = MatchQ[{(_List | _Hyperedge) ...}];
> ```

If we implement `Hyperedge` in this way, we won't be able to make them ordered/unordered in the future. It might be better to do it as `Hyperedge[{vertices...}, symmetry]`. Or, if we require uniform symmetries for all edges, I don't see any reason not to keep them as simple lists.
___
*[Kernel/Hypergraph.m, line 72 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNleL3K57ThwS__Hl:-MfJNleL3K57ThwS__Hm:blf2l88) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L72)):*
> ```objc
> (* HypergraphPlot *)
> 
> Hypergraph /: HypergraphPlot[hypergraph_Hypergraph ? HypergraphQ, opts___] := HypergraphPlot[Normal[hypergraph], opts];
> ```

The problem with implementing it like this (instead of in `HypergraphPlot`) is that it won't be able to generate messages correctly if something fails, e.g.,

```wl
HypergraphPlot[Hypergraph[{{2, 3}, {1, 2}, {2, 3}}], f -> 3]
```
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJIssSCWpSrCbAulNG:-MfJIssSCWpSrCbAulNH:bu2dhk2) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*
> ```objc
>       Hypergraph,
>       hypergraph,
>       HypergraphPlot[hypergraph, ImageSize -> {29, 29}],
> ```

Note that this fails for empty hyperedges, e.g.,

```wl
Hypergraph[{{}, {}, {1, 2, 3}}]
```
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJMUR8cPWC8b9HA91:-MfJJMUR8cPWC8b9HA92:b-3m1zzk) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*
> ```objc
>       Hypergraph,
>       hypergraph,
>       HypergraphPlot[hypergraph, ImageSize -> {29, 29}],
> ```

The plots should depend on the hypergraph type and should probably be disabled for now for anything other than ordered and cyclic hypergraphs.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJMUbKDh-rhuEvYhC6:-MfJMUbKDh-rhuEvYhC7:b-dm0bdq) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*
> ```objc
>       Hypergraph,
>       hypergraph,
>       HypergraphPlot[hypergraph, ImageSize -> {29, 29}],
> ```

Also note cases like this:

```wl
Hypergraph[{{{2, 3}, {2, 3}}, {{2, 3}, {2, 3}}}]
```

Although I think the correct solution is to stop `HypergraphPlot` from being listable (it was a bad idea in the first place).
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJPF2d-gDSpM1Qh3xh:-MfJPF2d-gDSpM1Qh3xi:b-csxzge) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*
> ```objc
>       Hypergraph,
>       hypergraph,
>       HypergraphPlot[hypergraph, ImageSize -> {29, 29}],
> ```

The plot should be disabled for large hypergraphs.
___
*[Kernel/Hypergraph.m, line 98 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJ_mK85D59Tc1UC3g:-MfJJ_mK85D59Tc1UC3h:bf05n8p) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L98)):*
> ```objc
> (* WolframModel *)
> 
> Hypergraph /: WolframModel[rule_, hypergraph_Hypergraph, args___] :=
> ```

`WolframModel` should only accept ordered hypergraphs.
___
*[Kernel/Hypergraph.m, line 98 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJO9GpEXl9RTkPbsqJ:-MfJO9GpEXl9RTkPbsqK:b-songx) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L98)):*
> ```objc
> (* WolframModel *)
> 
> Hypergraph /: WolframModel[rule_, hypergraph_Hypergraph, args___] :=
> ```

Same problem here as `HypergraphPlot`: it won't be able to generate messages correctly if something fails.
___
*[Kernel/Hypergraph.m, line 103 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJOWec3fSsQZqRJDc3:-MfJOWec3fSsQZqRJDc4:b-2rcnhs) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L103)):*
> ```objc
> (* HypergraphToGraph *)
> 
> Hypergraph /: HypergraphToGraph[hypergraph_Hypergraph, args___] :=
> ```

This depends on whether the hypergraph is ordered/unordered/directed as well.
___
*[Kernel/Hypergraph.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJOvzu8T1DIH09SDY_:-MfJOvzu8T1DIH09SDYa:b-8b90st) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L112)):*
> ```objc
> (*Does CanonicalHypergraph take into consideration if the hypergraph is ordered?*)
> 
> Hypergraph /: (func : ResourceFunction["CanonicalHypergraph"])[hypergraph_Hypergraph ? HypergraphQ] :=
> ```

We cannot have dependencies on resource functions.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJSD64EGMPNQtlKFbV:-MfJZGYL1pPECNBixck7:bqog7zb) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

If `Hyperedge` is returned by `EdgeList`, it should be exported.
</blockquote></details>

Agree.
___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfJZdtM0NtlSg8vXPVy:b3l8976) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

I suggest changing ordered/unordered to an enum (could be strings or symbols for now), as we might want to expand to more than two varieties in the future. For example, `"Ordered"`, `"Unordered"`, `"Cyclic"`, `"Directed"`.
</blockquote></details>

I know we discussed this before, but I forgot about the differences between these types of edges.
I think "Ordered" means `{a, b, c}` is different from `{a, c, b}`.
Does "Directed" relate to how this hyperedge would be converted to a set of edges? E.g. `{a, b, c}` would become `{a -> b, b -> c}`.
"Cyclic" (Directed) then means `{a, b, c}` -> `{a -> b, b -> c, c -> a}`?
___
*[Kernel/Hypergraph.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJOvzu8T1DIH09SDY_:-MfJgHRu8bqNH2HvSGme:bxtlao9) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L112)):*

We cannot have dependencies on resource functions.
</blockquote></details>

Yes, forgot to remove these.

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: all files reviewed, 14 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfJo_LWDGYdztz6QD6v:b-6l72jb) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

I know we discussed this before, but I forgot about the differences between these types of edges.
I think "Ordered" means `{a, b, c}` is different from `{a, c, b}`.
Does "Directed" relate to how this hyperedge would be converted to a set of edges? E.g. `{a, b, c}` would become `{a -> b, b -> c}`.
"Cyclic" (Directed) then means `{a, b, c}` -> `{a -> b, b -> c, c -> a}`?

</blockquote></details>

* Ordered hyperedge is a sequence. There is no symmetry. Every position is different. Same as `List` `{a, b, c}`.
* Directed is a pair of multisets, like an event hypergraph. It has two distinct parts, but vertices are not ordered within each part: `{<|a, b, c|>, <|d, e, f|>}`. This can correspond to event inputs and outputs.
* Cyclic means there is a cyclic symmetry. `{a, b, c}`, `{b, c, a}` and `{c, a, b}` are treated as identical.

So, to answer your questions:
1. Yes.
2. No, `{a, b, c}` is not a valid edge for a directed hypergraph. It should be `{a, b} -> {c}`, `{a} -> {b, c}` or something like that.
3. Kind of, except there is a difference between `{{a, b, c}}` and `{{a, b}, {b, c}, {c, a}}`.

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: 0 of 1 files reviewed, 14 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfLx8XmC9wJ4i9A30jI:bpvpcmx) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

* Ordered hyperedge is a sequence. There is no symmetry. Every position is different. Same as `List` `{a, b, c}`.
* Directed is a pair of multisets, like an event hypergraph. It has two distinct parts, but vertices are not ordered within each part: `{<|a, b, c|>, <|d, e, f|>}`. This can correspond to event inputs and outputs.
* Cyclic means there is a cyclic symmetry. `{a, b, c}`, `{b, c, a}` and `{c, a, b}` are treated as identical.

So, to answer your questions:
1. Yes.
2. No, `{a, b, c}` is not a valid edge for a directed hypergraph. It should be `{a, b} -> {c}`, `{a} -> {b, c}` or something like that.
3. Kind of, except there is a difference between `{{a, b, c}}` and `{{a, b}, {b, c}, {c, a}}`.
</blockquote></details>

Thanks for the explanation!
I've supported these as "enums"... but now that I read what you said, for `Directed` I need to change the check I do to the hyperedges (`{___List}` is not enough anymore).
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNE-qBiwMsRf2U-jA:-MfLwvvH2G0r6OJDnGZz:b-xidwe1) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*

It needs to generate messages if `hyperedges` is not valid. (We need to keep this consistent with the rest of SetReplace until we figure out what to do with the new approach.)
</blockquote></details>

Returning `Failure`'s for now... let me know what you think
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJkVd8iZdd4j0OTsU:-MfLx0IR7NlDXyX2SpzA:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*

Why would it default to `False`? We almost always deal either with ordered or directed hypergraphs.
</blockquote></details>

Done.
___
*[Kernel/Hypergraph.m, line 46 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJK9x8CqXUhppHgTE3:-MfLwrD7-5jHy6wXnumc:bt5haux) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L46)):*

If we implement `Hyperedge` in this way, we won't be able to make them ordered/unordered in the future. It might be better to do it as `Hyperedge[{vertices...}, symmetry]`. Or, if we require uniform symmetries for all edges, I don't see any reason not to keep them as simple lists.
</blockquote></details>

Keeping it as simple lists :) 
___
*[Kernel/Hypergraph.m, line 72 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNleL3K57ThwS__Hl:-MfLwoeUB_XpM1Tx7eX-:bqog7zb) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L72)):*

The problem with implementing it like this (instead of in `HypergraphPlot`) is that it won't be able to generate messages correctly if something fails, e.g.,

```wl
HypergraphPlot[Hypergraph[{{2, 3}, {1, 2}, {2, 3}}], f -> 3]
```
</blockquote></details>

Agree.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJPF2d-gDSpM1Qh3xh:-MfLwAPl6fqDhH7zI0Ts:bajegsc) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

The plot should be disabled for large hypergraphs.
</blockquote></details>

What should be the cutoff?
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJMUbKDh-rhuEvYhC6:-MfLwH9r3bE8ebMYam5K:b-nc13d5) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

Also note cases like this:

```wl
Hypergraph[{{{2, 3}, {2, 3}}, {{2, 3}, {2, 3}}}]
```

Although I think the correct solution is to stop `HypergraphPlot` from being listable (it was a bad idea in the first place).
</blockquote></details>

Yes, in hindsight that looks wrong. I'll remove that once I support `Hypergraph` in `HypergraphPlot`.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJIssSCWpSrCbAulNG:-MfLwbqK7cwVmEXvT_r-:b-l47hs) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

Note that this fails for empty hyperedges, e.g.,

```wl
Hypergraph[{{}, {}, {1, 2, 3}}]
```
</blockquote></details>

So this should be handled when `HypergraphPlot` receives a `Hypergraph`, right?
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJMUR8cPWC8b9HA91:-MfLwjBs32JaMqbATxw7:b8rdyze) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

The plots should depend on the hypergraph type and should probably be disabled for now for anything other than ordered and cyclic hypergraphs.
</blockquote></details>

What should be the default icon? The one from `WolframModelEvolutionObject`?
___
*[Kernel/Hypergraph.m, line 98 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJO9GpEXl9RTkPbsqJ:-MfLwZ_rC9T1RLIHG8vG:b-k87ebk) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L98)):*

Same problem here as `HypergraphPlot`: it won't be able to generate messages correctly if something fails.
</blockquote></details>

Agree, I'll move the definition over there.
___
*[Kernel/Hypergraph.m, line 112 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJOvzu8T1DIH09SDY_:-MfLw5Hc9Uo_yKtoQ7GN:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L112)):*

Yes, forgot to remove these.
</blockquote></details>

Done.

maxitg (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: all files reviewed, 9 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfPYxEdALpPnncU1veM:b-pyc5re) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

Thanks for the explanation!
I've supported these as "enums"... but now that I read what you said, for `Directed` I need to change the check I do to the hyperedges (`{___List}` is not enough anymore).
</blockquote></details>

Yeah, maybe for "Directed" we should use something like this: `DirectedEdge[{1, 2, 3}, {3, 4, 5}]`?
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNE-qBiwMsRf2U-jA:-MfPSrn143dNXz1VUjO6:b-cq2pir) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*

Returning `Failure`'s for now... let me know what you think
</blockquote></details>

I think we should do the same thing everywhere. So, I suggest using the mechanism from `A0$messages.m` for now to return unevaluated. I also still think throwing failures would be better than returning them. The main issue with returning failures is that it's too silent. If a function is not supposed to return anything, or if what it returns is ignored (or interpreted incorrectly by the subsequent code), it would be impossible to tell that something wrong has happened.

Also, note that it currently prints messages for an incorrect number of arguments. That needs to be consistent as well.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJIssSCWpSrCbAulNG:-MfPVnbR1-BJBKrJt_Q_:bu9xbly) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

So this should be handled when `HypergraphPlot` receives a `Hypergraph`, right?
</blockquote></details>

Well, yes, ideally we want to fix `HypergraphPlot` so that it can display empty edges in some way (which is mostly a graphical design question).
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJMUR8cPWC8b9HA91:-MfPWCCsBbT83phD6wVe:b5smrqf) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

What should be the default icon? The one from `WolframModelEvolutionObject`?
</blockquote></details>

I'm not sure we should display an icon at all. It might be misleading if the icon is usually correct, but not always. However, if we do have an icon, maybe it's better to use one that does not look like a `HypergraphPlot` at all?
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJPF2d-gDSpM1Qh3xh:-MfPXQBK1kknh8Fw_BNj:b9qvyot) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

What should be the cutoff?
</blockquote></details>

That's a good question. Maybe a second (inferred from the number of edges)? That would cause issues if one tries to return lots of hypergraphs simultaneously, but it's not any worse than what `Graph` is doing. Ideally, we want to plot asynchronously so that we don't interfere with the front end, but I don't think that's possible.
___
*[Kernel/Hypergraph.m, line 24 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPRgbc-aShFzAV4ru2:-MfPRgbc-aShFzAV4ru3:b6cd94b) ([raw file](https://github.com/maxitg/setreplace/blob/0c2ae1fa511ff29efbc90263d80a87c141241495/Kernel/Hypergraph.m#L24)):*
> ```objc
> $hypergraphSymmetries = {"Ordered", "Unordered", "Cyclic", "Directed"};
> 
> SetUsage @ "HypergraphSymmetry[hg$] returns the symmetry of the hypergraph hg$.";
> ```

I think this needs some explanation of what symmetry is.
___
*[Kernel/Hypergraph.m, line 41 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPS9NRA8pcgmMOkxjx:-MfPS9NRA8pcgmMOkxjy:b-395srl) ([raw file](https://github.com/maxitg/setreplace/blob/0c2ae1fa511ff29efbc90263d80a87c141241495/Kernel/Hypergraph.m#L41)):*
> ```objc
>   <|
>     "ObjectType" -> Hypergraph,
>     "Symmetry" -> HypergraphSymmetry,
> ```

```wl
HypergraphSymmetry[obj]
```
___
*[Kernel/Hypergraph.m, line 54 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPUgTXEMM7uyQIvlph:-MfPUgTXEMM7uyQIvlpi:bkzvofw) ([raw file](https://github.com/maxitg/setreplace/blob/0c2ae1fa511ff29efbc90263d80a87c141241495/Kernel/Hypergraph.m#L54)):*
> ```objc
>   System`Private`ConstructNoEntry[Hypergraph, hyperedges, symmetry];
> 
> Hypergraph::invalidHyperedges = "The argument at position 1 should be ";
> ```

This does not appear to be used anywhere.
___
*[Kernel/Hypergraph.m, line 56 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPUn05BX8AAnBjXlEK:-MfPUn05BX8AAnBjXlEL:bb7mq5g) ([raw file](https://github.com/maxitg/setreplace/blob/0c2ae1fa511ff29efbc90263d80a87c141241495/Kernel/Hypergraph.m#L56)):*
> ```objc
> Hypergraph::invalidHyperedges = "The argument at position 1 should be ";
> 
> hypergraph[hyperedges_] := hypergraph[hyperedges, "Ordered"];
> ```

Isn't this redundant with `Default[Hypergraph, 2]` above?

daneelsan (commented):
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: 0 of 1 files reviewed, 9 unresolved discussions (waiting on @daneelsan and @maxitg)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfiPqTs2DujIpY3-xVG:brf9qxm) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

Yeah, maybe for "Directed" we should use something like this: `DirectedEdge[{1, 2, 3}, {3, 4, 5}]`?
</blockquote></details>

That makes sense. But that would not be "global" symmetry, right?
___
*[Kernel/Hypergraph.m, line 39 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJNE-qBiwMsRf2U-jA:-MfZF9aR6jPI47hwGmVM:b-whnj07) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L39)):*

I think we should do the same thing everywhere. So, I suggest using the mechanism from `A0$messages.m` for now to return unevaluated. I also still think throwing failures would be better than returning them. The main issue with returning failures is that it's too silent. If a function is not supposed to return anything, or if what it returns is ignored (or interpreted incorrectly by the subsequent code), it would be impossible to tell that something wrong has happened.

Also, note that it currently prints messages for an incorrect number of arguments. That needs to be consistent as well.
</blockquote></details>

Yes, I agree that returning `Failure`'s is too silent. I'll change back to messages + unevaluated code.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJIssSCWpSrCbAulNG:-MfiPTTG72Qht9jrAavp:bwkicza) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

Well, yes, ideally we want to fix `HypergraphPlot` so that it can display empty edges in some way (which is mostly a graphical design question).
</blockquote></details>

I thought `{}` had to be ignored :thinking: 
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJMUR8cPWC8b9HA91:-MfiPd7MAn8j-uJuVB6Z:balr58b) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

I'm not sure we should display an icon at all. It might be misleading if the icon is usually correct, but not always. However, if we do have an icon, maybe it's better to use one that does not look like a `HypergraphPlot` at all?
</blockquote></details>

Yeah, I can't design a generic hypergraph icon...
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJPF2d-gDSpM1Qh3xh:-MfiPi1G1v9jfBJfq0ts:bruwm75) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

That's a good question. Maybe a second (inferred from the number of edges)? That would cause issues if one tries to return lots of hypergraphs simultaneously, but it's not any worse than what `Graph` is doing. Ideally, we want to plot asynchronously so that we don't interfere with the front end, but I don't think that's possible.
</blockquote></details>

Using 100 edges as the cutoff for now... 
___
*[Kernel/Hypergraph.m, line 41 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPS9NRA8pcgmMOkxjx:-MfebDbK4J5jXP3sp259:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/9a5dc9c3aee2cf2d5503ea72647c262edfd0ed1f/Kernel/Hypergraph.m#L41)):*

```wl
HypergraphSymmetry[obj]
```
</blockquote></details>

Done.
___
*[Kernel/Hypergraph.m, line 54 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPUgTXEMM7uyQIvlph:-MfiOyBrC20VjjSGOLnJ:b-896fix) ([raw file](https://github.com/maxitg/setreplace/blob/9a5dc9c3aee2cf2d5503ea72647c262edfd0ed1f/Kernel/Hypergraph.m#L54)):*

This does not appear to be used anywhere.
</blockquote></details>

Done.
___
*[Kernel/Hypergraph.m, line 56 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPUn05BX8AAnBjXlEK:-MfiOyxp11F6a-Se7bb4:buckc1v) ([raw file](https://github.com/maxitg/setreplace/blob/9a5dc9c3aee2cf2d5503ea72647c262edfd0ed1f/Kernel/Hypergraph.m#L56)):*

Isn't this redundant with `Default[Hypergraph, 2]` above?
</blockquote></details>

I would have liked to use `Default`, but I don't know how to make this work:
```
((expr : Hypergraph[args___]) ? System`Private`HoldEntryQ) /; CheckArguments[expr, {1, 2}] :=
```

Somewhere in there should be a `symmetry_.`, but don't know how.

maxitg (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/664)* status: all files reviewed, 7 unresolved discussions (waiting on @daneelsan)

___
*[Kernel/Hypergraph.m, line 7 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJHI8RC-qqrpfK7rnN:-MfmLHoI6lb-mOuKM-6L:b-lmmecg) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L7)):*

That makes sense. But that would not be "global" symmetry, right?
</blockquote></details>

It cannot be a single group for all hyperedges because they can contain different numbers of vertices, e.g., `{1, 2} -> {3}` and `{1} -> {2, 3}`. But the same would also be the case for other symmetries since the symmetry groups would have different sizes based on the hyperedge spec. So, we cannot specify the symmetry group globally in any case. It has to be a function from a hyperedge spec to a group, in which case it works for the `"Directed"` case as well.

Essentially,

```wl
"Ordered" -> (SymmetricGroup[1] &)
"Unordered" -> (SymmetricGroup @* Length)
"Cyclic" -> (CyclicGroup @* Length)
"Directed" -> PermutationGroup[Join[
  GroupElements[SymmetricGroup[Length[#[[1]]]]],
  GroupElements[SymmetricGroup[Length[#[[2]]]]] /. n_Integer :> n + Length[#[[1]]]]] &
```
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJIssSCWpSrCbAulNG:-MfmVBbP68B8pkhrldcG:b-wr8ovt) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

I thought `{}` had to be ignored :thinking: 
</blockquote></details>

Why would they be ignored? They are perfectly reasonably 0-arity hyperedges. So perhaps we can plot them as circles away from the main graph. I.e., the same as unary edges but without a vertex inside.

So, if a graph has 2 0-arity edges, we might plot it like these:

```wl
In[] := HypergraphPlot[{{}, {}, {1, 2, 3}, {3, 4}, {4, 1}}]
(* HypergraphPlot[{{0}, {0}, {1, 2, 3}, {3, 4}, {4, 1}}, VertexStyle -> <|0 -> White|>] *)
```



Or, perhaps, like this:

```wl
In[] := HypergraphPlot[{{}, {}, {1, 2, 3}, {3, 4}, {4, 1}}]
(* HypergraphPlot[{{0}, {-1}, {1, 2, 3}, {3, 4}, {4, 1}}, VertexStyle -> <|n_ /; n <= 0 -> White|>] *)
```



It wouldn't even be difficult to implement because one can use the trick of adding transparent vertex/vertices.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJJMUR8cPWC8b9HA91:-MfmX-fBEQ89r8KOeGq6:b-k2voh) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

Yeah, I can't design a generic hypergraph icon...
</blockquote></details>

I think the one for `Graph` is just a zoomed-in example of a `GraphPlot`. Maybe, we can do something similar with `HypergraphPlot` as well.
___
*[Kernel/Hypergraph.m, line 88 at r1](https://reviewable.io/reviews/maxitg/setreplace/664#-MfJPF2d-gDSpM1Qh3xh:-MfmXBpB408-7kkidml3:b-32w0iv) ([raw file](https://github.com/maxitg/setreplace/blob/5bd3dba7e313fc55e137f55973b65022aa61a035/Kernel/Hypergraph.m#L88)):*

Using 100 edges as the cutoff for now... 
</blockquote></details>

Sounds reasonable.
___
*[Kernel/Hypergraph.m, line 56 at r2](https://reviewable.io/reviews/maxitg/setreplace/664#-MfPUn05BX8AAnBjXlEK:-MfmSUyREaj7eSlVNMl2:b-ku7q2f) ([raw file](https://github.com/maxitg/setreplace/blob/9a5dc9c3aee2cf2d5503ea72647c262edfd0ed1f/Kernel/Hypergraph.m#L56)):*

I would have liked to use `Default`, but I don't know how to make this work:
```
((expr : Hypergraph[args___]) ? System`Private`HoldEntryQ) /; CheckArguments[expr, {1, 2}] :=
```

Somewhere in there should be a `symmetry_.`, but don't know how.
</blockquote></details>

Why not simply put it as

```wl
hypergraph[hyperedges : {___List}, symmetry : Alternatives @@ $hypergraphSymmetries : "Ordered"] :=
  System`Private`ConstructNoEntry[Hypergraph, hyperedges, symmetry];
```
___
*[Kernel/Hypergraph.m, line 64 at r3](https://reviewable.io/reviews/maxitg/setreplace/664#-MfmRMDX5Y0hTLdu6DZX:-MfmRMDX5Y0hTLdu6DZY:br7txrs) ([raw file](https://github.com/maxitg/setreplace/blob/6885ceda1be05b9adea4c3e012d6602213dff444/Kernel/Hypergraph.m#L64)):*
> ```objc
> 
> hypergraph[hyperedges_, symmetry : Alternatives @@ $hypergraphSymmetries] :=
>   throw[Failure["invalidHyperedges"]];
> ```

You need to pass an empty association as the second argument of `Failure`. Otherwise, `message` fails because it cannot join `"expr"` into it.
___
*[Kernel/Hypergraph.m, line 95 at r3](https://reviewable.io/reviews/maxitg/setreplace/664#-MfmUgz50lVe3W9MOtCO:-MfmUgz50lVe3W9MOtCP:btmu8k1) ([raw file](https://github.com/maxitg/setreplace/blob/6885ceda1be05b9adea4c3e012d6602213dff444/Kernel/Hypergraph.m#L95)):*
> ```objc
> 
> getIcon[hg_] /; (!disablePlotQ[hg] && MemberQ[$edgeTypes, HypergraphSymmetry[hg]]) :=
>   HypergraphPlot[EdgeList[hg], HypergraphSymmetry[hg], ImageSize -> {29, 29}];
> ```

It looks like the image size is not consistent between `HypergraphPlot`s and `$evolutionObjectIcon`. Perhaps, it might be better to try the same code as `$graphIcon` uses, i.e.,

```wl
ImageSize -> Dynamic[{
  Automatic,
  3.5` CurrentValue["FontCapHeight"] / AbsoluteCurrentValue[Magnification]}]
```

PR #668, opened 2021-08-29
Enable compiler optimizations for WolframScript builds.

georgemorgan:
## Changes

I noticed that the `WolframModel` evolutions were taking an extremely long time to evaluate. I also noticed that the fairly straightforward evaluation of `RandomWolframModel` was completely locking up my computer. After looking at the dylib produced by `DevUtils/BuildLibSetReplace.m` I noticed that it was not being build with compiler optimizations, so I added them.

## Comments

Fairly simple, low risk change here assuming we trust the compiler. Should yield more performance for all who want to use this library. :)

## Examples

Run `./install.wls` to build with compiler optimizations.

---
This change is [Reviewable](https://reviewable.io/reviews/maxitg/setreplace/668)

### DIFF
--- a/DevUtils/BuildLibSetReplace.m
+++ b/DevUtils/BuildLibSetReplace.m
@@ -173,11 +173,11 @@ $warningsFlags = {
 
 $compileOptions = Switch[$OperatingSystem,
   "Windows",
-    {"/std:c++17", "/EHsc"},
+    {"/std:c++17 /O3", "/EHsc"},
   "MacOSX",
-    Join[{"-std=c++17"}, $warningsFlags, {"-mmacosx-version-min=10.12"}], (* for std::shared_mutex support *)
+    Join[{"-std=c++17 -O3"}, $warningsFlags, {"-mmacosx-version-min=10.12"}], (* for std::shared_mutex support *)
   "Unix",
-    Join[{"-std=c++17"}, $warningsFlags]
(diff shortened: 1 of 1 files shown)


maxitg (approved):
Thanks! `CreateLibrary` uses `-O2` automatically, so it's not as significant as it seems, but there appears to be a few-percent improvement with `-O3` (and tests still pass), so it's a good idea (ignore "Multiset addition" below — it's not using the C++ code and is probably due to throttling):

```console
$ ./performanceTest.wls master enable_optimizations 10
Testing 3727fe34f8481326798c1382c469b0131ce2795e

Testing abf307fc33ab137373f65a436ebaa90a9059ed55

Single-input rule                       -1.5 ± 0.7 %
Medium rule                             3.4 ± 0.6 %
Sequential rule                         1.8 ± 1.2 %
Large rule                              2.8 ± 0.5 %
Exponential-match-count rule            -0.8 ± 0.8 %
CA emulator                             -2.3 ± 1.7 %
Multiset addition                       -26. ± 16. %
```

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/maxitg/setreplace/668#-:-MiIOZiR-BtQhX9YO_cl:baqrlms)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @georgemorgan)

PR #678, opened 2025-01-31
Show test counter in test.wls

maxitg:
## Changes

* Add a test counter to indicate progress while the tests are running.
* Reset the timeout for no terminal output to 10 minutes.

## Comments

* CircleCI times out if there is no terminal output for 10 minutes by default and for 20 minutes given our current config.
* We could extend the timeout further, but it's better to show actual progress. It's also useful to get an estimate for how long one has to wait for the tests to be done.

## Examples

Screenshot 2025-01-31 at 03 12 19

- - -
This change is [Reviewable](https://reviewable.io/reviews/maxitg/SetReplace/678)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,6 @@ jobs:
 
       - run:
           name: Test
-          no_output_timeout: 20m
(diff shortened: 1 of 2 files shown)


daneelsan (approved):
image
Cool!

